PR #2, opened 2016-06-09
Make toolbar accessible

MathuraMG:
[Trello card](https://trello.com/c/dQ2cgjhf/1-ui-elements)

This PR:

* changes the play and stop elements to buttons 
* makes the stop button always visible ( to increase accessibility)
* adds title to the iframe

### DIFF
--- a/shared/components/Toolbar/Toolbar.jsx
+++ b/shared/components/Toolbar/Toolbar.jsx
@@ -10,11 +10,11 @@ class Toolbar extends React.Component {
 	render() {
 		let playButtonClass = classNames({
 			"toolbar__play-button": true,
-			"playing": this.props.isPlaying
+			"toolbar__play-button--selected": this.props.isPlaying
 		});
 		let stopButtonClass = classNames({
 			"toolbar__stop-button": true,
-			"stopped": !this.props.isPlaying
+			"toolbar__stop-button--selected": !this.props.isPlaying
 		});
 
 		return (
--- a/shared/components/Toolbar/Toolbar.jsx
+++ b/shared/components/Toolbar/Toolbar.jsx
@@ -12,18 +12,20 @@ class Toolbar extends React.Component {
 			"toolbar__play-button": true,
 			"playing": this.props.isPlaying
 		});
+		let stopButtonClass = classNames({
+			"toolbar__stop-button": true,
+			"stopped": !this.props.isPlaying
+		});
 
 		return (
 			<div className="toolbar">
 				<img className="toolbar__logo" src={logoUrl}/>
-				<div className={playButtonClass} onClick={this.props.startSketch}>
+				<button className={playButtonClass} onClick={this.props.startSketch}>
 					<Isvg src={playUrl} alt="Play Sketch" />
-				</div>
-				{ this.props.isPlaying ? 
-					<div className="toolbar__stop-button" onClick={this.props.stopSketch}>
-						<Isvg src={stopUrl} alt="Stop Sketch" />
-					</div> 
-				: null }
+				</button>
+				<button className={stopButtonClass} onClick={this.props.stopSketch}>
+					<Isvg src={stopUrl} alt="Stop Sketch" />
+				</button>
 			</div>
 		);
 	}
--- a/styles/components/_toolbar.scss
+++ b/styles/components/_toolbar.scss
@@ -11,6 +11,12 @@
 
 .toolbar__stop-button {
 	@extend %toolbar-button;
+	&.stopped {
+		background-color: $light-button-background-hover-color;
+		& g {
+			fill: $light-button-hover-color;
+		}
+	}
 }
 
 .toolbar__logo {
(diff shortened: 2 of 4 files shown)


catarak:
Now that the play/pause buttons are button elements, they have an outline when you click them. Does adding `outline: none` affect tabbing in accessibility?

Also, for the css, I prefer to use [BEM](http://getbem.com/naming/)

catarak:
Yes exactly!

MathuraMG:
I added BEM style to the buttons, however both styles look similar now.
What do you think of merging both to just `.toolbar__button `and `.toolbar__button--selected` ?

catarak:
The BEM philosophy is for each HTML element to have one class (though I find myself breaking this more than occasionally), because CSS specificity rules get annoying fast and are hard to debug. I recommend reading about BEM to understand why it's used. There are lots of other CSS philosophies, like OOCSS, but for me, I find using BEM the easiest.

In the Toolbar component, the buttons should be either `.toolbar__<type>-button--selected` or `.toolbar__<type>-button`. 

Then, to make all of this work with SCSS, what you can do is in the %toolbar-button placeholder in _placeholders.scss, change it to
```scss
%toolbar-button {
   //...existing styles...
  &--selected {
	background-color: $light-button-background-hover-color;
	& g {
		fill: $light-button-hover-color;
	}
  }
}
```

Then, in _toolbar.scss, the stop button would look like
```scss
.toolbar__stop-button {
	@extend %toolbar-button;
	&--selected {
		@extend %toolbar-button;
		@extend %toolbar-button--selected;
	}
}
```

MathuraMG:
Thanks for the above, I'll make the changes !

I stand corrected on what I said about `outline:none` . I came came across a lot of articles that mention that it is quite important, especially on `:focus` ( for instance - [article](http://webaim.org/blog/plague-of-outline-0/) ) 
To take care of this, the simple solution would be to add the below in `%toolbar-button`
       ```
&:focus {
		outline: 1px solid #dddddd;
	}
```
But I see how this throws off the aesthetics. As this works well with screen readers right now ( with `outline:none;` ), what do you think off making this a separate item so that we can also address the design?

**Comment by catarak** [comment]
I'm okay with figuring out the outline stuff later! For this project I know that there is some design budget so we can just save it for when we need a bunch of design input.

**Comment by MathuraMG** [comment]
Great, I've added a card to look at this - [trello card](https://trello.com/c/0wB9SJP3/8-have-a-visual-cue-like-outline-when-buttons-are-in-focus)

File: shared/components/Toolbar/Toolbar.jsx [hunk 1]
```diff
@@ -10,23 +10,25 @@ class Toolbar extends React.Component {
 	render() {
 		let playButtonClass = classNames({
 			"toolbar__play-button": true,
```
catarak: this line should be 
`"toolbar__play-button": !this.props.isPlaying`

shared/components/Toolbar/Toolbar.jsx
```diff
@@ -10,23 +10,25 @@ class Toolbar extends React.Component {
...
-			"playing": this.props.isPlaying
+			"toolbar__play-button--selected": this.props.isPlaying
+		});
+		let stopButtonClass = classNames({
+			"toolbar__stop-button": true,
```
catarak: this line should be
`"toolbar__stop-button": this.props.isPlaying`

styles/components/_toolbar.scss
```diff
@@ -1,16 +1,16 @@
 .toolbar__play-button {
 	@extend %toolbar-button;
 	margin-right: #{15 / $base-font-size}rem;
-	&.playing {
-		background-color: $light-button-background-hover-color;
-		& g {
-			fill: $light-button-hover-color;
-		}
+	&--selected {
+		@extend %toolbar-button--selected;
```
catarak: this should be
```
@extend %toolbar-button;
@extend %toolbar-button--selected;
```

styles/components/_toolbar.scss
```diff
@@ -1,16 +1,16 @@
...
 	}
 }
 
 .toolbar__stop-button {
 	@extend %toolbar-button;
+	&--selected {
+		@extend %toolbar-button--selected;
```
catarak: this should be
```
@extend %toolbar-button;
@extend %toolbar-button--selected;
```

MathuraMG:
In the current code, the html works as follows.
**when play is selected**

```
<button class="toolbar__play-button toolbar__play-button--selected">
<button class="toolbar__stop-button">

```
**when stop is selected**

```
<button class="toolbar__play-button">
<button class="toolbar__stop-button  toolbar__stop-button--selected">

```
Based on my findings while reading up on BEM, the standard seems to be that the modifier class is added to the base class instead of replacing it.
[BEM Naming](http://getbem.com/naming/)

```
<div class="block block--mod">...</div>
	<div class="block block--size-big
		block--shadow-yes">...</div>
```

Additionally, it seems like replacing it has a few drawbacks.
[link](http://getbem.com/faq/#why-the-modifier-classes-are-prefixed)

That's why I have structured the code as is. Please let me know if I'm missing something and we should go for the single class approach.

catarak:
Cool, my bad, let's leave it that way then. 😄

PR #12, opened 2016-07-13
add aria tags and roles

MathuraMG:
In terms of accessibility, this PR
* Add aria-roles to 'nav bar' and 'values' in preferences
* Add titles to lists 
* add 'region role' to the editor and iframe - it now shows up in the 'Landmarks' in voiceover
* make codemirror inputStyle contenteditable
@CleezyITP could you verify if the roles are ok?

It also adds parseInt to updating preferences via the input field. 
There was a bug where if you increased the value of say, fontSize after editing it in the input field - it added '2' as a string.
For instance, if I entered 12 in the input field and the pressed the increase button - the font became 122.

### DIFF
--- a/client/components/Nav.js
+++ b/client/components/Nav.js
@@ -3,8 +3,8 @@ import { Link } from 'react-router';
 
 function Nav(props) {
   return (
-    <nav className="nav">
-      <ul className="nav__items-left">
+    <nav className="nav" role="navigation">
+      <ul className="nav__items-left" title="project-menu">
         <li className="nav__item">
           <a
             className="nav__new"
@@ -29,7 +29,7 @@ function Nav(props) {
           </p>
         </li>
       </ul>
-      <ul className="nav__items-right">
+      <ul className="nav__items-right" title="user-menu">
         <li className="nav__item">
           {props.user.authenticated && <p>Hello, {props.user.username}!</p>}
           {!props.user.authenticated && <p><Link to="/login">Login</Link> or <Link to="/signup">Sign Up</Link></p>}
--- a/client/components/Nav.js
+++ b/client/components/Nav.js
@@ -3,7 +3,7 @@ import { Link } from 'react-router';
 
 function Nav(props) {
   return (
-    <nav className="nav" role="navigation">
+    <nav className="nav" role="navigation" title="main-navigation">
       <ul className="nav__items-left" title="project-menu">
         <li className="nav__item">
           <a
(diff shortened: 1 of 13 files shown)


MathuraMG:
@CleezyITP Am adding your comments here -

1) I'm worried about the "title" tag being used for the preference-menu:
https://www.paciellogroup.com/blog/2012/01/html5-accessibility-chops-title-attribute-use-and-abuse/
can we also call it a section element?
2) similarly witth the side links. instead of sidebar
Can we give it a semantic tag? like 
<aside>?

MathuraMG:
@CleezyITP 
* I did change the preference to a section tag and tried adding the section role. However, if I don't add the title element, it reads the below on focus - 
screen shot 2016-07-13 at 11 08 42 pm

* Would `<nav>` work instead of `<aside>` - since, it's being used to navigate through the files?

CleezyITP:
Ah, shucks. I guess I didn't know how a section behaves. Your suggestion of the <nav> sounds reasonable, but I worry it might confuse because it isn't a real nav...let's give it a try and hear some feedback about what people would prefer!

catarak:
There's already a `<nav>` on the page already, so would it be confusing to have two?

MathuraMG:
It should be okay to have 2 `<nav>` if we label then differently I think (we would have to use `title` though). Also seems like it might prove helpful for [screenreaders](https://www.w3.org/TR/html5/sections.html#the-nav-element) and we can navigate straight to it using landmarks.

MathuraMG:
@CleezyITP  Do the new tags look ok for testing?

catarak:
Let when this PR is ready to be merged!

MathuraMG:
We got a bit more of feedback right now from Chancey - will just add those too and let you know!

MathuraMG:
@CleezyITP Included labels for all the buttons as mentioned by Chancy. 

Since the buttons don't have labels - one option was to included a `title ` for each of them but that was being overused and the other option was to create a label and style as `display:none`. However, screen readers do not read that.
So, I followed the [suggestion on WebAIM](http://webaim.org/techniques/css/invisiblecontent/#absolutepositioning) for styling them - hope this works!

MathuraMG:
Hi @catarak , spoke to @CleezyITP - looks like we can merge this.
Would it available on the link you sent us after it's merged?

catarak:
All looks good! FYI, there are some labels on the Sign In/Login pages that are `display: none` and not positioned absolutely off the screen, that should probably be fixed eventually.

MathuraMG:
Sure, I'll do that!

PR #13, opened 2016-07-18
Console

therewasaguy:
- a basic, non-interactive Console component
- previewFrame injects a script into the iframe that uses postMessage to send errors and console messages. It catches referenceErrors with window.onError. It doesn't catch syntax errors (seems like these are already caught with linting?)

### DIFF
--- a/client/styles/components/_console.scss
+++ b/client/styles/components/_console.scss
@@ -1,0 +1,14 @@
+.preview-console {
+	position: fixed;
+	width:100%;
+	height:60px;
+	right:0px;
+	bottom: 0px;
+	background:grey;
+	z-index:1000;
+
+	& > {
+		position:relative;
+		text-align:left;
+	}
+}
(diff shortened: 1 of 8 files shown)


client/styles/components/_console.scss
```diff
@@ -0,0 +1,27 @@
...
+	background:grey;
+	z-index:1000;
+
+	& > {
+		position:relative;
+		text-align:left;
+	}
+
+	// assign styles to different types of console messages
+	.log {
```
catarak: for this project I'm using BEM naming convention, so these class names should be .preview-console__log, .preview-console__warn, and .preview-console__error. Let's also make these colors the same as the linting error colors, #ffbe05 and #ff5f52. all of the other styling is fine for now.

therewasaguy: sounds good! I'll fix that

therewasaguy:
updated with BEM style

PR #50, opened 2016-08-16
Create text output for canvas

MathuraMG:
This PR creates a text output for the canvas. This involves the following elements - 

* an extra play button that can be used to turn on the text output (which is hidden)
* a preference option that can be used to turn on and off the text output (this is not hidden)
* the output itself which is split into summary and details (this is hidden)
* `intercept-p5.js` and `interceptor-functions.js` contains the functions that run the monkey patch on the existing p5 functions. They are being included in `PreviewFrame.js`
* `loadData.js` is another static file that is being used to load the `data.json` file from the following [location](https://raw.githubusercontent.com/processing/p5.js-website/9f1a4cd299c1330b046373407d42894e274d20e7/reference/data.min.json)(cc @lmccart ) 
* 'TextOutput.js` is the react component to render the output. The logic for populating them is within the interceptor files

Below picture shows all elements (screenshot is before they were hidden)
screen shot 2016-08-15 at 8 29 34 pm

(cc @CleezyITP)


static/loadData.js
```diff
@@ -0,0 +1,17 @@
...
+            if (httpRequest.status === 200) {
+                var data = JSON.parse(httpRequest.responseText);
+                if (callback) callback(data);
+            }
+        }
+    };
+    httpRequest.open('GET', path,false);
+    httpRequest.send();
+}
+fetchJSONFile('https://raw.githubusercontent.com/processing/p5.js-website/9f1a4cd299c1330b046373407d42894e274d20e7/reference/data.min.json', function(data){
```
catarak: We shouldn't rely on serving a file from Github. You could put this file in the static directory, and then parse that, or, maybe it an API endpoint that serves this JSON.

client/modules/IDE/components/PreviewFrame.js
```diff
@@ -127,12 +127,18 @@ class PreviewFrame extends React.Component {
...
-    // htmlFile = htmlFile.replace(/(?:<head.*?>)([\s\S]*?)(?:<\/head>)/gmi, `<head>\n${htmlHeadContents}\n</head>`);
+    if (this.props.textOutput || this.props.isTextOutputPlaying) {
+      const htmlHead = htmlFile.match(/(?:<head.*?>)([\s\S]*?)(?:<\/head>)/gmi);
+      const headRegex = new RegExp('head', 'i');
+      let htmlHeadContents = htmlHead[0].split(headRegex)[1];
+      htmlHeadContents = htmlHeadContents.slice(1, htmlHeadContents.length - 2);
+      htmlHeadContents += '<script src="/loadData.js"></script>\n';
+      htmlHeadContents += '<script src="/interceptor-functions.js"></script>\n';
+      htmlHeadContents += '<script src="/intercept-p5.js"></script>\n';
+      htmlHeadContents += '<script type="text/javascript" src="http://chir.ag/projects/ntc/ntc.js"></script>';
```
catarak: We probably want to serve this file locally as well, even though I know the person said you can link directly to this file. Just in case their site is ever down or whatever.

catarak:
Also, any ideas, from a UI/UX perspective, on the Lint Warning Sound and Text Output preference options? If a user doesn't have a screen reader and has these options on, he or she may be confused because nothing is happening. Is there a way to denote that you need a screen reader in order to use these features?

static/loadData.js
MathuraMG: Sure, I'll move both the files to the static folder

MathuraMG:
How about a heading within preferences that says 'Accessibility options' ? And maybe some way to read more on it? ( a drop down text or a link to some documentation)

Or since the lint warning beeps even without a screen reader, we can probably change the text from 'Hidden text output' to 'accessible canvas - used with screen reader' ?

@CleezyITP - what do you think of this?

CleezyITP:
@omnikrll is working on lint warning preferences atm. We talked about having a drop down for tone options and I think it would be great to link to further accessibility documentation. Picking up this work some more this weekend! 

I would love for the text output to be the default and have it be an option to turn it off for optimization...That option aside, I think that calling it '**accessible text-based canvas**' is most appropriate given the unique feature for screen reader and the research from American Foundation for the Blind (2016) regarding the issue of separate screen reader (SR) or "text­only" versions vs. accessibility of default site:

"Most users told us they do not appreciate having a separate SR or "text only" version of a site. They were concerned that they might not be getting the same thing as the default site, and that it might not be updated as regularly. Those who did appreciate the "SR version," because of its
name, also expected it would be specially designed for them (but in this case saw that as a
positive, assuring ease of use and appropriate content). If a separate version must exist, they
definitely recommended calling it a "text only" version, because this makes it a "universal design"
feature."

catarak:
Okay, makes sense. As much as I would like to have the text output on by default, I think it will be too sluggish sometimes for users not using screen readers. I like 'accessible text-based canvas'. For now, though this is an imperfect solution, can I just add a header for those options, something like "Accessible Enhancements", after I merge in this PR? Just to make something clear for users not using  screen reader, and then in another PR we can improve the UI/UX for accessible users.

CleezyITP:
I agree that it _could_ be sluggish at times which is why it should be made easy to turn off as possible. Maybe a pop-up that reads "The Processing Foundation supports accessibility, so enhanced accessibility options have been turned on by default. If you would like to turn these off to optimize performance CLICK HERE". This has been an issue at the forefront of efforts like this in the past and since the beginning of the work with p5.

MathuraMG:
Considering we can also turn on the text output using the play button (the hidden one), would it be as much of an issue if we keep it off in preferences?
Also, for now ( till we finish testing the text output the next weekend ), there is a clause I have added that says if there are more than 50 objects, only give a summary and the details will not be included. This does ensure that non screen reader users will not have too many speed issues - for now.

CleezyITP:
Ah @MathuraMG, thanks. 

Yes, sorry @catarak was confused by the 'default' word. As long as it is the default when screenreader users press play, no problem. I was just thinking it would be counter to the effort of accessibility if we made people search for the feature to turn it on.

catarak:
@MathuraMG I think the best option, for now is to keep the text output option out of preferences, and have the hidden play button for screen readers always have the text canvas. Seems like a happy medium, yes?

MathuraMG:
@CleezyITP , @catarak 
I understand the confusion it causes being in preferences as the output itself is not tangible for non screen reader users. But up until we get a bit more user feedback on it would it be okay to leave it in the preferences the following way - with a subtitle and default of 'off'.
screen shot 2016-08-19 at 2 41 34 pm

catarak:
@MathuraMG I'm okay with it. Want to add that to this PR and then I'll merge?

Issue #62, opened 2016-08-24
EditorAccessibility: items in {messages} array need key property


omnikrll:
The elements get painted to the DOM properly, but the following shows up in the console:

"Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `EditorAccessibility`. See https://fb.me/react-warning-keys for more information.
    in p (created by EditorAccessibility)"

PR #63, opened 2016-08-24
add key property to lint message elements

omnikrll:
fixes https://github.com/processing/p5.js-web-editor/issues/62

### DIFF
--- a/client/modules/IDE/components/EditorAccessibility.js
+++ b/client/modules/IDE/components/EditorAccessibility.js
@@ -9,7 +9,7 @@ class EditorAccessibility extends React.Component {
     if (this.props.lintMessages.length > 0) {
       for (let i = 0; i < this.props.lintMessages.length; i++) {
         messages.push(
-          <li>
+          <li key={i}>
             {this.props.lintMessages[i].severity} in line
             {this.props.lintMessages[i].line} :
             {this.props.lintMessages[i].message}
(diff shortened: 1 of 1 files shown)


client/modules/IDE/components/EditorAccessibility.js
```diff
@@ -9,7 +9,7 @@ class EditorAccessibility extends React.Component {
     if (this.props.lintMessages.length > 0) {
       for (let i = 0; i < this.props.lintMessages.length; i++) {
```
catarak: I didn't notice this earlier, but these days in JS Land I think it's better practice to use forEach than a for loop.

catarak:
i've been meaning to fix this, so ty!

client/modules/IDE/components/EditorAccessibility.js
```diff
@@ -7,18 +7,19 @@ class EditorAccessibility extends React.Component {
...
+            {lintMessage.message}
           </li>
         );
-      }
+      });
     } else {
       messages.push(
-        <p tabIndex="0"> There are no lint messages </p>
+        // no array index so let's just call it 0
+        <p tabIndex="0" key={0}> There are no lint messages </p>
```
omnikrll: Is there a reason this is a <p> and not a <li>, if it's going to go inside a <ul>?

catarak: @MathuraMG do you know why? Does this affect the output of the screen reader?

MathuraMG: I initially did it in hope that the screen reader would read it differently, but it didn't. I believe I forgot to fix it back.
We can change it to `<li>`

PR #70, opened 2016-08-25
Accessibility changes to preferences and toolbar

MathuraMG:
This PR -
* changes the options to radio inputs - this is so screen readers can identify which option is selected. There are no visual or functional changes. 
* changed the labels on the toolbar based on user feedback

(cc @CleezyITP)

### DIFF
--- a/client/modules/IDE/components/Preferences.js
+++ b/client/modules/IDE/components/Preferences.js
@@ -150,47 +150,85 @@ class Preferences extends React.Component {
         <div className="preference">
           <h4 className="preference__title">Autosave</h4>
           <div className="preference__options">
-            <button
-              className={autosaveOnClass}
+            <input
+              type="radio"
               onClick={() => this.props.setAutosave(true)}
               aria-label="autosave on"
-            >On</button>
-            <button
-              className={autosaveOffClass}
+              name="autosave"
+              id="autosave-on"
+              className="preference__radio-button"
+              value="On"
+              checked={this.props.autosave}
+            />
+            <label htmlFor="autosave-on" className={autosaveOnClass}>On</label>
+            <input
+              type="radio"
               onClick={() => this.props.setAutosave(false)}
               aria-label="autosave off"
-            >Off</button>
+              name="autosave"
+              id="autosave-off"
+              className="preference__radio-button"
+              value="Off"
+              checked={!this.props.autosave}
+            />
+            <label htmlFor="autosave-off" className={autosaveOffClass}>Off</label>
           </div>
         </div>
         <div className="preference">
           <h4 className="preference__title">Lint Warning Sound</h4>
           <div className="preference__options">
-            <button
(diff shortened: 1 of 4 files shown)


catarak:
There's a little CSS trick you can do so that you don't need the `preference__option--selected` class, if you put the label after the radio button in the HTML:

```scss
.preference__option {
    color: $light-inactive-text-color;
}

.preference__value:checked + .preference__option {
    color: $light-primary-text-color;
}
```

MathuraMG:
A quick question - would this be to replace `preference__option--selected` throughout or specifically for the option using radio button ? 
Cause we are using `preference__option--selected` for all the preferences. I could make the changes in a separate PR then ?

catarak:
I meant specifically for the options using a radio button, as I think it will clean up the code in the Preferences JSX component. 

Should the tabs/spaces option also be radio buttons?

MathuraMG:
Got it!
My bad, I missed that - I'll make the changes for that as well

client/modules/IDE/components/Preferences.js
```diff
@@ -143,54 +112,112 @@ class Preferences extends React.Component {
             <h6 className="preference__label">Increase</h6>
           </button>
           <div className="preference__vertical-list">
-            <button className={preferencesSpaceOptionClass} onClick={this.props.indentWithSpace} aria-label="indentation with space">Spaces</button>
-            <button className={preferencesTabOptionClass} onClick={this.props.indentWithTab} aria-label="indentation with tab">Tabs</button>
+            <input
+              type="radio"
+              onClick={this.props.indentWithSpace}
```
catarak: In the JS console, I'm getting the following error: 
Warning: Failed form propType: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`. Check the render method of `Preferences`.

I think this is because this should be `onChange` instead of `onClick`.

client/modules/IDE/components/Preferences.js
```diff
@@ -143,54 +112,112 @@ class Preferences extends React.Component {
...
+              name="indentation"
+              id="indentation-space"
+              className="preference__radio-button"
+              value="Spaces"
+              checked={!this.props.isTabIndent}
+            />
+            <label htmlFor="indentation-space" className="preference__option">Spaces</label>
+            <input
+              type="radio"
+              onClick={this.props.indentWithTab}
```
catarak: In the JS console, I'm getting the following error: 
Warning: Failed form propType: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`. Check the render method of `Preferences`.

I think this is because this should be `onChange` instead of `onClick`.

PR #83, opened 2016-09-01
add preview; output focus

MathuraMG:
Based on user feedback - [Beta User Testing](https://docs.google.com/spreadsheets/d/1VWlJQrEKwQFvZjXJrPiOWz59Sanv9eK58RRvbo1e83A/edit#gid=732808470) - this PR contains the following changes -
* Add a preview to the sound in lint warning
* If output text is On, the focus moves to the text output when user plays the sketch
* Add a heading to the output text

(cc @CleezyITP)

### DIFF
--- a/client/modules/IDE/components/TextOutput.js
+++ b/client/modules/IDE/components/TextOutput.js
@@ -2,7 +2,7 @@ import React from 'react';
 
 class TextOutput extends React.Component {
   componentDidMount() {
-
+    document.getElementById('canvas-sub').focus();
   }
   render() {
     return (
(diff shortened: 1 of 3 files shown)


client/modules/IDE/components/TextOutput.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 
 class TextOutput extends React.Component {
   componentDidMount() {
-
+    document.getElementById('canvas-sub').focus();
```
catarak: instead of using `document.getElementById()`, use [React refs](https://facebook.github.io/react/docs/more-about-refs.html)

PR #94, opened 2016-09-12
detect infinite loop

### LINKED ISSUE #77: Catch infinite loops before they crash editor
If there are infinite loops, the whole editor will crash. The editor should check for this before inserting the code into the iframe.

yining1023:
Detecting infinite loop: whenever the code in the editor changes, it stops the sketch from running, checks for any infinite loop, then either shows error messages below the code, or restart the sketch.

When there is an infinite loop, and user tries to click the run button, show one alert.

**Issues:** Fixes #77 

screen shot 2016-09-14 at 7 48 10 pm

screen shot 2016-09-14 at 7 53 06 pm


yining1023:
I just fixed some conflicts and force pushed.

catarak:
I have the merge conflicts fixed locally, so don't worry about that, but I found an issue if you put an infinite loop inside of the `draw()` loop:
screen shot 2016-09-20 at 11 43 36 pm
It keeps appending the error!

client/styles/components/_editor.scss
```diff
@@ -80,4 +80,12 @@
...
   }
-}
\ No newline at end of file
+}
+
+.lint-error {
+  font-family: Inconsolata, monospace;
+  font-size: 100%;
+  background: #FFBEC1;
+  color: red;
```
catarak: Could this be the SCSS variable $console-error-color?

client/styles/components/_editor.scss
```diff
@@ -80,4 +80,12 @@
...
     display: block;
   }
-}
\ No newline at end of file
+}
+
+.lint-error {
+  font-family: Inconsolata, monospace;
+  font-size: 100%;
+  background: #FFBEC1;
```
catarak: Also, could this be the same color except with an alpha?

yining1023: Yes! and yes.

yining1023:
Sorry about the 'appending error in the draw() function' error. I just fixed it. And I used the SCSS variable $console-error-color on color and background-color.
screen shot 2016-09-22 at 5 15 49 pm

client/modules/IDE/components/Editor.js
```diff
@@ -47,23 +49,30 @@ class Editor extends React.Component {
...
-          }
-        })
+        }
       }
     });
 
     this._cm.on('change', debounce(200, () => {
       this.props.setUnsavedChanges(true);
       this.props.updateFileContent(this.props.file.name, this._cm.getValue());
+      this.checkForInfiniteLoop(debounce(200, (infiniteLoop, prevs) => {
```
catarak: you don't need this debounce here--it's actually not doing anything!

client/modules/IDE/components/Editor.js
```diff
@@ -47,23 +49,30 @@ class Editor extends React.Component {
...
           });
-          if (this.props.lintMessages.length > 0 && this.props.lintWarning) {
-            this.beep.play();
-          }
-        })
+        }
       }
     });
 
     this._cm.on('change', debounce(200, () => {
```
catarak: I noticed that typing in the editor is a little sluggish/lags a bit from what you're typing. If you make this debounce a little longer (I tried 1000ms), it's smooth. I think this is because it's too slow rebuilding the iframe over and over again.

catarak:
Good news: it works!
Bad news: I noticed two things,
1. it's a little sluggish to type, i think because the iframe is getting rebuilt so often. I added a comment with one way to fix it.
2. Because you're building a iframe separate from `PreviewFrame`, I noticed that errors are getting added to the console twice:
screen shot 2016-09-23 at 6 52 30 pm
This is not the end of the world, but do you think there's a way to either run the loop protect script in PreviewFrame, or suppress the output of the loop protect iframe? To me, the first option is preferable.

yining1023:
- You are right, we don't need that debounce. I took it off.
- I changed the `.on 'change'` debounce to 1000ms, also I kept only one  loop protect iframe, and rewrote its content every time instead of rebuilding an iframe every time.
- I suppressed the output error of the loop protect iframe. Now only the previewFrame is printing the error. I agree that it's definitely better to run the loop protect script in PreviewFrame. I chose the second option because it's a quick fix. But I'm very open to implement the first option. Just let me know.
- In addition, if there is an infinite loop, the play button will be disabled instead of having an alert window popping out. The color of the play button will be gray too. I think in this way, we might have a better user experience.

screen shot 2016-09-24 at 12 12 58 am

screen shot 2016-09-24 at 12 16 30 am

catarak:
just tried to run this locally on my computer and i can't get it to work... is it working for you?

yining1023:
Um... It's working on my computer locally. Did you see any exceptions or errors?

catarak:
Okay, I pinpointed the issue: it seems to not be working for me if the project has not been saved for the first time.

yining1023:
Thank you for pinpointing the issue. It didn't happen for me. It works when the project is not saved too.

catarak:
Hmmm, okay, we'll look at it tomorrow 😄

yining1023:
😆I'm sure we are almost there. See you tomorrow, and thanks for being so
patient!

On Tue, Sep 27, 2016 at 9:27 PM, Cassie Tarakajian <notifications@github.com

lmccart:
you guys are awesome

ymittal:
Looks like it's still happening (at least on brave browser). I've forgotten to save my sketch many times and run into an infinite loop and lost progress ...

PR #118, opened 2016-10-01
fix for serial port console crashing

lmccart:
I added a throttle function that will only let messages post to console at max of one every 100ms. This eliminates the crashing we were seeing previously. My question is whether 100ms is sufficiently small of an interval? Or would it be ok to increase this a bit to say 250ms? cc @tigoe


lmccart:
oops I realized there were a few other css changes that got in there before I realized I was working on the wrong branch for styling. sorry @catarak! ok to just merge these little ones in now (the rest will be coming in a PR later)? otherwise I can try to undo them.

tigoe:
I think it might cause more problems than it solves. Throttling will slow
down the time between a message arriving and the user seeing it, won't it?
Or does this just slow down the console but not the sketch? Is the patch
online so we can try it?

lmccart:
Throttling doesn't deliver messages with a delay but instead drops some messages so no more than 1 per 100ms comes through.

The issue in general is that the protocol that receives the native console messages and posts them to our console frame (window.postMessage) seems to just have a limit on how fast it can process messages. So there's not really anything to "fix", as far as I can tell, besides reduce the number of messages coming through. But I'm very open to other ideas if anyone has them.

tigoe:
OK, that makes more sense. Is the patch live to try? I can give it a test if you like.

tigoe:
Not sure if the current editor has the patch or not, but here's a more detailed report, hopefully more useful.

* tested with this sketch: http://alpha.editor.p5js.org/projects/BJidS43T
* running p5.serialserver in the bckground, of course
* tested using the Arduino sketch called AnalogReadSerial (Arduino IDE File Menu, Examples -> Basics -> AnalogReadSerial)
* Tested using an Uno, an Arduino 101, and a MKR1000. Same results on all.

With the sketch as is, responsiveness is real-time. Add a println(sensorValue); after line 32 and it slows to a crawl. 

Response is the same whether console is minimized or not.

I'm gonna try from your branch now, Lauren.

lmccart:
Sorry, it hasn't been merged in yet.. waiting for @catarak to review, I think she's in on wednesdays. Are you doing serial this week? I can probably go ahead and merge this one if so. If you want to give it a test first, that'd be great.

tigoe:
Yeah, Synthesis Day is on Friday, so we're in a testing frenzy. I can test on a dev server if you have one running, but I haven't got Mongo et al installed on  my machine, unfortunately.

vanevery:
In p5.serialcontrol I did this to control the flow to the console.  It works pretty well, might be worth a shot in the editor:

`
var consoleBuffer = [];
var lastConsoleLogTime = Date.now();
var LOGWAIT = 500;

function seriallog(txt) {			
	consoleBuffer.push(txt);
	if (lastConsoleLogTime + LOGWAIT < Date.now()) {
		if (serialConsole.elt.value.length >= 800)
		{
			serialConsole.elt.value = serialConsole.elt.value.substring(400);
		}
		serialConsole.elt.value += consoleBuffer.join("\n");
		serialConsole.elt.scrollTop = serialConsole.elt.scrollHeight;
				
		lastConsoleLogTime = Date.now();
		consoleBuffer.length = 0;
	}
}
`

tigoe:
@vanevery, the Alpha 4 version of P5.serialcontrol was illuminating. Might shed light on the problem beyond the console.

I ran it for a few minutes with its console enabled. When I downloaded and opened the SerialAnalogInput sketch as a local file in the browser, it ran at full speed, unaffected by the console enabled in the electron app.  But when I ran the sketch in the web editor, it ran considerably slower, even with no println in the sketch to slow it down. I thought maybe it could be auto-refresh, but that was turned off.

is there anything else the web editor is doing to monitor the sketch, or that the p5.serialport library is doing to communicate with the p5.serialcontrol server, that might slow things down?

@lmccart, sorry I haven't seen Jen, Mathura, or Yining to test your latest, but I will as soon as I see them.

vanevery:
Thanks for testing glad that my snippet seems to have worked in the p5.serialcontrol app.  Just as an FYI, that app uses p5 and the p5.serialport library just like we do in the editor.

catarak:
Finally got a chance to test this. There's a little CSS bug
screen shot 2016-10-04 at 7 02 28 pm

And also, when I left throttle at 100ms, it still causes the editor to lock up for me, but it was okay at 250ms.

tigoe:
I tested with Jen and Yining, and we had ther same results, still looking up at 100ms. Cassie, is it possible to post a compiled version online at 250ms I can try out?


Tom

catarak:
Since it isn't going to make anything worse, I'll just merge this in as it as with 100ms throttle time, and let me know how it is. If it's still slow, I can make the time longer and redeploy.

tigoe:
Bad news: the throttle seems to slow it down even when there is no println(). Sorry to say I think it made things work. I tested the SerialAnalogInput.js sketch in the editor and in the browser without the editor, and it's real-time without the editor, 3-5 seconds behind with the editor.

vanevery:
@catarak here is a test sketch http://alpha.editor.p5js.org/projects/BJidS43T
Here is the p5.serialcontrol alpha 4: https://github.com/vanevery/p5.serialcontrol/releases/tag/0.0.4

catarak:
I just pushed to production a version of the console that batch delivers the messages instead of throttling the console messages, as @vanevery suggested. It seemed to be working okay for me, but let me know after some experimentation.

vanevery:
Seems to be working great!  I am not experiencing any delays or crashes with println in the serialEvent on this sketch: http://alpha.editor.p5js.org/projects/BJidS43T

tigoe:
It's working fine with an Uno or a 101, but anything based on the M0 chip, it's still deadly slow. Not sure why. For what it's worth, there is something in p5.min.js on line 4 that's doing a console log of the serial data, and that could be part of the problem. When I open the sketch locally (no editor), I get stuff in the JS console from line 4 of p5.min.js, and I shouldn't get anything, unless I println myself, right?

lmccart:
hm that's weird, can you try linking to [p5.js](https://cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.4/p5.js) instead to identify the exact line number? (they're all mashed together for the min version)

tigoe:
trying it now with local files...

tigoe:
Found it, line 9769 in p5.js. commenting that out speeds things up a bit.

Still having a problem where the ARM chips are able to fill the serial buffer faster than the sketch can empty it. Not sure why, yet. I'm going to try it in node by itself, see what I get.

tigoe:
OK, here's a data point: it's also slow using an Arduino Leonardo (which I know Dano is using). The Leonardo and the M0 boards are all doing native USB rather than going through an FTDI chip like the Uno, which makes me think there's a USB speed issue at hand.

running a bare node.js serialport script, I can get all the boards to work in realtime. So it might have something to do with how p5.serialserver's handling things as well, @vanevery. Testing with ws next, will let you know.

tigoe:
I've got good news and bad news.

The good news is that it's not the console anymore. I  can get realtime performance from the web editor and console when talking to it from node.js.

The bad news is that I can't do that from p5.serialport.

I got a basic serial to ws server working here: https://github.com/tigoe/NodeExamples/tree/master/SerialToWsServer  and I got a basi ws client sketch working here: http://alpha.editor.p5js.org/projects/rJDfv8N0. But it only works with ASCII strings, lacks the functionality @vanevery has in p5.serialport. However, it can handle the data in realtime.  So @vanevery, I think the issue is in either the server or the client lib, sorry to say.

PR #124, opened 2016-10-06
updating editor fonts

lmccart:
* removing avenir, replacing with montserrat
* updating font size
* adding console symbol in console header


catarak:
Confused, I thought Avenir was the official font and Montserrat was the fallback? Or is this a change from the new designs?

lmccart:
Yeah, sorry for the confusion, this is an update from the original... the Avenir was removed because the license cost is based on traffic and becomes prohibitively expensive at scale.

p5js-ide-styles-foundation-pt-01 2x

catarak:
@lmccart did you mean to close this? Should I still merge this in?

lmccart:
I pushed some other commits to that branch so it's in a kind of in between state right now. I'll close this now and reopen when it's in a stable point. And I'll be sure to create new branches in the future if prs are still open so there's less confusion, sorry about that.

Issue #127, opened 2016-10-07
Button (at upper-right) will be blocked by vertical scroll bar


zhuoxi:
![screen shot 2016-10-07 at 1 37 16 am](https://cloud.githubusercontent.com/assets/8460819/19179778/e922fb82-8c2e-11e6-98c8-1267d312e5e9.png)

▲ No vertical scrollbar, button at the upper-right corner is accessible

![screen shot 2016-10-07 at 1 37 37 am](https://cloud.githubusercontent.com/assets/8460819/19179780/ed95fe6c-8c2e-11e6-9223-7988374932a0.png)

▲ Vertical scrollbar comes out, blocking the button

![screen shot 2016-10-07 at 1 38 28 am](https://cloud.githubusercontent.com/assets/8460819/19179782/efa68c08-8c2e-11e6-8c18-d559ee6e37a5.png)

▲ Button is hard to access

PR #128, opened 2016-10-07
fixed upper-right editor button being blocked by vertical scroll bar

zhuoxi:
moved the upper-right button in editor a little to the left, so that
vertical scroll bar will not block this button

#127

### DIFF
--- a/client/styles/components/_editor.scss
+++ b/client/styles/components/_editor.scss
@@ -80,7 +80,8 @@
   }
   position: absolute;
   top: #{5 / $base-font-size}rem;
-  right: #{5 / $base-font-size}rem;
+  // right: #{5 / $base-font-size}rem;
+  right: #{5 / $base-font-size * 4}rem; // move a little left to avoid vertical scroll bar
   z-index: 1;
 }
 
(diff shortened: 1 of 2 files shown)


client/styles/components/_editor.scss
```diff
@@ -80,7 +80,8 @@
   }
   position: absolute;
   top: #{5 / $base-font-size}rem;
-  right: #{5 / $base-font-size}rem;
+  // right: #{5 / $base-font-size}rem;
+  right: #{5 / $base-font-size * 4}rem; // move a little left to avoid vertical scroll bar
```
catarak: Just for the sake of readability, could this be
```scss
right: #{20 / $base-font-size}rem;
```

client/styles/components/_editor.scss
```diff
@@ -80,7 +80,8 @@
   }
   position: absolute;
   top: #{5 / $base-font-size}rem;
-  right: #{5 / $base-font-size}rem;
+  // right: #{5 / $base-font-size}rem;
```
catarak: There's no need to leave this in commented out

PR #130, opened 2016-10-07
fixed upper-right button being blocked by vertical scrollbar

zhuoxi:
Moved the upper-right editor option button a little to the left, so that once vertical scrollbar appears, this button will not be blocked by the scrollbar.  #127 

Before:

image

This fix:

image

### DIFF
--- a/client/styles/components/_editor.scss
+++ b/client/styles/components/_editor.scss
@@ -80,8 +80,7 @@
   }
   position: absolute;
   top: #{5 / $base-font-size}rem;
-  // right: #{5 / $base-font-size}rem;
-  right: #{5 / $base-font-size * 4}rem; // move left to avoid vertical scroll bar
+  right: #{20 / $base-font-size}rem; // move left to avoid vertical scroll bar
   z-index: 1;
 }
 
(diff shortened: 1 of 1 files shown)


client/styles/components/_editor.scss
```diff
@@ -80,7 +80,8 @@
   }
   position: absolute;
   top: #{5 / $base-font-size}rem;
-  right: #{5 / $base-font-size}rem;
+  // right: #{5 / $base-font-size}rem;
+  right: #{5 / $base-font-size * 4}rem; // move left to avoid vertical scroll bar
```
catarak: For readability, could this be
```scss
    right: #{20 / $base-font-size }rem;
```

client/styles/components/_editor.scss
```diff
@@ -80,7 +80,8 @@
   }
   position: absolute;
   top: #{5 / $base-font-size}rem;
-  right: #{5 / $base-font-size}rem;
+  // right: #{5 / $base-font-size}rem;
```
catarak: There's no need to leave this in, commented out. Just delete it!

PR #151, opened 2016-10-16
Fix lint warning sound

MathuraMG:
The code for lint warning sound  seems to not be working right after [this change](https://github.com/processing/p5.js-web-editor/pull/94/commits/43052cb675fa0bc500467f5c7c6643f3636493ae?diff=split#diff-86c03a3ffefe63a80376f3a2db405fa1L50).
@yining1023 - just wanted to ask what the context for this change was. Would it be okay to revert it?
Please let me know if I'm missing something!

### DIFF
--- a/client/modules/IDE/components/Editor.js
+++ b/client/modules/IDE/components/Editor.js
@@ -49,19 +49,17 @@ class Editor extends React.Component {
       gutters: ['CodeMirror-lint-markers'],
       keyMap: 'sublime',
       lint: {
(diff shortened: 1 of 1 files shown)


yining1023:
Yes. Before I used lint code to append error messages to the code. But now we are only showing error messages in the console. So we don't need this change anymore. Please revert it. Sorry about this!

MathuraMG:
Awesome, thank you @yining1023 !

PR #153, opened 2016-10-18
fixes #148

### LINKED ISSUE #148: Automatically saving updated sketch name
When a sketch is saved, then I change the sketch name, it seems like I need to click on the "save" button again to save the new sketch name. Otherwise if I open "My Sketches" list, my sketch has the old name, and if I close the "My Sketches" list, my sketch will go back to its old name. 

Perhaps we could automatically save the sketch after user update the name?

yining1023:
fixes #148 
This PR saves project after a user changes the sketch name.

### DIFF
--- a/client/modules/IDE/components/Toolbar.js
+++ b/client/modules/IDE/components/Toolbar.js
@@ -103,7 +103,9 @@ class Toolbar extends React.Component {
             onBlur={() => {
               this.validateProjectName();
               this.props.hideEditProjectName();
-              this.props.saveProject();
+              if (this.props.project.id) {
+                this.props.saveProject();
+              }
             }}
             onKeyPress={this.handleKeyPress}
           />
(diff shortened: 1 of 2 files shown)


catarak:
I'm not sure if editing the project name should save the entire project, though I'm not entirely sure this is a problem either... thoughts? If a user has autosave off, this may be confusing behavior. 

I just looked at CodePen to see how they handle this, and I think this could be a good solution. When the user edits the title and hits enter to finish editing, a toast appears saying "Pen saved", if autosave is on or off.

yining1023:
I agree. For now, same as CodePen, when the user edits the title and hit enter, a toast appears saying "Project saved", regardless if autosave is on or off.

But I noticed that in CodePen, when a sketch is unsaved(not created, doesn't have a sketch id or url yet), and when a user changes the sketch name, the entire project won't be saved. So I changed it to the following:

sketch is unsaved(not created) --> editing the sketch name --> saving the name, but not the entire project
sketch is saved(created) --> editing the sketch name --> saving the entire project

Issue #158, opened 2016-10-19
Adding a circle icon near the filename when the file is unsaved


yining1023:
If the sketch has never been saved, and a user has edited a sketch, generate a toast saying that the project has not yet been saved.

catarak:
Regarding the discussion in #146, I think it's better to not show a toast for this and rely on other visual cues instead. I do still think we need #159, and to leave the Save toasts that exist now.

catarak:
The functionality has been added here, we just need to replace the asterisk with a circle icon:
<img width="1277" alt="aebdb85e-96bc-11e6-81e1-441c0cf48348" src="https://cloud.githubusercontent.com/assets/6063380/20199314/a06ddb3c-a777-11e6-843a-945d10261ecc.png">

PR #177, opened 2016-11-02
Added a star icon for unsaved file name, added 'Saved:  xx time ago'

### LINKED ISSUE #157: Displaying timestamp that says 'Last saved: 1 minute ago'
Displaying a timestamp when the project was last saved. e.g. Last saved: 1 minute ago, next to the sketch name.

yining1023:
fixed #157 #158 
screen shot 2016-11-02 at 4 26 33 pm


client/styles/components/_editor.scss
```diff
@@ -79,8 +79,8 @@
     @extend %icon;
   }
   position: absolute;
-  top: #{5 / $base-font-size}rem;
-  right: #{20 / $base-font-size}rem; // move left to avoid vertical scroll bar
+  top: #{0 / $base-font-size}rem;
```
catarak: this can just be `top: 0`

catarak:
I noticed that if you leave the editor running and don't touch it, the last saved text won't get updated as time passes. I think this is because it doesn't know it has to rerender. I think the way to fix this is:
1. Make the time display into its own component (so that the entire `Editor` component doesn't have to rerender).
2. In this component, call a function periodically that will force it to rerender, or update a prop periodically so that it will rerender.
I know that's fairly high level, so I'm happy to go into detail with you about how to implement this.

catarak:
I also noticed that this causes a little styling weirdness:
screen shot 2016-11-02 at 10 06 43 pm
If you look closely, you can see that the `Sidebar` component has a `border-top`, which looks kind of funny on its own. I would either (1) add a right border so it looks okay for this merge, or (2) implement this how it looks in the design. If you want to do (1), then the new design can be implemented in another PR.

yining1023:
Got it. Thanks! I will do (1) (add a right border), and I'm excited about making a new component. 😄

client/modules/IDE/components/Timer.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  componentWillMount() {
+    if (this.interval) {
+      clearInterval(this.interval);
+    }
+  }
+
+  componentDidMount() {
+    this.interval = setInterval(() => this.forceUpdate(), 10000);
```
yining1023: Forcing timer component to re-render every 10 seconds.

catarak: 🙏

client/modules/IDE/components/Timer.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  showSavedTime() {
+    if (Math.abs(moment().diff(this.props.projectSavedTime)) < 10000) {
```
catarak: I think this makes sense. Autosave runs every 30 seconds and is on by default, so I think it makes sense to have some time steps in between 0 and 30 seconds.

client/modules/IDE/components/Timer.js
```diff
@@ -0,0 +1,46 @@
...
+
+class Timer extends React.Component {
+  constructor(props) {
+    super(props);
+    this.showSavedTime = this.showSavedTime.bind(this);
+  }
+
+  componentWillMount() {
+    if (this.interval) {
+      clearInterval(this.interval);
```
catarak: instead of clearing the interval here, maybe it makes more sense to clear the interval in [componentWillUnmount](https://facebook.github.io/react/docs/react-component.html#componentwillunmount)?

catarak:
This PR has also alerted me to the fact that there's a bug with `unsavedChanges`:
(1) edit one file, unsavedChanges is true
(2) switch to a new file, unsavedChanges becomes false, despite the fact that there are unsaved changes
But this is a separate issue that you don't have to fix in this PR 😄

client/modules/IDE/components/Timer.js
yining1023: Total agree. Just pushed the new changes.

client/styles/components/_timer.scss
```diff
@@ -0,0 +1,8 @@
+.timer__saved-time {
+  @include themify() {
+    color: getThemifyVariable('inactive-text-color');
+  }
+  font-size: #{12 / $base-font-size}rem;
+  padding-right: #{30 / $base-font-size}rem;
+  float: right;
```
catarak: This is the last thing, and this this PR is good to merge!

kaganjd:
@yining1023 and @catarak, I'm still not seeing the star next to unsaved projects even if I'm logged in. Is this what you see?
screen shot 2016-11-16 at 6 35 06 pm

yining1023:
I think if you type something in the editor, the star will appear.

kaganjd:
Ohhh! Got it, thanks.

PR #184, opened 2016-11-04
Generate a two-word project name

### LINKED ISSUE #101: Sketch Naming
A source of confusion for students was naming sketches (when to name them, many sketches with the same name, etc).   Again, this may need to wait until after the design review, but here are some thoughts that came up in class.
1.  Sketch names for a user account have to be unique.  This would also help with #68 so that each sketch can have a url like `/username/sketch-name/`.
2. When you "duplicate" or "save as" you are prompted to name the sketch (with the previous name filled in already.)
3. When you make a "new" sketch, you are prompted to name the sketch or it is named with something automatic perhaps date and time related?
4. Whenever you name a sketch, a save action is performed?  I'm not sure about this, but this was the expectation of some users.   Maybe irrelevant if we force naming and save on the creation of a new sketch anyway.

kaganjd:
Closes #101 
screen shot 2016-11-03 at 9 33 13 pm


client/modules/IDE/reducers/project.js
```diff
@@ -1,7 +1,10 @@
 import * as ActionTypes from '../../../constants';
+const generate = require('project-name-generator');
```
catarak: To ES6-ify this line:
```javascript
import generate from 'project-name-generator';
```

client/modules/IDE/reducers/project.js
```diff
@@ -1,7 +1,10 @@
...
+
+const generatedName = generate({ words: 2 }).dashed;
 
 const initialState = {
-  name: 'Hello p5.js'
+  name: `${generatedName}`
```
catarak: You should move the function call into the initial state, i.e. change this to 
```javascript
const initialState = {
    name: `${generate({ words: 2}).dashed}`
};
```
As it is now, the following happens, which I think is a bug:
1. User navigates to the editor and starts working on a project
2. User saves the project.
3. User clicks "New"
4. The new project has the same name as the previous project.

catarak:
I'm not saying that I don't like it, but is there any reason you decided that the name should be dashed rather than spaced? I'm not sure that I have an opinion about it, just curious. Also, do you think that the first work should be capitalized? Am I thinking about this too hard?

kaganjd:
1. i tried spaced and didn't like it because with 0 capitalized words, they just looked like words floating-- not a project name-- so i went with dashed instead.
2. i think spaced with first word capitalized is a good idea. i'm still a little worried it won't read as a generated name, but tbd!
3. no.

client/modules/IDE/reducers/project.js
kaganjd: i tried this and found that moving the function call doesn't fix the problem, i think because newProject() doesn't actually create a new project; it just returns the initialState name. this was fine when all new sketches had the same name. hmm, i'll have to think about this a little more... has there been discussion of "untitled" as the sketch name prior to saving? and then the generated name appears on save?

catarak: Oh, right! The way to make this work is to make initialState into a function rather than a variable. See the files reducer for an example of how to do this.

catarak:
I'm leaning towards first word capitalized + spaced rather than dashed, I think.

PR #194, opened 2016-11-09
Change text output content

MathuraMG:

* Render sketch when text output status changes
* Make text output less verbose - change file structure ( moved the helper functions for the interceptor to a separate file )
* Include area; coordinates in text output

Tested with the [following example codes.](https://github.com/MathuraMG/monkeyPatchTest/tree/master/examples)


catarak:
This PR looks 👍, but just a thought: do you think, at some point, it would make sense to package all of this interceptor code into its own Github repository? Then this project could link to the interceptor functions as a submodule. Just curious for some thoughts about this.

cc @lmccart @CleezyITP

MathuraMG:
@catarak That sounds like a great idea!
We are talking about adding the alternate types of text outputs as well and it feels like this would be a good option. 
Curious to hear everyones thoughts!

lmccart:
Definitely agree. I couldn't believe there weren't already modules like this floating around when we did a search earlier.

Issue #100, opened 2016-09-16
sketches directory for non-existent user


shiffman:
If you go to a link like:

https://alpha.editor.p5js.org/blahblahblah/sketches

You cannot exist the sketches list.  And perhaps a message should say "this user does not exist" or something. . .

Issue #149, opened 2016-10-16
"Open" to "Examples" when a user is not logged in


yining1023:
When a user is not logged in, "Open" links to the sketches of a user called p5, which would be a few example sketches.
Perhaps we could 
- change "Open" to "Examples". And keep the "Examples" after a user logs in.
- add more examples to this sketch list.

lmccart:
yes, this makes much more sense. with the desktop editor, there was a task that would update the example set by pulling from the website repo on release build, I wonder if there is a way to do something similar with the web editor so we can be sure the examples stay up to date and correct.
https://github.com/processing/p5.js-editor/blob/master/gulpfile.js#L248

catarak:
I don't think it would be too hard to set up a similar task! We could just write a node script, run it with npm, and then set up a cron job that runs this task periodically.

PR #198, opened 2016-11-15
Fixed #158 and #100

### LINKED ISSUE #158: Adding a circle icon near the filename when the file is unsaved

epiqueras:
<h2> #158 </h2>

Changed unsaved changes asterisk to an svg circle.

<h2> #100 </h2>

Unmatched routes are handled by react-router on the client side
and a single wildcard route on server.routes.js renders the index
html.
When the /:username/sketches route is matched and the username is not
valid, the user will be redirected to the index route and a toast
will explain what happened. When the username is 'p5' (default when
logged out) it will show all sketches. Maybe this should be changed
to just public or 'local' sketches?


catarak:
Woohoo! In the future, could individual tickets be individual pull requests? It's just easier to code review that way 😄

server/controllers/project.controller.js
```diff
@@ -89,14 +89,28 @@ export function getProjects(req, res) {
 
 export function getProjectsForUser(req, res) {
   if (req.params.username) {
-    User.findOne({ username: req.params.username }, (err, user) => {
-      Project.find({user: user._id}) // eslint-disable-line no-underscore-dangle
+    // Should we return all sketches when the username is p5? Maybe just public ones?
+    if (req.params.username === 'p5') {
```
catarak: In production, the idea is that the "p5" user will have a bunch of projects intended to be examples (#149). So we don't need this 😄

server/controllers/project.controller.js
```diff
@@ -89,14 +89,28 @@ export function getProjects(req, res) {
...
         .select('name files id createdAt updatedAt')
         .exec((err, projects) => {
-          res.json(projects);
+          return res.json(projects);
         });
-    });
+    } else {
+      User.findOne({ username: req.params.username }, (err, user) => {
+        if (!user) {
+          return res.send(['user not found']);
```
catarak: This should return a 404 error, i.e. look something like
```javascript
return res.status(404).json({message: 'User with that username does not exist.'});
```

server/routes/server.routes.js
```diff
@@ -1,48 +1,13 @@
...
-});
-
-router.route('/about').get((req, res) => {
-  res.send(renderIndex());
-});
-
-router.route('/:username/sketches').get((req, res) => {
+// Wildcard route.
+// 404s are handled by the client (react-router) and redirect to the index route.
+router.route('*').get((req, res) => {
```
catarak: Make sure that this gets imported last of all the other routers in`server/server.js`! Otherwise I think it will override some custom routes.

Also this wildcard route makes me nervous, though it maybe be totally unfounded. I guess I just prefer to whitelist things and understand what's happening rather than allow all options and blacklist from there. Again, could totally not make sense. 

But this also begs a UX question: is it better to redirect to the root url or render a 404 page? If the latter, I think the 404 page should be rendered server side. I'm open to both options.

client/modules/IDE/components/SketchList.jsx
```diff
@@ -20,6 +21,15 @@ class SketchList extends React.Component {
     document.getElementById('sketchlist').focus();
   }
 
+  componentDidUpdate() {
+    if (this.props.sketches[0] === 'user not found') {
```
catarak: This works, but not conventional IMO. I would prefer to render a 404 page over this (it can be fugly for now and we can design a nice one.)

server/views/404Page.js
```diff
@@ -0,0 +1,13 @@
+export function render404Page() {
```
epiqueras: TODO: Make this page better. Maybe play a random sketch every time it's rendered?

server/server.js
```diff
@@ -88,6 +89,20 @@ app.get('/', (req, res) => {
   res.sendFile(renderIndex());
```
epiqueras: This route is not being matched since it's already declared in server.routes. Even if it did get matched it would throw an error because you can't pass renderIndex() to sendFile().

client/modules/IDE/components/SketchList.jsx
```diff
@@ -20,6 +21,13 @@ class SketchList extends React.Component {
     document.getElementById('sketchlist').focus();
   }
 
+  componentDidUpdate() {
+    if (this.props.sketches.length === 0) {
```
epiqueras: The server now returns a 404 error when the user is not found. This means that the response is not being passed as a prop anymore so all I can do is check if the array is empty and toast that no sketches were found.

client/images/unsaved-changes-dot.svg
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
epiqueras: test

server/views/404Page.js
catarak: or a maybe a 404 p5 sketch?

catarak:
Finally got a chance to review this, most of the stuff you changed looks great! I just have one request.

epiqueras:
I'm going to start working on the 404 sketch. Any ideas?
I was thinking maybe a mouse looking for the way out of a maze or a random retro game every time it's rendered?

lmccart:
@epiqueras great! I'd say the more pleasantly unexpected and surprising the better.

catarak:
@epiqueras i agree, have fun with it 😄 

For the overall design of the 404 page, can you add some copy so it looks something like this?


I think it would be good to have some text copy, just for clarity, and also for accessibility reasons. Could you also add a `<title>` to the HTML page, something like "404 Page Not Found - p5.js Web Editor"?

catarak:
my bad, it turns out I had a pending review that I wrote like two weeks ago and never submitted

catarak:
thanks for adding the header, it looks great! there's still one outstanding change in `SketchList.jsx` that i've asked for, let me know if you have any issues making that happen. 

one super super small thing is that I think the SVG dot is a little far from the file name and it looks weird:
screen shot 2017-01-05 at 5 19 16 pm

epiqueras:
Oops, I forgot to render the 404 page when the user is not found. It's fixed now.

PR #200, opened 2016-11-20
Move the interceptor code to a git module

MathuraMG:
This PR does the following - 
* The entire interceptor code is moved into a [submodule](https://github.com/MathuraMG/p5-interceptor). 
* added a post install script to ensure the submodule gets loaded during `npm install`
* Adjusted the logic in PreviewFrame for the accessible text output

@catarak - this is the current method for setting this up. What do you think about it?

Once you pull this, you would have to run 
'git submodule add git://github.com/MathuraMG/p5-interceptor.git static/p5-interceptor'
in the project root folder. This adds the submodule link to the project.

After that, if any changes are made to the submodule, it gets pulled when `npm install` is run and needs to be committed.

cc @cleezyITP


catarak:
I just did some reading about submodules, and I think since with this PR you're committing the file `.gitmodules`, the submodule is already added to the project so other users won't have to run 'git submodule add git://github.com/MathuraMG/p5-interceptor.git static/p5-interceptor'. 

I think users will have to run
```bash
$ git submodule init
```

then, according to [this stackoverflow answer](http://stackoverflow.com/a/33835815/4086967), I think the npm postinstall script should be
```bash
$ git submodule update --remote --recursive
```

we should both test this though!

MathuraMG:
Oh yes, we can use  `git submodule init`  to initialize it. 
The only issue with ` git submodule update ` is that it won't pull the latest commits in the submodule - which is why I thought we could do `git submodule foreach --recursive git pull origin master`
The other option is that we can update this manually as well as [shown in this.](http://stackoverflow.com/a/5828396)
Which do you think works better?

catarak:
I read that answer again, and I think the reason `git submodule update --remote --recursive` isn't working is because the branch to pull from isn't specified in the `.gitmodules` file. I think it needs to look like this:
```
[submodule "static/p5-interceptor"]
	path = static/p5-interceptor
	url = git://github.com/MathuraMG/p5-interceptor.git
	branch = master
```
I don't know if this is a _huge_ deal, it just seems like this is the recommended way to add submodules.

MathuraMG:
I just tested this, sorry I missed this option! I have made the necessary changes to the code.

catarak:
no worries 😄 I'm new to using submodules in a project and I just wanted to make sure we use them in the conventional/recommended way

PR #203, opened 2016-11-23
Fetch examples from p5 website repo

### LINKED ISSUE #149: "Open" to "Examples" when a user is not logged in

yining1023:
Fixes #149 

To run this, we need to 
- add github Oauth application Client ID and Client Secret in the `.env` file, like:
`CLIENT_ID=xxxxxxxx`
`CLIENT_SECRET=xxxxxxxxx`
- `$ npm install`
- `$ npm run fetch-examples`
Then open `Examples` in the editor, it should show the sketch list below:
screen shot 2016-11-23 at 3 16 29 am

To do list and questions:
- [x] Some examples require assets(images and audio files), save those assets on S3
- [ ] Set up a cron job that runs this task periodically

For now, each example's name is `Category : sketch name`, e.g: `Hello P5 : song`. What do you think of the naming method? @catarak. I'm open to other format.  I also think it would be good to have category folders in the sketch list eventually.


server/examples.js
```diff
@@ -0,0 +1,209 @@
...
+  padding: 0;
+}
+`;
+
+const a = objectID().toHexString();
+const b = objectID().toHexString();
+const c = objectID().toHexString();
+const r = objectID().toHexString();
+
+const client_id = process.env.CLIENT_ID;
```
catarak: could these environment variables be called `GITHUB_ID` and `GITHUB_SECRET`?

server/examples.js
```diff
@@ -0,0 +1,209 @@
...
+const client_secret = process.env.CLIENT_SECRET;
+
+const headers = {'User-Agent': 'p5js-web-editor/0.0.1'};
+
+mongoose.connect('mongodb://localhost:27017/p5js-web-editor');
```
catarak: There's an environment variable called process.env.MONGO_URL, I think it would make sense to use that here so that if we decide to move the mongodb somewhere else, we only have to change the url in one place 😄

catarak:
This is awesome! 🎉 🎉 🎉 

I think the naming convention is good. I think it's clear what it means. It would be nice to have folders, but we'll get there when we get there, considering how long it took me to add folders to files...

yining1023:
Yay! Thank you! I made some changes that you suggested. 

For the next step(saving assets on S3), my question is:
should I do something like [this](https://github.com/processing/p5.js-web-editor/blob/master/server/utils/db_reformat.js#L68)? It can find a file in one project and update it. Some projects need assets, some don't, so I need to go through each project's `sketch.js` and find out if it needs what particular assets and then add it?

catarak:
I think you would use the (s3 node package)[https://github.com/andrewrk/node-s3-client], and generate a unique id for each file in the way I do when uploading files in the editor (see (aws.controller.js)[https://github.com/processing/p5.js-web-editor/blob/master/server/controllers/aws.controller.js#L11], upload the files, and then you would do something similar to the code you linked to.

server/examples.js
```diff
@@ -153,57 +153,112 @@ function getSketchContent(projectsInAllCategories) {
...
-        });
 
-        newProject.save( (err, newProject) => {
-          if (err) throw err;
-          console.log('Created a new project in p5 user: ' + newProject.name);
-        });
+            const fileID = objectID().toHexString();
+            newProject.files.push({
+              name: assetName,
+              url: `https://raw.githubusercontent.com/processing/p5.js-website/master/dist/assets/examples/assets/${assetName}?client_id=${
```
catarak: i just tried it, and you don't need to include your github credentials in the url, e.g. https://raw.githubusercontent.com/processing/p5.js-website/master/dist/assets/examples/assets/mask.png works!

catarak: Also, apparently you're not supposed to use the raw.githubusercontent.com files directly, for a number of reasons discussed in this [Stack Overflow answer](http://stackoverflow.com/a/18049842/4086967). I would follow the directions for non-production environments. 

I'm glad you figured out a much better solution to this than uploading all of these assets to S3 😄

catarak:
I'm just going to merge this, and we can break out the known broken examples into separate issues:
- [ ] [Instance Container](http://p5js.org/examples/instance-mode-instance-container.html) needs to be handled as a special case: broken down into four sub examples and added as a HTML file rather than a JS file
- [ ] [Operator Precedence](http://p5js.org/examples/math-operator-precedence.html) Something involving the loop protect code is breaking this example
- [ ] Reverb, Convolution Reverb, and another sound example are broken as discussed in #212

catarak:
Also I noticed that the Convolution Reverb example is loading two assets with no name and a messed up url:
screen shot 2016-12-01 at 5 44 22 pm
screen shot 2016-12-01 at 5 45 06 pm

yining1023:
Oh, sorry I forgot to fix this.
In the Convolution Reverb example code, it has two lines like this:
`rawImpulse = loadSound('assets/' + cVerb.impulses[currentIR].name);`
And the `exmaple.js` is picking up empty assets names which is not necessary. I just fixed it.

catarak:
no worries, you are a boss for testing all of these examples

PR #221, opened 2016-12-07
Fixed example instance container

### LINKED ISSUE #217: Fix example: Instance Container
In the `p5` user sketch list,
`Instance Mode: Instantiating Container` needs to be handled as a special case: broken down into four sub examples and added as a HTML file rather than a JS file.

yining1023:
fixes #217 , split example `instance container` into four examples.
To update the sketch list, just need to do `$npm run fetch-examples` again.
screen shot 2016-12-07 at 4 11 30 pm


catarak:
This is super minor but could there be a space in between "Instance Container" and the number?

Issue #167, opened 2016-10-25
When deleting files in the editor hosted on S3, delete them from S3


catarak:
Right now, the reference to these files is deleted from the MongoDB. This also needs to happen when projects are deleted.

PR #222, opened 2016-12-09
#167 delete files in S3 when deleted in editor

PaliwalSparsh:
Thanks for the help. Please can you review this.

### DIFF
--- a/server/controllers/file.controller.js
+++ b/server/controllers/file.controller.js
@@ -1,5 +1,7 @@
 import Project from '../models/project';
+import each from 'async/each';
 import { resolvePathToFile } from '../utils/filePath';
+import { deleteObjectsFromS3 } from '../utils/s3Operations';
 
 // Bug -> timestamps don't get created, but it seems like this will
 // be fixed in mongoose soon
@@ -38,8 +40,15 @@ function getAllDescendantIds(files, nodeId) {
 }
 
 function deleteMany(files, ids) {
-  ids.forEach(id => {
+  let s3ObjectUrlList = [];
+  each(ids, (id, callback) => {
+    if (files.id(id).url !== undefined) {
+      s3ObjectUrlList.push(files.id(id).url);
+    }
     files.id(id).remove();
+    callback();
+  }, (err) => {
+    deleteObjectsFromS3(s3ObjectUrlList);
   });
 }
 
--- a/server/controllers/project.controller.js
+++ b/server/controllers/project.controller.js
@@ -2,6 +2,8 @@ import Project from '../models/project';
 import User from '../models/user';
 import archiver from 'archiver';
 import request from 'request';
+import each from 'async/each';
+import { deleteObjectsFromS3 } from '../utils/s3Operations';
 
 
 export function createProject(req, res) {
@@ -63,12 +65,29 @@ export function getProject(req, res) {
     });
 }
 
-export function deleteProject(req, res) {
-  Project.remove({ _id: req.params.project_id }, (err) => {
-    if (err) {
-      return res.status(404).send({ message: 'Project with that id does not exist' });
+function deleteFilesFromS3(files, fileCallback) {
+  let s3ObjectUrlList = [];
+  each(files, (file, callback) => {
+    if (file.url !== undefined) {
+      s3ObjectUrlList.push(file.url);
     }
-    return res.json({ success: true });
+    callback();
+  }, (err) => {
+    fileCallback();
+    deleteObjectsFromS3(s3ObjectUrlList);
+  });
+}
+
+export function deleteProject(req, res) {
+  Project.findById(req.params.project_id, (err, project) => {
+    deleteFilesFromS3(project.files, () => {
+      Project.remove({ _id: req.params.project_id }, (err) => {
+        if (err) {
+          return res.status(404).send({ message: 'Project with that id does not exist' });
+        }
+        return res.json({ success: true });
+      });
+    });
   });
 }
 
--- a/server/utils/s3Operations.js
+++ b/server/utils/s3Operations.js
@@ -1,0 +1,40 @@
+import s3 from 's3';
+import each from 'async/each';
+
+let client = s3.createClient({
+	maxAsyncS3: 20,
+	s3RetryCount: 3,
+	s3RetryDelay: 1000,
+	multipartUploadThreshold: 20971520, // this is the default (20 MB) 
+	multipartUploadSize: 15728640, // this is the default (15 MB) 
+	s3Options: {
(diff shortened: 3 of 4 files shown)


server/utils/s3Operations.js
```diff
@@ -0,0 +1,31 @@
...
+		Delete: {
+			Objects: [
+				{
+        			Key: objectKey,
+      			},
+			],
+		},
+	};
+	let del = client.deleteObjects(params);
+	del.on('end', function() {
```
PaliwalSparsh: Please can you tell what to do on successful deletion.

catarak: What you want to do here is add a second parameter to `deleteObjectFromS3`, that is a callback function, i.e. `export function deleteObjectFromS3(url, callback) { ...`. Then, here, you call that callback. 

For functions that use `deleteObjectFromS3`, you would use it like this:
```javascript
deleteObjectFromS3(url, function() {
  //code here that you want to get executed after file is deleted from S3
});
```

server/controllers/project.controller.js
```diff
@@ -63,12 +64,24 @@ export function getProject(req, res) {
...
+    if (!(file.url === undefined)) {
+      deleteObjectFromS3(file.url);
     }
-    return res.json({ success: true });
+  });
+}
+
+export function deleteProject(req, res) {
+  // Deleting project files form S3
+  Project.findById(req.params.project_id, (err, project) => {
```
catarak: I left a thorough answer for this in another comment, but the short answer is callbacks 😄

PaliwalSparsh:
@catarak Made all the required changes. I never had used async it was really awesome. Thanks :smile:, made my day.

server/controllers/project.controller.js
```diff
@@ -63,12 +65,29 @@ export function getProject(req, res) {
     });
 }
 
-export function deleteProject(req, res) {
-  Project.remove({ _id: req.params.project_id }, (err) => {
-    if (err) {
-      return res.status(404).send({ message: 'Project with that id does not exist' });
+function deleteFilesFromS3(files, callback) {
+  each(files, (file, callback) => {
```
catarak: I would rename `callback` here to `fileCallback` for the sake of clarity.

server/controllers/project.controller.js
```diff
@@ -63,12 +65,29 @@ export function getProject(req, res) {
...
+    if (!(file.url === undefined)) {
+      deleteObjectFromS3(file.url, () => {});
```
catarak: The callback setup here is not quite right! This should be
```javascript
deleteObjectFromS3(file.url, fileCallback);
```
And then line 73 is extraneous and should be removed.

server/controllers/file.controller.js
```diff
@@ -39,6 +40,10 @@ function getAllDescendantIds(files, nodeId) {
 
 function deleteMany(files, ids) {
   ids.forEach(id => {
+    // Delete file from S3 bucket
+    if (!(files.id(id).url === undefined)) {
```
catarak: I _think_ this is slightly more clear as 
```javascript
if (files.id(id).url !== undefined) {
```

server/controllers/file.controller.js
```diff
@@ -39,6 +40,10 @@ function getAllDescendantIds(files, nodeId) {
...
+      deleteObjectFromS3(files.id(id).url);
+    }
     files.id(id).remove();
```
catarak: should this be within the `deleteObjectFromS3` callback? will this cause an issue where the API returns but the files are still be deleted from S3 and then they don't get deleted properly?

PaliwalSparsh:
@catarak I am really really sorry I pushed some more changes while you were reviewing.
Actually before it was deleting all files using different requests, now i made it to delete all files in one request.

server/controllers/file.controller.js
```diff
@@ -38,8 +40,15 @@ function getAllDescendantIds(files, nodeId) {
...
 function deleteMany(files, ids) {
-  ids.forEach(id => {
+  let s3ObjectUrlList = [];
+  each(ids, (id, callback) => {
+    if (files.id(id).url !== undefined) {
+      s3ObjectUrlList.push(files.id(id).url);
+    }
     files.id(id).remove();
+    callback();    
+  }, (err) => {
```
PaliwalSparsh: 3rd parameter (err) => {} is called when [async](https://caolan.github.io/async/docs.html#each) for all the element gets completed or hits an error. Thus, upon completion ie getting urlList I had given the list to deleteObjectsFromS3.

server/controllers/project.controller.js
```diff
@@ -63,12 +65,29 @@ export function getProject(req, res) {
...
+function deleteFilesFromS3(files, fileCallback) {
+  let s3ObjectUrlList = [];
+  each(files, (file, callback) => {
+    if (file.url !== undefined) {
+      s3ObjectUrlList.push(file.url);
     }
-    return res.json({ success: true });
+    callback();
+  }, (err) => {
+    fileCallback();
```
PaliwalSparsh: Here this
`deleteObjectsFromS3(s3ObjectUrlList, fileCallback);`
was not used because if the project had many files on S3, only when all these files got deleted from S3 we get the `fileCallback` running. Thus, when we click on the delete button in front of a project in the "My sketches" dialogue box, the project remains there for a while and after few seconds it would disappear, when `fileCallback` gets executed upon getting response from S3 that files have been deleted.

Also once we get list of URL we need to delete from S3 we don't need anything else. We can safely delete files locally and S3 delete operation can be done after that.

PaliwalSparsh:
I have made the changes you suggested. Please can you review the PR again. Thankyou.

catarak:
A couple things I noticed:
1. If a project has never been saved, or a file has been uploaded and the project has not been saved since the file has been uploaded, the file will not get deleted from S3. I guess this will either have to happen on the front-end, or an API request to the web editor API will have to be made to delete the file.
2. I'm actually not able to get this working, but I think it's a configuration option on my end. I think it has to do with the region in the URL? I'll look into it, but it may mean we have to include the region in the `.env` file.
screen shot 2016-12-12 at 12 10 50 pm

catarak:
I also thought of another bug that this causes:
If a user creates a project, uploads a file to S3 for that project, then duplicates that project, then two projects reference the same S3 file via its URL. As of right now, the file is not duplicated on S3. Then, if one of those projects is deleted, then the other project will be broken because the file it references will be deleted. 

So I think the solution to this is to copy the files on S3 when a user duplicates a project. Definitely not the most $$$ efficient, but we'll put a limit to how much a user can upload which will help.

Issue #220, opened 2016-12-07
last line of text in editor is obscured by console toggle


kjhollen:
This just came up in the last week or so of class. The last line of text in the console isn't visible. Usually what this means is that the last closing brace } in the code is invisible, if it's the only thing on the last line and user has scrolled to the bottom.

PR #223, opened 2016-12-11
1 line margin at bottom

Akarshit:
Fix for #220


Akarshit:
I will just test it again with what happens as the font changes when I use `rem`. If it works I will modify the PR.

catarak:
Ohhhh I didn't think about that, you're totally right, in this case it makes sense to use `em`'s 😄

PR #224, opened 2016-12-12
Added google signup

Akarshit:
Added the backend support for Google signup/signin. There is no UI added yet as I thought the UI might need some attention to handle Github and Google signin(though adding just the logo looks cool enough).

The following has to be added to .env
```
GOOGLE_CLIENT_ID=appId
GOOGLE_CLIENT_SECRET=appSecret
```
Also a redirect has to be set up in the google project for `DOMAIN/auth/google/callback.`

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "build": "NODE_ENV=production webpack --config webpack.config.prod.js --progress",
     "test": "echo \"Error: no test specified\" && exit 1",
     "fetch-examples": "node fetch-examples.js",
-    "postinstall" : "git submodule update --remote --recursive"
+    "postinstall": "git submodule update --remote --recursive"
   },
   "main": "index.js",
   "author": "Cassie Tarakajian",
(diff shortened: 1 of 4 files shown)


catarak:
This is great but I'm not sure why you did this as there aren't any open issues about it! I think it's a higher priority issue to fix the Github login, and we hadn't discussed any social login aside from that.

Akarshit:
@catarak I did is because I felt that google login is something that is more common and generally the no. of people who have account on google would be more than those from github.
I could work on the github issue, if you could tell me what the issue is.

catarak:
@lmccart @shiffman do you have any opinion on other social login besides Github? Is it convenient or annoying? i could be convinced either way.

lmccart:
I think at this point, I'd rather limit the options just to leave less room for bugs and other errors and let us focus on the essential features and fixes before we expand the scope.

shiffman:
agree with @lmccart  But if we can really having a "singing" way for people to login I am all for that! (see excellent github issue subject typo).

Akarshit:
@catarak So what's the verdict on this one? 😛

catarak:
I think unfortunately at the moment I'm not going to be able to merge this in... I realized that before we can add Github or Google login to the app, we need a user settings page, like I've mentioned in #230, that allows users to link social accounts to an existing account. Otherwise if a user registers with a username, and then tries to login with Github, it's ambiguous what should happen.

Akarshit:
OK cool!
Regarding linking of account. It think the least painful experience will be to just link the accounts automatically.
For example if a user registers with a username and then at some point of time tries to login with Github, just link both the accounts automatically and sign her in.

PR #227, opened 2016-12-14
implement new design for sketchbook

lmccart:


lmccart:
I'm not sure how you feel about the removal of seconds from the sketchlist view. It seemed like this might be more information than needed, and it might be clearer to read without. I'm happy to revert this if you feel otherwise though!

PR #228, opened 2016-12-14
Taking login and sign pages out + new design for login and signup page

yining1023:
1. made `Loginview` and `SignupView` two separated pages
2. set `previousPath` in the `App` component, so the `previousPath` can be updated not only when users are navigating inside `IDEview`, but also between each separated pages.
3. implemented the new design for login and signup page
screen shot 2016-12-14 at 11 03 44 am
screen shot 2016-12-14 at 11 03 25 am


client/modules/User/components/LoginForm.jsx
```diff
@@ -5,27 +5,27 @@ function LoginForm(props) {
   const { fields: { email, password }, handleSubmit, submitting, pristine } = props;
   return (
     <form className="login-form" onSubmit={handleSubmit(props.validateAndLoginUser.bind(this, props.previousPath))}>
+      <p className="login-form__subtitle">Email</p>
```
catarak: The HTML5 standard way to add a label for an input is to use a label tag, i.e.
```html
<label for="email-input">Email</label>
<input type="text" id="email-input>
```

client/styles/abstracts/_variables.scss
```diff
@@ -110,3 +110,13 @@ $console-error-color: #ff5f52;
 
 $toast-background-color: #979797;
 $toast-text-color: $white;
+
+$form-title-color: rgba(51, 51, 51, 0.87);
```
catarak: I think it would make sense to add all of these colors to the big color map. Or maybe it doesn't since maybe it doesn't make sense to have these forms have different themes. We'll have to discuss this one.

client/styles/components/_login.scss
```diff
@@ -1,20 +1,31 @@
...
 .login-form__email-input,
 .login-form__password-input {
-	width: #{300 / $base-font-size}rem;
+	width: #{360 / $base-font-size}rem;
+	height: #{40 / $base-font-size}rem;
 }
 
 .login-form__field {
-	margin: #{20 / $base-font-size}rem 0;
+	margin: 0 0 #{24 / $base-font-size}rem 0;
```
catarak:
```sass
margin-bottom: #{24 / $base-font-size}rem;
```
is easier to read 😄

catarak:
I'm not sure about the cancel button at the bottom of the page. I think the expected UX is for the button to be at the top or top-left to go back, e.g.
screen shot 2016-12-14 at 11 39 29 am
screen shot 2016-12-14 at 11 38 47 am

client/styles/main.scss
```diff
@@ -34,6 +34,7 @@
 @import 'components/toast';
 @import 'components/timer';
 @import 'components/force-authentication';
+@import 'components/form-container';
```
catarak: The only error I found was that you need to remove the following lines from this file:
```scss
@import 'components/signup';
@import 'components/login';
```
since you deleted those files! 😄

yining1023: You are very right! 😄 I just deleted them.

PR #230, opened 2016-12-14
Email verification

Akarshit:
Added basic email-verification.

Now let's discuss where to user this and how.


catarak:
thank you for finally making a nice email template 🙏 🙏 🙏

server/controllers/user.controller.js
```diff
@@ -1,8 +1,8 @@
 import User from '../models/user';
 import crypto from 'crypto';
 import async from 'async';
-import nodemailer from 'nodemailer';
-import mg from 'nodemailer-mailgun-transport';
+import mail from '../utils/mail';
+import auth from '../util/auth';
```
catarak: this line for me is throwing an error in node, I think it should be
```javascript
import auth from '../utils/auth';
```

server/utils/mail.js
```diff
@@ -0,0 +1,81 @@
...
+import mg from 'nodemailer-mailgun-transport';
+
+const auth = {
+  api_key: process.env.MAILGUN_KEY,
+  domain: process.env.MAILGUN_DOMAIN,
+};
+
+class Mail {
+  constructor() {
+    console.log(auth);
```
catarak: also can you remove this `console.log`?

catarak:
As for how to use the email verification... we'll need to have a warning on the bottom half of the file uploader when a user is not logged in, something like:
verify
I think it would also be nice to have a place where a user could check if their account is verified, and resend the email if they never received it. I think it would be on a page that needs to exist, a settings page, which would be similar to the CodePen settings page:
screen shot 2016-12-14 at 4 27 38 pm
screen shot 2016-12-14 at 4 28 23 pm
screen shot 2016-12-14 at 4 29 48 pm
We don't need to add all of the different settings immediately, but at least just the email verification stuff and being able to link social accounts.

Akarshit:
@catarak I should have mentioned that this PR is still WIP. I opened it to discuss which places to user the `verified` field. 
> I think it would also be nice to have a place where a user could check if their account is verified, and resend the email if they never received it.

This I would do i a separate PR. Though I will modify this one so that it could be compatible after wards. Basically `verified` would become a integer instead of boolean with the following states -
-1. verifiacation email sent but not verified.
0. verified 
1. verification mail sent again. Still not verified.

This was we can know from the DB which people are having problems receiving mail from us and then we can try sending them mail through a different service(maybe).

catarak:
separate PR for improved email verification is fine!

Akarshit:
@catarak I am done with this on the back-end side. Can you please review the PR and let me know if I have to make any changes.

server/controllers/user.controller.js
```diff
@@ -24,11 +24,19 @@ export function createUser(req, res, next) {
...
             return next(loginErr);
           }
-          res.json({
-            email: req.user.email,
-            username: req.user.username,
-            preferences: req.user.preferences,
-            id: req.user._id
+          mail.send('email-verification', {
+            body: {
+              link: `http://${req.headers.host}/api/verify?t=${auth.createVerificationToken(req.body.email)}`
```
catarak: Since all routes that contain `/api/` only serve JSON, this link should be
```javascript
`http://${req.headers.host}/verify?t=${auth.createVerificationToken(req.body.email)}`
```
and then that page will make an AJAX request to
```
http://${req.headers.host}/api/verify?t=${auth.createVerificationToken(req.body.email)}
```

Akarshit: @catarak Do you want me to redirect `/verify` to `/api/verify' in routes/server.routes.js?

Akarshit:
@catarak Hey could take a look at this when your are free. I need some of the files used in this PR to make the next PR. 
Let me know what needs to be changed.

server/controllers/user.controller.js
catarak: No. Remember that this app is all front-end communicating with an API, therefore the redirection should be on the front end. Once the status of the verification is received, the user will be redirected to a verification status/success/whatever makes sense page.

catarak:
My bad, for some reason I had it in my head that this was dependent on #248 but this is just the backend so it's fine. Still one small thing to fix but otherwise looks great.

Akarshit:
@catarak Made the changes.

catarak:
i ended up having to revert this because password reset emails weren't sending for me, can you look into this?

PR #232, opened 2016-12-17
New design for ResetPassword view and NewPassword view

yining1023:

1. made `ResetPasswordview` and `NewPasswordView` two separated pages 
2. implemented the new design for resetPassword and newPassword page

I made the exit icon on both pages going to the home page. People can only go to `resetPassword` page through `login` page, and if the exit icon leads people back to `login` page, the exit icon on `login` page will lead people back to `resetPassword` page, which would be an endless loop.

I couldn't test the whole functionality about resetting password on my local computer, could you test it again? I just want to make sure I didn't break anything. It should be fine though, I only updated `token` to `this.props.params.reset_password_token` in `NewPasswordView`(instead of getting it from `IDEView`).
screen shot 2016-12-16 at 6 03 57 pm
screen shot 2016-12-16 at 6 03 38 pm


client/styles/components/_reset-password.scss
```diff
@@ -1,29 +1,13 @@
...
-  margin: #{20 / $base-font-size}rem 0;
-}
-
 .reset-password__submitted {
-  width: #{300 / $base-font-size}rem;
+  width: #{360 / $base-font-size}rem;
   display: none;
   text-align: left;
-  font-size: #{12 / $base-font-size}rem;
+  font-size: #{11.65 / $base-font-size}rem;
```
catarak: why 11.65 pixels? I think most browsers will round to the nearest pixel anyway...

catarak:
I also noticed the view for `/reset-password/<token>` doesn't look right:
screen shot 2016-12-19 at 12 53 10 pm

yining1023:
I found that when people go to `/reset-password/<token>` directly(without going to ideView first), any style with the minxin`themify(){}` cannot be injected properly. I think an easy way to fix this it to create styles for these`input` and `icon` that not rely on `themify()`, since we don't need different colors on those pages. What do you think?

catarak:
yes, I think that's the right thing to do 😄

yining1023:
Cool! Just made some changes, now the styles should be fine.😄

Issue #247, opened 2017-01-04
Web login saves username/password, but email required to login


thepackrat:
The default Chrome or Safari password autofill workflow from the p5 web editor sign up saves the username and the password used to sign up, but when you arrive back at the login screen, it requires the email to log in.

Could the login flow be modified to use either the email address or username?

PR #250, opened 2017-01-05
Use username also to log in

Akarshit:
Fix for #247

### DIFF
--- a/client/modules/User/components/LoginForm.jsx
+++ b/client/modules/User/components/LoginForm.jsx
@@ -6,7 +6,7 @@ function LoginForm(props) {
   return (
     <form className="form" onSubmit={handleSubmit(props.validateAndLoginUser.bind(this, props.previousPath))}>
       <p className="form__field">
-        <label htmlFor="email" className="form__label">Email</label>
+        <label htmlFor="email" className="form__label">Email/Username</label>
         <input
           className="form__input"
           aria-label="email"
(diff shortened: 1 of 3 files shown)


client/modules/User/components/LoginForm.jsx
```diff
@@ -6,7 +6,7 @@ function LoginForm(props) {
   return (
     <form className="form" onSubmit={handleSubmit(props.validateAndLoginUser.bind(this, props.previousPath))}>
       <p className="form__field">
-        <label htmlFor="email" className="form__label">Email</label>
+        <label htmlFor="email" className="form__label">Email/Username</label>
```
catarak: To me this is more clear as "Email or Username", to me "Email/Username" implies that they are one in the same

client/modules/User/components/LoginForm.jsx
```diff
@@ -6,7 +6,7 @@ function LoginForm(props) {
...
         <input
           className="form__input"
           aria-label="email"
```
catarak: could you update this to
```
aria-label="email or username"
```

catarak:
ty this is great 👼 just have a couple tiny changes!

client/modules/User/components/LoginForm.jsx
Akarshit: Do we need this even after changing `Email/Username` to `Email or Username`?

Issue #299, opened 2017-02-06
UI Detail: color of "by" in author/account name


machinic:
We may make design updates that change the color of the author and/or the title text but `by` should be the same gray as everything else is currently (connecting words like `by` likely will not change from that gray).

![editor-author-type-detail-color](https://cloud.githubusercontent.com/assets/3985997/22649597/a2c27caa-ec49-11e6-89b6-bb6d5bbe216a.png)

Katy310:
May I work on this issue?

PR #303, opened 2017-02-14
Fix UI Detail: color of 'by' in author/account name #299

Katy310:


client/styles/components/_toolbar.scss
```diff
@@ -103,6 +103,7 @@
 
 .toolbar__project-owner {
 	margin-left: #{5 / $base-font-size}rem;
+	color: #bfbfbf;
```
catarak: Because the editor has different themes, you need to use a SCSS mixin/map to set colors so that they are correct in all of the themes. This style should be
```scss
.toolbar__project-owner {
  @include themify() {
    margin-left: #{5 / $base-font-size}rem;
    color: getThemifyVariable('inactive-text-color');
  }
}
```
if you look at `client/styles/base/_base.scss`, you'll see that all `<a>` tags have this color.

Katy310:
I've made the required change.

catarak:
in order for the SCSS function `getThemifyVariable` to work, it needs to be wrapped with the mixin `@include themify() { ... }` or else it doesn't work. I know the syntax is a little annoying but I swear it's less verbose than if the themes were all done manually, i.e.
```scss
.light  .toolbar__project-owner {
    color: $light-inactive-text-color;
}
.dark  .toolbar__project-owner {
    color: $dark-inactive-text-color;
}
.high-contrast  .toolbar__project-owner {
    color: $highcontrast-inactive-text-color;
}
```

Katy310:
Thanks for the help! I am new to this project and js and sass in general and am trying to explore the code base through these issues in an attempt to learn.

catarak:
No worries, it's awesome that you want to learn \ :smile: /

This is _almost_ there. I'm seeing the following error: 
```
Invalid CSS after "}": expected "}", was ""
      in /Users/ctarakajian/Development/projects/p5.js-web-editor/client/styles/components/_toolbar.scss (line 132, column 2)
```

This is because you are missing a closing curly brace on line 109 for the selector `.toolbar__project-owner`.

Katy310:
Hopefully, it's fixed now.

Issue #298, opened 2017-02-06
UI Detail: vertical alignment of Title text


machinic:
The baselines of all text in this row should align. Title text is `2px` too high off of it. 

![text-alignment-title-detail](https://cloud.githubusercontent.com/assets/3985997/22649564/7bb320e2-ec49-11e6-9f95-d01251317c1c.png)

saumya1906:
I wish to work on this issue. Any pointers / guidelines?

PR #305, opened 2017-02-14
Issue#298 fix vertical alignment

Katy310:


catarak:
just tested this, and to me it seems like the alignment is still off.
screen shot 2017-02-14 at 3 33 12 pm
I noticed that you added 10px padding to the top, but in #298 there is only an alignment issue of 2px. Maybe only 2px of padding should be added, or maybe the pencil icon is pushing down the title and causing the misalignment?

Katy310:
2px Does not solve the problem. 
selection_045

catarak:
okay, then I'm sure there's another way to fix it! if you need help figuring this out let me know 😄

Katy310:
I will try and investigate whether the pencil icon is creating a problem.

PR #316, opened 2017-03-04
Style change to toasts

### LINKED ISSUE #294: UI detail: Notification (toast) bar styling
Though there's a save example in the last design drop, the design guidelines could use some shoring up with respect to behavior and placement of toasts and other bar notifications — those are forthcoming. 

In the meantime, if we can adjust the styling to better match the guide that would be Awesome. 

Compare current implementation:

bar-notif-detail

Design guide for bar notif: 

bar-notif-design-guide

Background-color is `#4A4A4A`. Close affordance same color as text. Size and positioning of message within should be adjusted as well. Some box shadow applied as well `box-shadow: 0 6px 6px 0 rgba(0,0,0,0.10);`

lonesword:
fixes #294

### DIFF
--- a/client/styles/abstracts/_placeholders.scss
+++ b/client/styles/abstracts/_placeholders.scss
@@ -51,6 +51,25 @@
 	padding: 0;
 }
 
+%icon-toast{
+	@include themify() {
+		color: $toast-text-color
+		& g {
+			fill: $toast-text-color
+		}
+		&:hover {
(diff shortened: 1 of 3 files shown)


catarak:
this is beautiful. it seems like you figured out the `themify` mixin alright.

lonesword:
Thanks :)

Issue #170, opened 2016-10-25
Switch between http and https


catarak:
Sometimes the sketch needs to be rendered using http, and sometimes it needs to be rendered using https. A UI element needs to be added to toggle between the two, probably with an explanation of why you would need to (i.e. there's a question mark next to the toggle which opens a modal).

Issue #139, opened 2016-10-11
Force https connection on certain pages


catarak:
For Login/Signup, the user should be redirection to the https version of the site, and redirected back to the http version of the site once they are logged in, unless their initial connection is https.

shiffman:
I'm not sure if this should be a separate issue, but some sketches require http and some require https.  Some examples:

* accessing webcam or microphone require https
* APIs served over https require https
* APIs served over http require http

I don't have any great ideas as to how we should make this evident / selectable by users beyond just changing the URL in the address bar.

catarak:
The nginx configuration has been updated so that if login, signup, or password reset are loaded sever side, they will be redirected to https. I still need to
1. Force the front end to render these pages https (i.e. force them to render server side)
2. Redirect to http again post login
3. #170: have some interface for users to switch between http and https

andrewn:
I've [prototyped a quick approach to redirecting login](https://github.com/andrewn/p5.js-web-editor/commit/e32f5c0a1032e12831b2cebea74d82b268e289c4) to HTTPS on the client. The higher-order component detects if the protocol is HTTPS when mounting the view component and redirecting if not.

Getting the view to redirect back to the original protocol afterwards is a bit trickier since the app is loading from scratch on `/login`.

1. We could pass a query string such as `/login?redirect=http` to tell the app whether to reload on HTTP afterwards;
2. If we have a global flag in local storage of the user's preference for HTTP/S (perhaps set by #170) then we could use that to determine where to redirect.

Does this approach make sense? Should I explore some alternatives?

catarak:
This seems like the right direction to me! 

I think that the preference for http vs. https actually needs to be set on a project basis, since depending on what the user wants to do for that specific sketch, they may need it to be http or https. But I think it this setting should be http by default.

PR #319, opened 2017-03-09
Force HTTPS redirection for log in and sign up

andrewn:
~**:warning: #324 and #323 might make this easier to test**~

**:warning: Merge #327 first I've merged it into this branch**

This implements client-side redirection to HTTPS of the `/signup`, `/login` and `/reset-password` views described in #139. If that view is mounted and is not on HTTPS then it's redirected, causing the app to reload.

New projects are created with a `serveSecure: false` flag. After sign-up/log-in/reset, this flag is used to decide which protocol to redirect back to. 

`serveSecure` defaults to `false`. Until some user-facing UI switch is built (see #170) then it will always redirect back to HTTP.


catarak:
🎉 🎉 🎉

andrewn:
I think this is finally ready for review!

- Account page will also redirect to HTTPS (see #327)
- Disables redirection in development mode, put `FORCE_TO_HTTPS=true` in your `.env` to test

catarak:
woohoo this looks great AND works great! 

this is definitely territory for a new issue/PR, but I just realized that we'll need to do some pushing/popping of the redux state into localStorage, i.e. if a user is not logged in and starts working on a sketch, they have to log in to save it.

PR #320, opened 2017-03-09
alignment issue, #298

kshivanku:


client/styles/components/_toolbar.scss
```diff
@@ -88,6 +88,7 @@
 		}
 	}
 	cursor: pointer;
+	line-height: 18px;
```
catarak: for reasons, I'm not using pixel measurements in this project, but instead `rem`'s and `em`'s. Could you make this
```
line-height: #{18 / $base-font-size}rem;
```

kshivanku: done, sent another PR

catarak:
you can amend commits to the same PR 😄 but i will close this one and merge the other one

PR #324, opened 2017-03-12
Enable CORS for all origins and requests on API

andrewn:
Enables the web client being served at `http://example.com` to make requests to the server API at `https://example.com/api` using [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS).

This means that when #323 is merged in, the client can be told to always access the API over HTTPS. It also makes #319 easier to test since [`local-ssl-proxy`](https://github.com/cameronhunter/local-ssl-proxy) can be used to proxy the app over SSL on a different port and the `API_URL` environment variable can point to the SSL API.

With the default current configuration, any webpage will be able to access the API. However, we could [configure the middleware](https://www.npmjs.com/package/cors#configuration-options) to only allow whitelisted origins to access it. We'd need to add another env variable to whitelist them.


catarak:
Yeah, we should probably whitelist certain origins, just to be safe. `localhost`, `alpha.editor.p5js.org`, `beta.editor.p5js.org` (for when we have a beta!), and `editor.p5js.org` should be on that.

Also, a dumb question, how does a AJAX request from an HTTP origin to an HTTPS API work? Is the request HTTPS or HTTP?

catarak:
I asked about the ajax request from http to https, and apparently the request is https, which is pretty rad. There's one gotcha though, that it's vulnerable to man-in-the-middle attacks: http://stackoverflow.com/a/7594254/4086967

andrewn:
I've whitelisted `*.p5js.org` origins for CORS. In development mode, `localhost` is also added.

I can confirm that setting the `API_URL` to the API served over HTTP makes login requests over HTTPS even when the page is served on HTTP. 🙌 

For testing, I've used the setup [described in this gist](https://gist.github.com/andrewn/953ffd5cb17ac2634dc969fc7bdaff3f).

catarak:
When running this locally, I'm not able to make this work. I have the local ssl proxy running, and after running `sudo npm start` and navigating to `http://localhost`, I see the following error when trying to make an ajax request to the API (in Chrome):
```
Failed to load resource: net::ERR_INSECURE_RESPONSE
```
I also tried in Firefox and it also doesn't work. Maybe there's something I'm not doing?

andrewn:
Have you accepted the self-signed certificate at `https://localhost`?

On Chrome this is done by visiting `https://localhost` and clicking "ADVANCED" and then "Proceed to localhost (unsafe)" on the warning page.

Once this is accepted then Ajax requests through the proxy from HTTP should work.

catarak:
cool, yes, that worked. could you make a note of having to accept the certificate before it will work in http in the gist? it seems obvious in hindsight but just so it's totally clear...

andrewn:
I mentioned it in step 5 but I've now changed it to be clearer with more **bold**.

PR #325, opened 2017-03-16
Adding User Settings View

### LINKED ISSUE #248: Add User Settings View
Permit logged-in user to see/change associated email address

There's currently no obvious way for a logged-in user to see or change the email address associated with their account. It should probably live in the right-hand user menu.

yining1023:
Fixes #248

This pull request - 
- adds `My account` to the upper-right user menu
- creates a `My account` page with `username`, `email`, `currentPassword`, and `newPassword`.
- checks if the `username` and `email` are in use, if the `currentPassword` is valid, and if `new password` is not empty when `currentPassword` is not empty. If yes, update all user settings.
screen shot 2017-03-15 at 10 51 24 pmscreen shot 2017-03-15 at 10 49 03 pmscreen shot 2017-03-15 at 11 13 44 pmscreen shot 2017-03-15 at 11 09 48 pm


client/modules/User/pages/AccountView.jsx
```diff
@@ -0,0 +1,115 @@
...
+          const error = {};
+          error[fieldToValidate] = response.data.message;
+          throw error;
+        }
+      });
+  }
+  return Promise.resolve(true).then(() => {});
+}
+
+function validate(formProps) {
```
catarak: since login/signup use the same regexes to do validation, it probably makes sense to set them to some other variable in another file and import them. I've played around with the idea of converting the `constants.js` file to a folder called `constants`, then changing `constants.js` to `actiontypes.js`, and then, more constant files can be added as necessary. Or maybe there's a better place to put this file.

catarak:
this looks great, you killed it 💃 

I noticed that the dropdown styling is a little funky... `Log Out` has some extra padding to the left and is a different color/has a different hover state. You mind matching it to the other two dropdown items?
screen shot 2017-03-16 at 1 26 16 am

yining1023:
Thank you!☺️

I just fixed the `Log out` button style.

client/modules/User/pages/AccountView.jsx
yining1023: I agree. I put the validate function into `utils/reduxFormUtils.js`

PR #332, opened 2017-03-31
Fix broken Markdown formatting

JunShern:


catarak:
lol i should definitely render Markdown before i push it

Issue #329, opened 2017-03-30
store redux state in localStorage for http -> https redirection


catarak:
When a user is redirected from http to https and vice versa, their redux state will be lost. Therefore, this will need to be stored for certain use cases.

PR #334, opened 2017-04-05
Persists Redux store to/from sessionStorage

andrewn:
As required by #329 we need this for login so the user's work isn't lost. 

- saves the Redux store to sessionStorage when the route changes to `/login` or `/signup`
- disables the "Are you sure" alert
- reinstates the store and clears from sessionStorage when the App loads

local/sessionStorage is only available for the current origin, so data saved under `http` is not available under `https`. This should be ok for the login use case as we'll redirect back to same protocol.

I initially tried to keep the whole store synced to `sessionStorage`. This gives you some nice features like being able to refresh the page and not losing any data, but I think this would be a lot more work to make work properly.


andrewn:
@catarak I think this is ready for review.

I've manually tested this by logging in/out a few times. Saving projects and loading them again. It seems ok but I think you might be able to find some edge cases.

catarak:
This is working for me if you start on http, but if you open the page on https, and then login or sign up, then it is redirected back to http. Maybe there's something up with the https redirection that we didn't catch? Let me know if you can reproduce this.

andrewn:
This is because we use the project's `serveSecure` flag to indicate whether to redirect back to HTTP or not. All new projects are created with `serveSecure: false` so it will always redirect back to HTTP.

Maybe new projects started on HTTPS should be created with the `serveSecure: true`?

catarak:
Got it. My instinct is that for now, new projects started on HTTPS should be created with `serveSecure: true`. Once there is a UI element indicating this to the user, it will be more clear whether the sketch should be served with HTTP or HTTPS.

andrewn:
Projects created on HTTPS now have `serveSecure: true`.

But... because of this, using this flag to figure out if we should redirect or not won't work as the blank project created on HTTPS will have `serveSecure: true` set even though it's not the real project we want to go back to. 😭 

So, I'm appending the querystring parameter `?source=<protocol>` when we changing protocol. This is used to figure out which protocol to redirect back to. Not as elegant as I'd like :-(

catarak:
Just tested this and it works great 🎉  I think the query string parameter is okay, as login/signup aren't urls that users will copy and paste.

Issue #87, opened 2016-09-06
Logging in with github account logs you in as aparrish


mimiyin:

catarak:
Wow, that's really strange. I'm not able to reproduce this, can you tell me

- did you have an account previously, before you tried to log in with github?
- the email/username of the github account you tried to login with
- were you using your own computer or a shared computer?

mimiyin:
I was able to repro with my github account.
I had an account already under mimi.yin@nyu.edu
But several students in my class reported it and they didn't already have accounts.

shiffman:
I can confirm that this just happened to me, tried with two github accounts.

catarak:
Okay, it seems like it's an issue when a user doesn't have an account yet and tries to login with Github. For now, I'll just remove logging in with Github, figure out why this is happening, and readd it.

aparrish:
hahahahaha

Akarshit:
I tried this in production, but am unable to reproduce this. I confirmed that the user doesn't have a account while doing sign-in.

catarak:
I think the issue is if a user doesn't have a public email--the github oauth needs to be configured to get the user's email even if they have not made it public.

shiffman:
end of an era

PR #337, opened 2017-04-09
Search for existing user account using Github emails

andrewn:
Although I can't reproduce #87 locally,  for me Github login didn't work because my Github email address isn't public and so wasn't returned in the profile data.

The `user:email` scope from Github returns the private email addresses associated with the user's account. Each address has a `verified` and `primary` status. We try and find an existing web editor account matching any of the verified email address and if not, use the `primaryEmail` to create a new account.

### DIFF
--- a/client/modules/User/pages/AccountView.jsx
+++ b/client/modules/User/pages/AccountView.jsx
@@ -7,6 +7,7 @@ import axios from 'axios';
 import { updateSettings } from '../actions';
 import AccountForm from '../components/AccountForm';
 import { validateSettings } from '../../../utils/reduxFormUtils';
+import GithubButton from '../components/GithubButton';
 
 const exitUrl = require('../../../images/exit.svg');
 const logoUrl = require('../../../images/p5js-logo.svg');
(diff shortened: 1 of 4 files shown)


catarak:
this seems to be working for me--the only thing that needs to be fixed is the button position, it should be centered:
screen shot 2017-04-12 at 1 11 08 pm

andrewn:
Weirdly, this appears as an (off-centered) link for me:

screenshot 2017-04-12 19 15 15

I'll center it either way.

catarak:
maybe I have a file that's somehow not in version control? v weird

andrewn:
I can see the `@extend %button;` SASS rule in the `.github-button` but can't see the button styles in the browser.

Maybe I have a different version of `node-sass` installed?

```bash
$ npm ls node-sass
p5.js-web-editor@0.0.1 processing/p5.js-web-editor
└── node-sass@3.13.1
```

andrewn:
@catarak Centered:

screenshot 2017-04-12 22 32 50

catarak:
That is definitely it--I have `node-sass@3.7.0`, which I probably should upgrade.

Issue #205, opened 2016-11-23
Find


catarak:
Actually research and implement find.

catarak:
The design for this exists and is on zeplin.

andrewn:
CodeMirror has a `search` addon that manages it's own dialog box and UI. We could use this as-is and use CSS to style the dialog. This is the approach Glitch seem to have taken:

<img width="458" alt="screenshot 2017-04-28 10 14 24" src="https://cloud.githubusercontent.com/assets/1762/25520346/e3a4b642-2bfb-11e7-8858-038a7932858a.png">

It doesn't look like this supports adding to the search dialog such as the Regex and  Case-sensitive search options in the design:

<img width="382" alt="screenshot 2017-04-28 10 17 51" src="https://cloud.githubusercontent.com/assets/1762/25520462/465c5952-2bfc-11e7-8df2-4ccdcfa43594.png">


Alternatively, we could fork the addon and convert it to use React which is more work but would give us more control.

catarak:
In my conversation with the developer at Glitch he advised against making changes to CodeMirror as it can be a big time suck. Though looking at the CodeMirror [documentation](https://codemirror.net/doc/manual.html#addon_dialog), it seems like regex is supported, so maybe  making the UI change wouldn't be so hard.

I say make it work for now without forking the addon, and then once it works, investigate how much time it would take to change the UI.

andrewn:
@catarak This issue can be moved from "In Progress" to "Done"

PR #346, opened 2017-05-03
Basic Find using CodeMirror

andrewn:
Implements basic find functionality (#205) using the [`codemirror-revisedsearch`](https://github.com/Maloric/CodeMirror-RevisedSearch#readme) module. This renders the Previous/Next/Close buttons in the dialog, making it much easier to style.

The dialog is full-width as the addon pushes the current code panedown so no code is obscured.

screenshot 2017-05-04 01 04 27


catarak:
Just looked at how find is implemented on Glitch--they're using the CodeMirror search addon, but they've edited (at least) the `search.js` addon to have a custom search dialog: https://gist.github.com/catarak/1c9453ad7e1ca0fb303652e4c02fbac1. I think it's fine to do the same.

Issue #169, opened 2016-10-25
Add view for all of a user's S3 files


catarak:

PaliwalSparsh:
I guess i will be making changes here [here](https://github.com/processing/p5.js-web-editor/blob/master/server/controllers/project.controller.js#L75). Here the project array contains the projects and the project contains all the files in it. So we can iterate over the projects and list the files with url.

Also I felt it would be great if we would manage S3 like this - 
http://docs.aws.amazon.com/AmazonS3/latest/dev/ListingKeysHierarchy.html

catarak:
I definitely think there is room to improve the S3 querying/storage, since when I implemented the S3 uploading I wasn't thinking about querying. Do you think it makes sense to store files on S3 by username and project id? Then, for example, say you wanted to upload a file `dog.jog` to your project with the id `abc123`, and your username is `catarak`:
`catarak/abc123/dog.jpg`
Then, we could use this prefix querying to get all files for a specific user just by querying with `catarak`.

catarak:
I guess this means we'd also have to write a script to migrate all of the existing files to the new naming scheme, but I don't think that would be too hard. We'd just loop through all of the projects stored in MongoDB, look at the files that have a URL, change the URL, and then update the file with the new URL.

PaliwalSparsh:
Exactly i thought about it similarly.
`<username>/<project_id>/<filename>`

Outcome of our investment in establishing hierarchy - 
Doing this will be really helpful for [**listing objects**](https://www.npmjs.com/package/s3#clientlistobjectsparams) present in S3 with a query, this was the place where I saw the use of having hierarchy could help. Though the same listing operations can be done by us on the server side only and we need not move to S3 ie by just looking at files with url's(as we plan to do in this issue). I don't think things like deleting or downloading files will be affected on introducing hierarchy much as they require directly the file key.

But surely more structured S3 won't hurt.

catarak:
I just realized that if a user uploads a file and the project has never been saved, then the project doesn't have an id. So I think we'll have to structure these like `<username>/<filename>`. I think this is fine anyway since we want to enforce a limit on the total size of the files a user can upload, not just the files per project.

PR #356, opened 2017-05-18
Create Asset List View and refactor overlay code

catarak:
This is a fix for issue #169. This creates a view for all of a user's assets on S3, and refactors the overlay component code a bit. Some open questions I have:

- Right now a user's assets are being stored on S3 in a folder that is their user id. Would it make sense to also store it in a folder that is the project id, i.e. <user_id>/<project_id>/<asset_name>? There's some discussion about this on #169. I say this because matching assets to their respective projects is a little slow, but maybe that is just fine.
- The overlay code is better but I think there's some redundancy in the redux state. I think we need _one_ variable that contains whether or not an overlay is visible, and another variable (probably an enum, which in JS can just be a string) that is the contents of the overlay. This can probably be a separate PR. 
- Should users be able to delete assets from this view? This brings up the question, if a user deletes an asset from this view, does the asset get deleted from the sketch too? Would it instead make sense to restrict deletion to within a sketch only?


client/modules/IDE/components/AssetList.jsx
```diff
@@ -0,0 +1,87 @@
...
+            <tr>
+              <th className="asset-list__delete-column" scope="col"></th>
+              <th>Name</th>
+              <th>Size</th>
+              <th>View</th>
+              <th>Sketch</th>
+            </tr>
+          </thead>
+          <tbody>
+            {this.props.assets.map(asset =>
```
andrewn: It might be nice to show a short message when the user has no assets and when the asset list is being fetched? Maybe it's quick enough not to be an issue but initially I was looking at the Overlay wondering if I should be seeing anything.

client/constants.js
```diff
@@ -1,3 +1,5 @@
+// TODO Organize this file by reducer type, ot break this apart into
+// multiple files
```
andrewn: 👍  I think having different files would be good as I struggled a bit figuring out where new action constants should go.

andrewn:
This looks really good! I made some small inline comments.

## Overlay

I think the `Overlay` refactoring is great and cleans things up a lot. I agree about having an Overlay Enum. If it's set then we can display the enum.

## Uploading same asset multiple times

I uploaded the same `pic.jpg` multiple times to the same sketch and it appeared 3 times in the project file browser. However, it is only listed once in the Asset List. I checked the S3 Developer Console and there are 3 assets listed there.

screenshot 2017-05-25 12 08 43

## Asset listing being slow / storing by project

> I say this because matching assets to their respective projects is a little slow, but maybe that is just fine.

Would this actually speed up the displaying of the list as we'd still need to fetch all the projects and assets to be displayed? This feels like something we could do later? I think that having some sort of loading spinner UI whilst the client is waiting for the data from the API might make it seem snappier?

## Deleting assets from Asset List view

My expectation would be that if I delete an asset from this view then it would disappear from the project view too. It might be enough to allow the Asset List columns to be sortable so that users can find what they're looking for and then go straight to the project to manage the assets for that project?

client/modules/IDE/components/AssetList.jsx
catarak: I agree that it's slow and it may be nice to have a loading state for this view, which could also be used for the `SketchList` view as well.

Also agree that this should handle the case in which there are no assets 😸

client/constants.js
catarak: Yeah this file has gotten crazy, and can definitely be broken apart. Not sure if it would be better to put this in a folder called "constants" or maybe follow the "ducks" redux design pattern.

catarak:
I purposefully hid the delete icon as I wasn't sure if it made sense to add the ability to delete an asset here, but I think it doesn't make sense so I'll remove it.

I'm not sure why the image you uploaded multiple times isn't appearing. That seems like a bug 😈

PR #359, opened 2017-05-25
Find

andrewn:
This implementation supersedes #346 with a forked version of Glitch's CodeMirror search add-on.

Features:
- regular expression search
- case sensitive search
- whole word search
- highlights all matches in the editor view
- next/previous buttons scroll to the result and highlight in a different colour
- `metaKey-F` triggers the dialog
- `metaKey-G` moves to the next match, `metaKey-Shift-G` moves to the previous

### DIFF
--- a/client/modules/IDE/components/Editor.jsx
+++ b/client/modules/IDE/components/Editor.jsx
@@ -24,6 +24,9 @@ import '../../../utils/htmlmixed';
 import '../../../utils/p5-javascript';
 import Timer from '../components/Timer';
 import EditorAccessibility from '../components/EditorAccessibility';
+import search from '../../../utils/codemirror-search';
+
+search(CodeMirror);
 
 const beautifyCSS = beautifyJS.css;
 const beautifyHTML = beautifyJS.html;
--- a/client/modules/IDE/components/Editor.jsx
+++ b/client/modules/IDE/components/Editor.jsx
@@ -24,6 +24,7 @@ import '../../../utils/htmlmixed';
 import '../../../utils/p5-javascript';
 import Timer from '../components/Timer';
 import EditorAccessibility from '../components/EditorAccessibility';
+
 import search from '../../../utils/codemirror-search';
 
 search(CodeMirror);
@@ -40,6 +41,16 @@ const downArrowUrl = require('../../../images/down-arrow.svg');
 const unsavedChangesDotUrl = require('../../../images/unsaved-changes-dot.svg');
 const rightArrowUrl = require('../../../images/right-arrow.svg');
 const leftArrowUrl = require('../../../images/left-arrow.svg');
+
+const metaKey = (() => {
+  if (navigator != null && navigator.platform != null) {
+    return /^MAC/i.test(navigator.platform) ?
+      'Cmd' :
+      'Ctrl';
+  }
+
+  return 'Ctrl';
+})();
 
 class Editor extends React.Component {
   constructor(props) {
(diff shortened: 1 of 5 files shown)


andrewn:
@catarak Added the "Whole Word" search so this is ready for review.

client/modules/IDE/components/Editor.jsx
```diff
@@ -34,6 +42,16 @@ const unsavedChangesDotUrl = require('../../../images/unsaved-changes-dot.svg');
 const rightArrowUrl = require('../../../images/right-arrow.svg');
 const leftArrowUrl = require('../../../images/left-arrow.svg');
 
+const metaKey = (() => {
```
catarak: there's also a similar function in `KeyboardShortcuts.jsx`... should this be moved to a utility function?

client/modules/IDE/components/Editor.jsx
```diff
@@ -75,7 +94,10 @@ class Editor extends React.Component {
       'Cmd-Enter': () => null,
       'Shift-Cmd-Enter': () => null,
       'Ctrl-Enter': () => null,
-      'Shift-Ctrl-Enter': () => null
+      'Shift-Ctrl-Enter': () => null,
+      [`${metaKey}-F`]: 'findPersistent',
```
catarak: could you put these keyboard shortcuts in the keyboard shortcut component?

client/utils/codemirror-search.js
```diff
@@ -0,0 +1,424 @@
...
+
+    return regexp;
+  }
+
+  var queryDialog = `
+    <h3 class="CodeMirror-search-title">Find</h3>
+    <input type="text" class="search-input CodeMirror-search-field" placeholder="Search for" />
+    <div class="CodeMirror-search-actions">
+      <div class="CodeMirror-search-modifiers button-wrap">
+        <button
```
catarak: it's possible for a user to select all of these options simultaneously, which I think is a bug. it doesn't make sense to select more than one of them, does it?

also from a accessibility perspective, should these be checkboxes instead of buttons?

client/modules/IDE/components/Editor.jsx
```diff
@@ -21,6 +25,10 @@ import '../../../utils/p5-javascript';
 import Timer from '../components/Timer';
 import EditorAccessibility from '../components/EditorAccessibility';
 
+import search from '../../../utils/codemirror-search';
+
+search(CodeMirror);
```
catarak: lol at codemirror

catarak:
added a few comments!

client/utils/codemirror-search.js
andrewn: ### Selecting options

These can all be used together. 

- Case Sensitive
- Whole word
- Regex: Does a regex search
- Regex + Case Sensitive: Adds `/i` flag
- Regex + Whole word: Adds `\b` to expression

### Checkbox instead of buttons

Makes sense to me. Are there any examples in the code of using a styled checkbox input? 

Alternatively, could I use the [`aria-checked` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_checkbox_role)?

catarak: ugh sorry for the delay in replying to this. usually what I do for styled checkboxes is style the label for the checkbox rather than the checkbox itself:
https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/components/Preferences.jsx#L263
and
https://github.com/processing/p5.js-web-editor/blob/82207a50d39d87872c257a441609baef8f817b9d/client/styles/components/_preferences.scss#L142
but `aria-checked` is probably okay too.

catarak: read a little more, `aria-checked` is okay but I think the buttons also need `role=checkbox`.

andrewn: Added `role="checkbox"` to these modifiers.

client/modules/IDE/components/Editor.jsx
andrewn: SADFACE.

andrewn:
@catarak Ok, ready to review again!

Issue #357, opened 2017-05-22
Auto-refresh isn't honored when loading new example sketch


Spongman:
- load an example sketch
- hit play
- check 'Auto-refresh'
- load another example sketch

the old sketch is still running.

PR #365, opened 2017-06-01
Current sketch will stop playing if a new example is opened (issue #357)

zrispo:


catarak:
I'm not sure this is the right place to dispatch this `stopSketch` action. There's a larger issue here, which is that anytime a user opens a different sketch or navigates to a different page, the sketch should stop. Maybe this action should be dispatched with the route changes?

client/components/Nav.jsx
```diff
@@ -91,7 +91,10 @@ class Nav extends React.PureComponent {
           })()}
           <li className="nav__item">
             <p className="nav__open">
-              <Link to="/p5/sketches">
+              <Link
+                to="/p5/sketches"
+                onClick={this.props.stopSketch}
```
catarak: with the `onRouteChange` handler I don't think you need this anymore!

zrispo: Oh that's right - there may be some extra `stopSketch` calls in the code, I'll go through and clean them up

zrispo:
Got rid of the extra `stopSketch` calls, should be ready to go ~

Issue #212, opened 2016-11-23
loadImage and non-literal file names


kjhollen:
A couple students in our class have seen an issue where they try to load a sequence of images but the paths don't resolve. For example they have images like:

folder/file-1.png
folder/file-2.png
folder/file-3.png

and want to use a for loop to load all the images in a folder, iterating over the numbers in the file name. I have a couple of examples that reproduce this:

[http://alpha.editor.p5js.org/projects/rJtoro7Gg](http://alpha.editor.p5js.org/projects/rJtoro7Gg)

[http://alpha.editor.p5js.org/projects/r1f0wsmzg](http://alpha.editor.p5js.org/projects/rJtoro7Gg)

They don't work in the web editor, but downloading the sketch and serving the files locally works as I'd expect (a grid of images is visible).

lmccart:
here's an even simpler example. so it really seems to be a problem with constructing strings from variables.
http://alpha.editor.p5js.org/projects/rkGtQ3QMe

catarak:
Before the sketch is rendered, it's replacing all of the local file names with the remote url for the files on S3. Since the names aren't in the actual text of the JS file, it's not able to replace it. Maybe there's a way to resolve these file names at runtime, though I'm not sure how.

catarak:
@yining1023 and I also encountered this issue when working on #203. A few of the sound examples, (e.g. [this reverb example](http://p5js.org/examples/sound-reverb.html)), don't use the literal file name, but will append an extension to the file name as an argument of `loadSound` based on the arguments of `soundFormats`.

I've been thinking about a general way to solve this, and I think, rather than swapping the literal file names in text with the S3 urls, the files need to be requested server-side and then served based on the URL that the preview is requesting, i.e. `http(s)://alpha.editor.p5js.org/projects/<project_id>/<path_to_file>`

JMandel1027:
Hi guys, I came across a similar issue while prepping for a class I'm teaching with p5.js. It seems the web editor cant load external image files from the web, I tried with both png and jpg's. 

`Access to Image at 'https://rlv.zcache.com/happy_smiley_face_with_sunglasses_sticker-rf1a85167493343bd9b1b4943541b5553_v9waf_8byvr_324.jpg' from origin 'http://alpha.editor.p5js.org' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://alpha.editor.p5js.org' is therefore not allowed access. `

[...]

Issue #219, opened 2016-12-01
Fix example: three examples under Sound category


yining1023:
`Sound : Reverb`, 
`Sound : Convolution Reverb`
`Sound : Filter LowPass`
are broken as discussed in #212.

PR #366, opened 2017-06-01
Server can serve individual assets from projects

zrispo:
Fixes issues #212 and #219.

### DIFF
--- a/client/modules/IDE/components/PreviewFrame.jsx
+++ b/client/modules/IDE/components/PreviewFrame.jsx
@@ -249,7 +249,8 @@ class PreviewFrame extends React.Component {
     let jsFileStrings = content.match(STRING_REGEX);
     jsFileStrings = jsFileStrings || [];
     jsFileStrings.forEach((jsFileString) => {
-      if (jsFileString.match(MEDIA_FILE_REGEX)) {
+      console.log(this.props.setBlobUrl);
+      /* if (jsFileString.match(MEDIA_FILE_REGEX)) {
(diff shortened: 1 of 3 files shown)


catarak:
this isn't working for me. when i run the examples from #219, it doesn't appear to be hitting the route that you created. it appears to be directed to a different route and this is what i'm seeing in developer tools:
screen shot 2017-06-06 at 11 09 58 am
looks like it's returning one of the html pages!

zrispo:
Added a fix so that assets in subfolders can be served - I think this was the problem you were running into

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -249,7 +249,8 @@ class PreviewFrame extends React.Component {
     let jsFileStrings = content.match(STRING_REGEX);
     jsFileStrings = jsFileStrings || [];
     jsFileStrings.forEach((jsFileString) => {
-      if (jsFileString.match(MEDIA_FILE_REGEX)) {
+      console.log(this.props.setBlobUrl);
+      /* if (jsFileString.match(MEDIA_FILE_REGEX)) {
```
catarak: I think it's still a good idea to serve as many files as possible from a blob url rather than server side—this will keep AWS costs down 😄

catarak:
hmmm also I'm still getting 404 errors for all of the assets, not sure if it's just me though.

catarak:
cool, this is working for me! also now for some reason the examples are working too \ 😄 /

Issue #44, opened 2016-08-11
Download Sketch as zip - Save Library Files


catarak:
It may cause issues if the library files are still just linked as CDNs, especially if a user expects to be able to run the project offline.

indefinit:
+1 to this enhancement!

catarak:
should this be the default? should this be an option that a user can turn on?

lmccart:
my intuition says let's just do this by default, and not add more options that could potentially confuse

meiamsome:
@catarak Looks like this issue has been solved by #367 and should be closed?

catarak:
yep!

PR #367, opened 2017-06-01
Bundle libraries when project is downloaded as zip

zrispo:
See issue #44 ~


catarak:
This works if a a user has loaded in their `index.html` only `p5.js`, but what if a user has other scripts included? At some point there will need to be a library management system, but I think the solution for now is a little more involved:

1. Find all externally hosted files in the `index.html` file.
2. Download them and include them in the zip file.
3. Change all paths in `index.html` to a local path to the file.

PR #369, opened 2017-06-05
Email verification

andrewn:
Send emails to user's email address containing a link so we can verify they own it.

- [x] Generates nice emails using [MJML](https://mjml.io/)
- [x] Convert "Reset Password" to use MJML template
- [x] Send verification email on Sign Up with `/verify` link
- [x] Verification endpoint that saves verified status if token is valid
- [x] Implements `/verify` UI on client that sends token to API
- [x] Displays current verification status on Account page
- [x] Button to trigger email confirmation to be re-sent
- [x] Nicer UI on Account page for showing current status
- [x] When changing email address, re-set verification status and send another email
- [x] Choose JWT or token-in-db approach

## Env vars

The following env vars are needed for this feature to work:

```
EMAIL_SENDER=<email-address-to-send-from>
EMAIL_REPLY_TO=<email-to-reply-to>
MAILGUN_KEY=<mailgun-api-key>
MAILGUN_DOMAIN=<mailgun-domain>
EMAIL_VERIFY_SECRET_TOKEN=whatever_you_want_this_to_be_it_only_matters_for_production
```


andrewn:
@catarak This is functional now so it'd be good if you could take a look. There are a couple of things left on the list that I think we need before it's done.

Can you think of anything else?

andrewn:
I've just noticed, the email verification code that was in a branch (which I've re-used parts of) uses JWT as the token and the Password Reset code stores the token in the database. 

We should also decide which approach we want to use.

andrewn:
@catarak Ready for review.

catarak:
are JWT's okay to use? i feel like every day i read a blog post that says they are the best ever and the next day i read one about how they are highly insecure.

andrewn:
I've read around the subject and there have been some vulnerabilities reported with JWTs but the library we're using is not vulnerable to that.

[This article has some best practices](https://stormpath.com/blog/jwt-the-right-way) that we're following:
- encrypting the user's email address so it can't leak
- setting an expiration time
- ensuring key is only on server and not shared

Saying that though, I think that we should use the same method for Password Reset and Email Confirmation. I also think that having something that's less exotic is also good.

So, I'll update this to store Email Confirmation in the database and not via JWT.

andrewn:
Ok, I've reworked the approach to mirror that of password reset. We generate a token and store it in the database along with an expiry time. We send the emails:
- on sign up
- when the user changes their email address
- when the user requests it from the account page

catarak:
woohoo this is great! there's a couple of server-sde linting errors when you run `npm run lint` but otherwise it's good and works.

question though, what's `EMAIL_REPLY_TO`?

andrewn:
@catarak The `h:Reply-To` field is used if you want the reply-to address of the email to be different from the sender's address. For example, we might send the email from `noreply@p5js.limo` but want user's to be able to reply to `support@p5js.computer`.

If you don't think we need it I can remove it.

andrewn:
Linting errors fixed. ⚡️

catarak:
sorry for the slow update on this, but yeah, I don't think we need the different reply to email. i'd prefer if we didn't have a reply to email anyway as we don't have a Processing Foundation support person who can handle those kind of emails

andrewn:
I've removed the "reply to" feature.

PR #376, opened 2017-06-28
Bundle libraries when project is downloaded as zip (New approach)

zrispo:
Switched to a different fork because I had to start over and try the new approach @catarak suggested! - the old PR is here: https://github.com/processing/p5.js-web-editor/pull/367


catarak:
this works for me! when I run `npm run lint` there are a few errors, could you fix these?

zrispo:
Fixed the linting errors in `project.controller.js` - there are still some errors happening in `aws.controller.js` but those were there already. I can fix them in this PR if that would be helpful though!

catarak:
`bundleExternalLibs` doesn't differentiate between local files (i.e. ones already stored in mongo) vs. external ones... would it be possible to not try to request local files, to remove unnecessary errors in the logs?

zrispo:
Added a check for valid URLs using https://www.npmjs.com/package/is-url

Had to add this library to package.json, let me know if that's ok!

PR #399, opened 2017-07-18
Update Find styling to new design

Katy310:
Fixes styling of the Find dialog box.

### DIFF
--- a/client/styles/abstracts/_variables.scss
+++ b/client/styles/abstracts/_variables.scss
@@ -44,6 +44,7 @@ $themes: (
     input-text-color: #333,
     input-border-color: #b5b5b5,
     about-list-text-color: #4a4a4a,
+    search-background-color: #ebebeb
   ),
   dark: (
     logo-color: $p5js-pink,
--- a/client/styles/components/_editor.scss
+++ b/client/styles/components/_editor.scss
@@ -85,9 +85,9 @@
 */
 
 .CodeMirror-dialog {
-  position: absolute;
+  position: fixed;
   top: 0;
-  right: 330px;
+  right: 1225px;
 
   z-index: 10;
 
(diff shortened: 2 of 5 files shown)


client/styles/components/_editor.scss
```diff
@@ -85,9 +85,9 @@
 */
 
 .CodeMirror-dialog {
-  position: absolute;
+  position: fixed;
```
catarak: This styling isn't working for me. Given that screen widths are different on different devices, I would use a percent value rather than a fixed pixel value. Locally, I've changed this to
```scss
.CodeMirror-dialog {
  position: fixed;
  top: 0;
  left: 50%;
  margin-left: - #{365/2/$base-font-size}rem;

  z-index: 10;

  width: 100%;
  max-width: #{365 / $base-font-size}rem;
 // ... the rest of the styles
}
```

client/styles/abstracts/_variables.scss
```diff
@@ -44,6 +44,7 @@ $themes: (
     input-text-color: #333,
     input-border-color: #b5b5b5,
     about-list-text-color: #4a4a4a,
+    search-background-color: #ebebeb
```
catarak: You should add this variable to the other themes, `dark` and `high-contrast`. For dark I would use the same color but for high contrast i would use `$white`.

server/controllers/user.controller.js~
```diff
@@ -0,0 +1,346 @@
+import crypto from 'crypto';
```
catarak: You accidentally included this file in the commit—it should be deleted!

catarak:
[Here's a link to a patch](https://www.dropbox.com/s/bihj1go7qzo4dwd/0001-code-review-changes.patch?dl=0) if you want to look at changes I made locally to get this working.

Issue #400, opened 2017-07-19
Sketches with for loops with certain syntax errors cause the editor to stop previewing sketches


zrispo:
Here's an example of a syntax error that will not just throw an error when it's run, but also break the editor (sketches won't run until you refresh the page and fix the code):

```
function setup() { 
  createCanvas(400, 400);
  
  for(var i = 0 ; i < 1000; i++) {
    {
  
} 

function draw() { 
  background(220);
}
```

I think the issue is that the loop-protect library isn't built to handle code with syntax errors. 

I'll look into finding a way around this so it doesn't crash the editor!

zrispo:
Also here's the error that loop-protect causes:

![screen shot 2017-07-19 at 5 24 39 pm](https://user-images.githubusercontent.com/4970417/28390247-393996ae-6ca7-11e7-90da-1403ceaea008.png)

catarak:
hmmmm yeah I'm not sure how to fix this one. the library we're using to check the code for infinite loops is definitely not optimal, but i'm not sure what else is out there.

zrispo:
Yeah I agree, the loop checking is really helpful & I think it should def stay!

As a hacky solution, it might be safe enough to just not check for infinite loops in code with syntax errors and just skipping that step altogether (the sketch usually won't run to completion anyway if there's a syntax error I think)

catarak:
The problem is that this becomes a chicken or the egg situation, where having an infinite loop causes issues with finding syntax errors, and syntax errors causes issues with loop detection  ¯\_(ツ)_/¯

We could try checking for syntax errors before putting the code through the loop detection, and see if that doesn't slow things down/cause other errors.

catarak:
just pasted this code into the web editor and it seems like this bug is no longer happening ✨

PR #406, opened 2017-07-27
Possible fix for loop-protect crashing the page when project has syntax errors (#400)

zrispo:
Here's a possible solution I discussed with @catarak ~

If the currently open sketch has any sort of syntax error, you get a warning (a little alert box) before the project runs. This ensures that no bad code gets sent to the preview window!

Let me know if this is a reasonable way to deal with this issue!


catarak:
The alert doesn't appear in live code mode. The alert also doesn't appear if you use the keyboard shortcuts (command + return). The alert also appears for syntax warnings rather than just syntax errors—I think the difference between warnings and errors is that warnings don't cause the JS to crash. 

I'm not sure that an alert is the best solution for this either. I like highlighting the lines of the errors more. Maybe we can just use the loop protect library more intelligently?

catarak:
i'm going to close this, since I think this is fixed in #456

Issue #251, opened 2017-01-07
Aid debugging by highlighting runtime errors console on in code editor


thepackrat:
It would be a useful aid to debugging programs if runtime errors printed to the console that had line numbers were also presented to the user as a third highlight colour (blue? green?) with mouse over to the errors being the js console error generated.

catarak:
Just to make sure I'm understanding what you mean, you're thinking of presenting runtime errors in a similar way to the linting errors? This is a cool idea. Do you think it would be difficult to distinguish between linting and runtime errors? is color enough?

thepackrat:
The traditional way to do this in an IDE would be a little alert symbol (triangle sign with ! in it) printed against the line where the error was identified. I think that would work pretty well.

zrispo:
I'm working on this!

PR #408, opened 2017-07-27
Highlight runtime errors in code editor (issue #251)

zrispo:
The styling for the line highlighting may need some work, but here it is!

screen shot 2017-07-27 at 2 18 05 pm


lmccart:
ahh this is great!

catarak:
I just pushed a commit that should fix the errors being on the wrong line. As for the styling on this, I would

1. Extend the line highlight to the gutter
2. Remove the border radius
3. Make sure the line highlight color is the same as the gutter color, so it looks continuous.

catarak:
i updated the styles for this on other commits

Issue #421, opened 2017-08-25
Log in screen returns to Sign up screen


aferriss:
1. Go to https://alpha.editor.p5js.org/signup
2. Click Log In at the bottom
3. Fill in your info and try to Log in

Afterwards you'll be redirected back to the Sign up screen. Clicking the X in the upper right corner cycles the Log In and Sign up screens back and forth. Going to alpha.editor.p5js.org will bring you to the editor and you'll be signed in.

catarak:
Thanks for submitting this! I can reproduce. After logging in or signing up, the app is supposed to redirect you to the last page you were on before authenticating. There's obviously a bug there 😄

PR #431, opened 2017-09-02
Login and signup page not visible after login

Akarshit:
This could be a fix for #421 
Doesn't cover all the cases, just makes sure that Login and Signup View are not rendered after the user is authenticated.

### DIFF
--- a/client/modules/User/pages/LoginView.jsx
+++ b/client/modules/User/pages/LoginView.jsx
@@ -27,6 +27,10 @@ class LoginView extends React.Component {
   }
 
   render() {
+    if (this.props.user.authenticated) {
+      this.gotoHomePage();
+      return null;
+    }
     return (
       <div className="form-container">
         <div className="form-container__header">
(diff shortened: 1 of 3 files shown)


catarak:
thanks for this! it probably also makes sense to redirect server-side as well, which I think with `passport` is as simple as checking if `req.user` exists in the `/login` and `/signup` routes.

Akarshit:
@catarak Ya that would definitely make sense.

server/controllers/session.controller.js
```diff
@@ -7,6 +7,10 @@ export function createSession(req, res, next) {
       return res.status(401).send({ message: 'Invalid username or password.' });
     }
 
+    if (req.user) {
```
catarak: this will redirect the API, which is not what we want. the only file you should need to change is `server/routes/server.routes.js`.

server/controllers/user.controller.js
```diff
@@ -25,6 +25,11 @@ export function findUserByUsername(username, cb) {
 const EMAIL_VERIFY_TOKEN_EXPIRY_TIME = Date.now() + (3600000 * 24); // 24 hours
 
 export function createUser(req, res, next) {
+  if (req.user) {
```
catarak: again, this will redirect the API rather than the route the user sees.

PR #432, opened 2017-09-04
Docker-based development setup

francisli:
Hi there,

I was checking out the great work you're doing with the p5.js web editor, and thought I'd contribute back this Docker-based development setup I used to try it out on my own computer.

If you're interested, this could make for a more streamlined development setup for contributors, as well as ensure everyone is running the EXACT same setup as everyone else- exact same Node version, exact same Mongo version, etc- independent of underlying platform. As I mention in the README notes I added for it, it also helps isolate work on this project from any other Node projects you may have on your computer (I do all of my web development work in isolated Docker-based configurations).

If not, no worries, feel free to just close the PR!

A few other notes on the changes in this PR- I use ```yarn``` instead of ```npm``` these days for all of my Node based projects to install Node modules, which resulted in adding a ```yarn.lock``` file. It's significantly faster and less buggy than npm 3 when it comes to resolving exact dependency versions- although they've since fixed most of their problems, I've had issues in the past where npm could install different versions of dependent node modules in different deployments, causing unexpected variations in behavior. The ```yarn.lock``` file helps yarn ensure that everyone who uses yarn to install their Node modules installs the EXACT same versions, period. If you're not interested in switching to yarn, I'd at least recommend upgrading npm to npm 5 and using package-lock files (the new npm equivalent, or npm shrinkwrap on older versions).

Finally, even though it's supposed to be the default, for some reason I had to explicitly add the eslint rule for no-return-assign to get past some linter errors that were breaking the build for me.


catarak:
Thank you so much for doing this! I am very down with any and all input on the dev ops front, as it's the thing I have the least experience with. I'll test this locally and see how easy it is to work with. I'm trying to keep the installation of this project locally as simple as possible, and this seems to not really add any more complication.

A couple other questions:

1. Would it then make sense to use Docker in production? Or at least just switch the production installation instructions to use yarn?
2. What are the benefits of yarn over npm 5 and package-lock files? I've seen enough projects using yarn that I'm cool with it, I just want to be 100% cool with it.

francisli:
To be honest, I haven't tried npm 5 so I can't really offer any pros/cons. I switched to yarn back when npm 3 came out slow and buggy, and it just worked, and never looked back. yarn is used by Facebook in their production deployments, which I think makes it a good candidate for ongoing support and reliable releases.

As for production deployments, I'd definitely recommend using either yarn or npm lock files to ensure a consistent and predictable deployment of node modules. 

I have less experience using docker in production, and while it might provide some flexibility with hosting platforms/management (including using AWS' built-in support for Docker), since the web app has a dependency upon AWS for S3 storage anyway, it's not that a big advantage. Personally, I find that with production environments you'll generally want to use fully hosted versions of the dependent services (dbs, etc) which come with support, backups, maintenance, etc. rather than community supported images running in containers which you'll still have to administer yourself.

boazsender:
Just chiming in to say +1 I would for sure use this for development as a first time contributor. Nice one!

catarak:
okay i've been way behind on a bunch of work and i'm finally getting around to looking at this. it rules! i'm gonna remove the yarn dependency from this PR due to the NPM 5 release and instead add a `package-lock.json` file. and also update the node version to `8.9.0`.

francisli:
Just a note that another thing you can do with this is to push pre-built Docker images up to a public repository (i.e. on Docker Hub/Cloud). Then contributors just do a `docker-compose pull` and it will pull it down and you can be ready to go without having to wait for the full images to build, again giving everyone the EXACT same development environment as everyone else. Note that, depending upon when you build your image, you may get different versions of dependencies installed from package repositories not managed by npm (i.e. installed by apt-get etc).

catarak:
that's good to know! once i get this working i will probably push the docker image.

which brings me to my next point, which is that i haven't been able to get this to build. when i run `docker-compose build`, it hangs forever, and when i run `docker-compose --verbose build` i get the error `requests.exceptions.ReadTimeout: UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)`. i've also tried increasing the `COMPOSE_HTTP_TIMEOUT` but it still is timing out. not really sure what's going on.

catarak:
UPDATE: i deleted all docker containers and reset docket to its default settings. when i run `docker-compose build`, it hangs on the step `Step 3/7 : ADD . $APP_HOME`. i assume there's an error that's not getting logged to std out or std error.

catarak:
i'm still working on getting this up and running but i've made some progress. i'm currently running into an issue with `node-sass` and Linux 64 bit, which i am trying to solve by updating some of the outdated node nodules. we will see if that fixes it.

another thing to note is that the docker installation takes up a lot of disk space. i just had to delete a bunch of files on my computer, which is totally cool but maybe something worth noting on the readme.

anyway, i'm going to keep updating this issue as i figure this out, which hopefully will be helpful to someone in the future, and is hopefully not too annoying.

catarak:
okay, finally got this to work! not sure if there is a way to fix this or not, but when i ran `docker-compose up` for the first time, node-sass was broken, so i had to open a shell to the docker server and run `npm rebuild node-sass --force`. does this need to be added to the Dockerfile or is there maybe a way around this?

PR #456, opened 2017-10-21
resolves syntax error freezing behavior in #428 

### LINKED ISSUE #428: Can't save sketch (sketch corrupted)
A student had a sketch in class today with some errors, which we fixed, but it would not run. It seems that maybe the root cause is that the code is not being saved on the server end, so it's running a stale version of the code instead of what's in the editor.

Original sketch link:
https://alpha.editor.p5js.org/lsm0821/sketches/r19_DZGY-

When I duplicated the sketch wouldn't save there either:
https://alpha.editor.p5js.org/kjhollen/sketches/Bk0I5t4FZ

The console complains of 'Unexpected Token ILLEGAL' error on line 8, even when it is fixed in the editor (see this sketch for repaired code http://alpha.editor.p5js.org/kjhollen/sketches/SySjoKEFW )—so I think the root of the problem is that the sketch doesn't save.

I'm not sure what the repro steps for this are; I'll update this post if I see this come up again.

boazsender:
It looks like decomment is throwing uncaught errors when it tries to strip comments from JavaScript code with syntax errors in it. This PR checks to see if there are syntax errors before decommenting and resolves #428 where users who encounter a state issue when making syntax errors in auto refresh mode.

I am also preemptively doing the same thing in this PR for loopProtect which I imagine could also error out on malformed JS.

This is a workable solution for now, but really, I think the longer term answer is to not "autorun" code with syntax errors in it at all, and to do that check based on the state of the editor (which already does all the analysis). I'm going to open another issue for that.

### DIFF
--- a/client/modules/IDE/components/PreviewFrame.jsx
+++ b/client/modules/IDE/components/PreviewFrame.jsx
@@ -162,6 +162,22 @@ class PreviewFrame extends React.Component {
     });
   }
 
+  jsPreprocess(sketchDoc) {
+    let newContent = sketchDoc;
+    // check the code for js errors before sending it to strip comments
+    // or loops.
+    JSHINT(newContent);
+
+    if (!JSHINT.errors) {
+      newContent = decomment(newContent, {
+        ignore: /noprotect/g,
+        space: true
+      });
+      newContent = loopProtect(newContent);
+    }
+    return newContent;
+  }
+
   injectLocalFiles() {
     const htmlFile = this.props.htmlFile.content;
     let scriptOffs = [];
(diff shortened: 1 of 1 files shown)


catarak:
thanks for this! this code is a little crazy right now and could be DRY'd up, and since JS can be added in any number of JS files or inline in the HTML, the loop protect is being called a few times. maybe we could add a function called `jsPreprocess` that checks for errors and runs the loop protect?

catarak:
u rule @boazsender, thanks for the update!

boazsender:
Thanks for the feedback @catarak! I just pushed a commit to make that syntax check a bit more reusable.

I still think this could be further refactored. The `addLoopProtect` func share logic with `jsPreprocess`, for example, but I think that warrants another issue bigger PR, and maybe some unit tests.

boazsender:
Wait, I think there's a problem with this PR. Please do not merge just yet.

catarak:
@boazsender i'll wait!

also i am very open to any input on unit tests and refactoring so feel free to open issues and PR's 😸

boazsender:
@catarak cool, I'll give it some thought. I think starting with an end to end test using webdriver might be a good place to start. My co

boazsender:
@catarak cool, I'll give that some thought. cc @jugglinmike.

In the meantime, @catarak I fixed my issue, and this PR is now good to merge from my perspective.

catarak:
resolved this by adding 9297496a5e2f8ddb39b8071366b88446f851073a

shiffman:
Amazing, so excited about this one!

Issue #448, opened 2017-09-29
Support static mode


sofian:
NOTE: This feature request was imported from [P5.js](https://github.com/processing/p5.js/issues/2220)

The P5.js editor does not currently support "static" mode which allows sketches that do not contain functions setup() and draw() to work.

Static mode is an interesting feature when introducing students to Processing. I wonder if it would even be possible to make it work but I'd be happy to give it a try if this is something others believe would be a benefit.

I personally feel it makes many students feel like it is an empowering feature tof Processing especially for beginners, as it allows to build confidence early on. I usually begin my classes with an empty sketch and show them that this still produces a gray-background canvas, and we build from there.

catarak:
Thanks for submitting this! I don't have a strong opinion about this, and could be convinced either way, so I'm curious for more input.

sofian:
In general, one point in favor of it is simply that it would more adequately match the "java" version of Processing. I also believe the IDE (and not the library) is the right place to implement this.

limzykenneth:
I quite like this idea, however, it would mean that there will be a difference between code written in the editor vs code written in the user's own text editor (ie. if you copy a static mode code from the web editor into a custom setup project in your own text editor it will fail).

This works for Processing because Processing's users will probably be using the editor as the code actually written are significantly pared down from actual fully compiler friendly Java and the compiling stage are also taken cared of. I feel p5.js does less out of the box being just a library itself and just some "magic" to make global mode possible.

That's what I think anyway, whether it is that much of a concern for the users, much can still be said.

sofian:
Yes this is the main drawback. No intermediate user really uses static mode, it is really more of an educational feature to introduce students to coding. So I believe that by the point users start using p5.js as a library outside of the IDE, they will be advanced enough to not use the static mode anymore, and it will be fine to just point out that static mode is not going to work in that context.

catarak:
I was looking at this https://github.com/processing/p5.js/wiki/Frequently-Asked-Questions#why-cant-i-assign-variables-using-p5-functions-and-variables-before-setup today, which made me lean towards not including a static mode (or static mode being fairly low priority), but instead explicitly including p5 outside of `setup` or `draw`:
```javascript
new p5();
print(random());
```

I think another option would be to explicitly have a 'static mode' in the editor which could be turned on by a preference option, but i wouldn't want static mode to be turned on implicitly.

francisli:
[...]

PR #488, opened 2017-12-04
#448 a first quick proof-of-concept implementation of static mode support

francisli:
I, too, am interested in having the equivalent of Processing's "static mode" for a workshop geared towards elementary school students. Although it seems like a minor and limited feature, I've always felt that it was a keen insight of the original Processing- to completely remove any distractions from the first interaction, giving students that blank slate/sheet of paper to draw upon incrementally.

This is obviously a somewhat frail implementation of it, but it does work, and it would probably work in the majority of workshop/classroom settings. If you rename the initial js file from sketch.js to something else it will break (of course, it will break your project regardless unless you know how to also modify the html file script reference). Having multiple files can also break, depending on how they are used (this is why the original Processing PDE concatenates the content from all of its .pde file "tabs" into a single source file before pre-processing).

So, I think by the time you're ready to let students open up the side tab, and start manipulating the project files, you've probably gotten far enough that they've learned function declarations and started using setup() and draw()...


francisli:
@catarak per the comments in #448 this is PR is now up to date with the latest from master, and is enabled only with an environment variable switch...

catarak:
thanks for updating this! sorry for the extremely long delay in looking at this. i'm going to put some of my thoughts on the issue itself.

catarak:
@francisli i really appreciate your work and thoughts about this. i'm hesitantly okay with the idea of enabling static mode via an environment variable. i'm not sure it's a good software development practice to enable features of a web app via environment variable, but it also seems like the side-effects would be small.

however, i don't think it's a good idea to enable static mode implicitly like this—could you instead add `<script>new p5();</script>` to the `index.html` with static mode enabled?

catarak:
@francisli i mean doing something like this: https://editor.p5js.org/cassie/sketches/rk4L89ay4

in this case, i'm able to call p5 functions outside of setup.

francisli:
@catarak as I tried to explain, I've been doing some experimentation, and although simple utility functions like random will work outside of setup with this approach, functions that draw to the screen will not. This is probably because p5js is waiting for the DOM ready event before initializing its reference to the canvas object- which means drawing code must be in setup() so that they can be called after everything is ready...

If you're uncomfortable with these changes, I'd recommend just rejecting this PR and closing it- I'll just maintain it in my fork!

catarak:
ahhh okay, got it! that's unfortunate. since there doesn't seem to be a way to make this work, except by secretly wrapping everything in a setup function, i don't think it makes sense to merge this.

Issue #123, opened 2016-10-05
text editor + canvas sometimes don't load on Safari


kjhollen:
A student emailed me about an issue where the editor failed to load in Safari. We worked around it by having her use Chrome instead. She was able to see the navigation and title bar for the sketch (play/stop, sketch name, settings), but no text editor or canvas. 

I looked at it with her in class today and the dev tools show the entire editor-preview-container div is missing. Oddly, everything is okay with Safari on my Mac. Let me know if there's any more info I can provide... or if there's a certain browser we should encourage the students to use.

catarak:
I'm going through and closing old issues. Going to close since I haven't heard this happening in a while, feel free to open a new issue!

Macintosh-Fan:
Still a bug

PR #499, opened 2017-12-07
Fixes #497 by adding support for an optional new env variable, S3_BUCKET_URL_BASE

### LINKED ISSUE #497: Support US East (N. Virginia) region for S3 buckets, CNAME for buckets
#### Nature of issue?

- [x] Found a bug
- [x] Existing feature enhancement
- [ ] New feature request

#### Details about the bug: 

I've created my bucket in US East (N. Virginia) so that its in the same region as my server. However, since the code is manually constructing URLs based on the region code, this is failing as the region code is "us-east-1", but being the very first s3 region, the host is simply "s3.amazonaws.com".

This handling of the S3 bucket URLs also prevents using CNAMEs for the bucket urls, which is what I'd really like to do so that the bucket can also be put behind a CDN.

francisli:
So, with these changes, you can support us-east-1 by just manually setting the S3_BUCKET_URL_BASE to https://s3.amazonaws.com/<bucketname>. Or, if you've got a CNAME mapped to your bucket, you can specify that instead.

Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,24 @@ If you don't have the full server environment running, you can launch a one-off 
 1. `$ npm install -g pm2`
 2. `$ pm2 start ecosystem.json`
 
+## Optional S3 bucket URL base configuration
+
+If your S3 bucket is in the US East (N Virginia) region (us-east-1), you'll
+need to set a custom URL base for it, because it does not follow the standard
+naming pattern as the rest of the regions. Instead, add the following to your
+environment/.env file:
+
+```S3_BUCKET_URL_BASE=https://s3.amazonaws.com```
+
+If you've configured your S3 bucket and DNS records to use a custom domain
+name, you can also set it using this variable. I.e.:
+
+```S3_BUCKET_URL_BASE=https://files.mydomain.com```
+
+For more information on using a custom domain, see this documentation link:
+
+http://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html#VirtualHostingCustomURLs
+
 ## Accessibility Guidelines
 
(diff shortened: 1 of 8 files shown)


catarak:
thanks for this! can you add a a description of this variable to the README?

francisli:
Sure thing... added a new section to the README and pushed it up...

catarak:
thanks!

simonlimon:
I see that when you set the `s3bucket` variable you use `S3_BUCKET_URL_BASE` as the full URL, which should include the bucket name in it: 
```js
const s3Bucket = process.env.S3_BUCKET_URL_BASE ||
                 `https://s3-${process.env.AWS_REGION}.amazonaws.com/${process.env.S3_BUCKET}/`;
```
So maybe we should rename it to `S3_BUCKET_URL` and update the README to be clear about this, or fix the code so that it uses it only as the base url. What do you think?

dhoizner:
This just bit me in the form of a few hours wrestling with what i thought was AWS being stubborn about CORS... @catarak are you hosting with buckets that are using this variable, or do you think it's safe enough to change it to actually be a base rather than the full url?

francisli:
@dhoizner  Sorry to see this conversation so long after it was started (somehow Github notifications aren't always reaching me)...

I think @simonpfish is correct that better naming and documentation of the variable is needed. However, updating the code to always append it with the S3_BUCKET would break its use with custom CNAME mapped buckets.

dhoizner:
@francisli ah, good call! For now I added a bit to the README in #1475

catarak:
> This just bit me in the form of a few hours wrestling with what i thought was AWS being stubborn about CORS...

Sorry the documentation wasn't clear!! Thank you for updating it.

PR #501, opened 2017-12-15
Fixes #500 removes CSRF token setting in cookie and validation on req…

### LINKED ISSUE #500: xsrf protection unnecessary...?
#### Nature of issue?

- [ ] Found a bug
- [x] Existing feature enhancement
- [ ] New feature request

So, I see that XSRF protection is enabled with csurf on the server side, and setting a cookie containing that token for the axios library on the client side to pick up and send on each request back...

So, one issue with this implementation is that it prevents CDN caching of the root (/) homepage, since everyone needs to get a token in order to be able to register/log in. Since the root is going to be the highest trafficked page, this will prevent being able to utilize the CDN to reduce load on the server.

Since the API is a JSON-only API to be invoked using AJAX calls (the requestOfType middleware returning an error otherwise), is this really necessary? Browser forms cannot post JSON (so you can't just put the server url into the action param of a FORM on a malicious site), and AJAX calls will be restricted by CORS (so you can't script it otherwise). So, I'm not sure there's any XSRF vulnerability to be addressed...?

francisli:
…uest

Thank you!

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -7,7 +7,6 @@ import session from 'express-session';
 import connectMongo from 'connect-mongo';
 import passport from 'passport';
 import path from 'path';
-import csurf from 'csurf';
 
 // Webpack Requirements
 import webpack from 'webpack';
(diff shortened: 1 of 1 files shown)


catarak:
oops don't know why i didn't merge this sooner

Issue #137, opened 2016-10-11
Include text output in full screen mode/while sharing


MathuraMG:
Currently, while viewing the sketch in full screen/while sharing the accessible text output is not present.
**Possible solutions :**
Include a textOutput parameter while viewing the sketch in fullscreen
[http://alpha.editor.p5js.org/full/<sketch_id>?textOutput=1]
And then in the share window, update the UI so there's a button to add the query string parameter to the URL.

catarak:
Not sure if this will be relevant with some new designs coming... would it make sense to include the accessible text output in an embedded view? Maybe that could be a query string parameter for the embed view?

machinic:
@MathuraMG I've tried to recreate the current experience by turning accessible text mode on (in regular view) and also going into full screen mode (through a share full screen URL) to better understand what is missing but I may not have the best sketch to use as an example (as so much is currently missing from that full-screen experience).  If you had an example or could describe the issue further then I'd be able to better bake in a solution to the design updates or weigh in on catarak's suggestion. thanks!

catarak:
# Design
<img width="512" alt="screen shot 2017-07-11 at 10 16 54 am" src="https://user-images.githubusercontent.com/6063380/28072829-1da93100-6622-11e7-82b9-567da9905676.png">

Need to make sure that all of the accessibility features work on the full screen view!

PR #508, opened 2018-01-30
Accessibility - Transition from interceptor to accessibility library

lm-n:
* added accessible elements to iframe using accessibility library
* removed interceptor
* added aria labels
* added css to iframe
* focus on iframe when play is pressed

@MathuraMG @CleezyITP


catarak:
yay this is very exciting! in `README.md`, in the installation instructions, there's a section about the submodule setup—could you remove those parts in this PR as well? thanks!

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -198,49 +198,31 @@ class PreviewFrame extends React.Component {
...
-      if (this.props.soundOutput) {
-        let soundInterceptorScripts = [];
-        soundInterceptorScripts = [
-          '/p5-interceptor/soundInterceptor/interceptorP5.js'
-        ];
-        interceptorScripts = interceptorScripts.concat(soundInterceptorScripts);
-      }
-      scriptsToInject = scriptsToInject.concat(interceptorScripts);
+    const accessiblelib = sketchDoc.createElement('script');
+    accessiblelib.setAttribute('src', 'https://cdn.rawgit.com/MathuraMG/p5-accessibility/e856365c/dist/p5-accessibility.js');
```
catarak: rather than link to the accessibility using raw git, i think it would be better to add the library as a dependency in `package.json`, then import the built script using webpack.

client/modules/IDE/reducers/files.js
```diff
@@ -13,15 +13,19 @@ const defaultHTML =
 `<!DOCTYPE html>
 <html>
   <head>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
```
catarak: why is this including jquery?

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -419,7 +401,7 @@ PreviewFrame.propTypes = {
   textOutput: PropTypes.bool.isRequired,
   gridOutput: PropTypes.bool.isRequired,
   soundOutput: PropTypes.bool.isRequired,
-  setTextOutput: PropTypes.func.isRequired,
+  // setTextOutput: PropTypes.func.isRequired,
```
catarak: should this be removed rather than commented out?

MathuraMG:
@catarak - We will fix the above asap. 
Also, a quick update. There is an issue with the accessibility library that we found yesterday. We will fix that and then get back on this PR, so that we have the right version of the library put in. Is that ok?
Thanks!

@lm-n

catarak:
no worries, that is of course okay! this brings up an important point, which is how we want to connect the version of the accessibility library to the editor in the `package.json`, so that when you run `npm install` in the editor, it doesn't install the latest commit, but a stable commit— we could 1. add a commit sha, or 2. tag a release version, using [semantic versioning](https://semver.org/). (1) could come first and then (2) could come later.

client/modules/IDE/reducers/files.js
MathuraMG: we will clear this out - our library has dependencies on jquery but we will make them vanilla javascript.

client/modules/IDE/components/PreviewFrame.jsx
MathuraMG: Will do.

client/modules/IDE/components/PreviewFrame.jsx
MathuraMG: cc @lm-n, @cleezyITP - We have been talking about this one, we would like the library to be a part of the HTML code as a CDN so that when a user shares their code (say, send a zip to someone, it will include the library. This would make sharing more accessible. What do you think?

MathuraMG: @catarak - was wondering what your thoughts on this are?

client/modules/IDE/reducers/files.js
```diff
@@ -18,10 +18,13 @@ const defaultHTML =
     <script src="https://cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.11/addons/p5.sound.min.js"></script>
     <link rel="stylesheet" type="text/css" href="style.css">
     <meta charset="utf-8" />
+
   </head>
   <body>
+    <section aria-label='accessible output' id='accessible-outputs'></section>
```
catarak: rather than add the accessibility elements and css to the default html file, i think it would be better to inject them in the `PreviewFrame.jsx` file, in the function `injectLocalFiles`. As it is now, a user could accidentally delete these elements from the HTML and then the accessibility features would be broken.

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -198,49 +198,31 @@ class PreviewFrame extends React.Component {
...
-      if (this.props.soundOutput) {
-        let soundInterceptorScripts = [];
-        soundInterceptorScripts = [
-          '/p5-interceptor/soundInterceptor/interceptorP5.js'
-        ];
-        interceptorScripts = interceptorScripts.concat(soundInterceptorScripts);
-      }
-      scriptsToInject = scriptsToInject.concat(interceptorScripts);
+    const accessiblelib = sketchDoc.createElement('script');
+    accessiblelib.setAttribute('src', 'https://cdn.rawgit.com/MathuraMG/p5-accessibility/9cb5bd0b/dist/p5-accessibility.js');
```
catarak: i still would prefer to link this file by putting it in the package.json and importing the build with webpack!

client/modules/IDE/reducers/files.js
MathuraMG: @catarak - I think some of the questions we had for you were deleted after we made the new commits :) 
@lm-n and myself were thinking it should be a part of the HTML (at the risk of being deleted) , the same way as p5-dom or p5-sound. This way if someone decides to share their code, they can share it with the accessibility lib and so the code stays accessible beyond the editor. What do you think about this?

catarak: got it. should it be added by default or only if accessibility features are turned on? and also, you want the script to point to a specific release of the library? or a specific commit? or to the latest on master?

MathuraMG: I think we tried to add it only if accessibility features are added but there was a problem? @lm-n ?

lm-n:
@catarak both should work. When I move into the accessibility branch the submodule directory does not exist anymore. It can be done also with `git submodule deinit -f [submodule name]` `rm -rf [submodule directory]` `git rm -f [submodule name]`

client/modules/IDE/reducers/files.js
```diff
@@ -30,6 +31,15 @@ const defaultCSS =
   margin: 0;
   padding: 0;
 }
+
+#accessible-outputs {
```
catarak: i think this is the last thing, but this CSS should also get injected by the library!

client/modules/IDE/components/PreviewFrame.jsx
catarak: @MathuraMG and i talked through this and decided it would be best to address this in the future ✨

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -207,6 +207,12 @@ class PreviewFrame extends React.Component {
     accessiblelib.setAttribute('src', 'https://cdn.rawgit.com/MathuraMG/p5-accessibility/9cb5bd0b/dist/p5-accessibility.js');
     const accessibleOutputs = sketchDoc.createElement('section');
```
catarak: should these accessible outputs be created in the library itself? this is necessary for the library to work, right?

MathuraMG: The reason we don't create the sections in the library is because the section is [currently the trigger](https://github.com/mathuramg/p5-accessibility#usage) to which output the user wants. So in this case we create that section when they select the option in preferences.

@lm-n

catarak: got it. would it instead make sense to make some boolean for the library instead that enables/disables the accessible output? then, in the future, if this ever needs to change you can change it in the accessibility library rather than here as well.

MathuraMG: Sorry, I didn't get what you mean by ' boolean for the library'? :)
The reason we are having sections now is so that beyond the editor the user can determine where they want the output in their program.

catarak: got it! what i mean is maybe there's a script that initializes the library, rather than just appending to an HTML element if it's there. for example, one includes the accessibility library, and then includes a script:
```JS
var accessibility = new p5.Accessibility(<reference to the HTML element you want to append to>);
```
or it could create an HTML element to append to by default. My inspiration for this is [dat.gui](https://github.com/dataarts/dat.gui). my instinct is that it is not the best that including an html element determines whether or not the library is included.

MathuraMG: Got it!  Yes, we will look more into this. 👍 
Would be okay if we track these refactors in a different issue that we tackle following this?

catarak: sure! should i merge this in then?

lm-n: sounds good!

lm-n: Here are the issues on the library repo: https://github.com/MathuraMG/p5-accessibility/issues/13 https://github.com/MathuraMG/p5-accessibility/issues/12

MathuraMG: @catarak Yes! That sounds good!

catarak: great, will do ⭐️

catarak:
just another thought that there's an open issue for, #137: right now, accessibility features won't be added to the full screen view or embed view. and if a user downloads a sketch it won't be included. open to thoughts!

PR #530, opened 2018-02-08
Fixes #517

### LINKED ISSUE #517: save and download?
At the moment, if you make a change to your code and then click "download" what downloads is the latest save and not the latest code. I think that download should probably trigger an "auto-save" so that it's really "save and download". . .thoughts?

Also, it might be useful to name the zip file with a timestamp, i.e. `project_yr_mm_dd_time` if a project is downloaded multiple times it allows for some baby steps into version control?

shinytang6:

Thank you! Fixes #517

### DIFF
--- a/server/controllers/project.controller.js
+++ b/server/controllers/project.controller.js
@@ -2,6 +2,7 @@ import archiver from 'archiver';
 import request from 'request';
 import moment from 'moment';
 import isUrl from 'is-url';
+import slugify from 'slugify';
 import jsdom, { serializeDocument } from 'jsdom';
 import Project from '../models/project';
 import User from '../models/user';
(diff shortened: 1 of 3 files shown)


shinytang6:
@catarak  please review:)

server/controllers/project.controller.js
```diff
@@ -261,7 +261,39 @@ function buildZip(project, req, res) {
     res.status(500).send({ error: err.message });
   });
 
-  res.attachment(`${project.name}.zip`);
+  function currentTime() {
```
catarak: included in this project is a library called [moment](https://momentjs.com/) which makes it easy to convert times to a formatted string. i would prefer if you used this library as it converts this function to one or two lines of code.

shinytang6:
Done. l didn't notice that library:)

catarak:
even though this idea wasn't in the issue, i think it would also make sense as part of this PR to slugify (an npm module which is already included in the project now) the sketch name, since it's annoying to have spaces in file names, see https://github.com/processing/p5.js-web-editor/blob/master/server/models/project.js#L43 for an example

shinytang6:
thanks for reminding this!

catarak:
this is working great!

PR #538, opened 2018-02-10
Fixes #520 

### LINKED ISSUE #520: When navigating to `/<username>/sketches/<random_string>`, does not return 404 if the sketch does not exist
#### Nature of issue?

- [X] Found a bug

#### Details about the bug: 

- Web browser and version: 
- Operating System: MacOSX
- Steps to reproduce this:

Go to https://alpha.editor.p5js.org/cassie/sketches/random_string_that_does_not_match_an_id. The editor does not return a 404, but instead just serves the default sketch, one that would be returned if you had created a new sketch. The expected behavior is that the user would be directed to a 404 page, as it is unclear that the user has done anything wrong right now.

shinytang6:

Thank you! Fixes #520

### DIFF
--- a/server/routes/server.routes.js
+++ b/server/routes/server.routes.js
@@ -2,7 +2,7 @@ import { Router } from 'express';
 import { renderIndex } from '../views/index';
 import { get404Sketch } from '../views/404Page';
 import { userExists } from '../controllers/user.controller';
-import { projectExists } from '../controllers/project.controller';
+import { projectExists, projectForUserExists } from '../controllers/project.controller';
 
 const router = new Router();
 
@@ -27,10 +27,8 @@ router.get('/projects/:project_id', (req, res) => {
 });
 
 router.get('/:username/sketches/:project_id', (req, res) => {
-  userExists(req.params.username, exists => (
-    exists ? projectExists(req.params.project_id, exists => ( // eslint-disable-line
-      exists ? res.send(renderIndex()) : get404Sketch(html => res.send(html))
-    )) : get404Sketch(html => res.send(html))
+  projectForUserExists(req.params.username, req.params.project_id, exists => (
+    exists ? res.send(renderIndex()) : get404Sketch(html => res.send(html))
   ));
 });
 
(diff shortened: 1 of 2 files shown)


shinytang6:
@catarak Can not think of a better way. please review:)

catarak:
since all of the links on the front end are generated by the front end, i think that the 404 page should be returned from the server (i.e. in `server.routes.js`). also, this is maybe not totally clear, but you changed one of the API routes endpoints, which don't need to change! the API urls should return a 404 status code (but still json) and the website endpoints should redirect to the 404 page. sorry this was not clear in the issue!

shinytang6:
get√ but l think maybe this api still needs the param `username`? It only checks `projectId` whether who owns it

catarak:
the `projectId`s are unique for each sketch so the `username` shouldn't be necessary! though maybe i'm confused about what you mean.

shinytang6:
emmm what l think is though the projectId is unique, the source doesn’t check whether it belongs a existing user

catarak:
i see what you're saying! there are a few different routes here:
 
from `server.routes.js`:
`/projects/:project_id`
`/:username/sketches/:project_id`
from `project.routes.js`:
`/api/projects/:project_id`

in this PR you changed the last route to `/api/:username/projects/:project_id`, which doesn't need to be changed. as of right now, since the API is not public, front end users are never going to api routes, and they are generated by the front end. it is important that the first two routes redirect to 404 pages if the `:project_id` or `:username` does not exist. does that make sense?

shinytang6:
cool :)

server/routes/server.routes.js
```diff
@@ -20,11 +21,17 @@ router.get('/signup', (req, res) => {
...
 router.get('/projects/:project_id', (req, res) => {
-  res.send(renderIndex());
+  projectExists(req.params.project_id, exists => (
+    exists ? res.send(renderIndex()) : get404Sketch(html => res.send(html))
+  ));
 });
 
 router.get('/:username/sketches/:project_id', (req, res) => {
-  res.send(renderIndex());
+  userExists(req.params.username, exists => (
```
catarak: can you also check if the sketch belongs to the user here?

shinytang6: done :) Tell me if there are any other mistakes!

PR #539, opened 2018-02-12
Fixes #508

lm-n:
Fixes issues in #508


lm-n:
@MathuraMG

MathuraMG:
@lm-n looks good! Can we change the accessibility lib link in the README as well?

Issue #437, opened 2017-09-13
unexpected revert to previous version of file


lbarron1:
All the code written for a project reverted to an earlier version unexpectedly. The project was saved regularly and no other issues seemed present. The file that the piece went back to was the one saved prior to closing the browser and reopening it to work on it later.

catarak:
I've noticed something similar in #419. I will investigate, and if you notice any more specific steps to reproduce add them here!

ajuusola:
I'm also having this issue. 
I've been attempting to work in p5 throughout the week, but I keep running into issues where my sketch will revert back to a previous duplicate I have made of it several minutes (sometimes even days) ago. No matter how many times I save, restart my computer/browser, or duplicate my documents, it still decides to jump back to a previous version of the sketch I've been using and erase all of the progress I've made. I ran into this issue last week as well. I tried completely re-imaging my computer (for additional reasons) and using a different browser, but it didn't work.

dhruvdutt:
Couldn't reproduce it. @catarak were you able to reproduce it by any chance?

catarak:
i'm thinking that this has to do with having multiple versions of the same sketch open.

catarak:
just pushed a fix (#1039) that hopefully fixes this!

Issue #419, opened 2017-08-24
Running one program actually runs a different program.


BenLuker:
I was working on two separate programs for my class. I opened program A and ran it. Then I opened program B and ran it. They were identical. I thought I had duplicated program A as program B, because they were identical, so I deleted program A. Then I ran program B again, and it was completely different! I just lost program A because of this.

catarak:
Oh no, I'm sorry to hear this! I will try to recreate, but can you also add a link to the sketch? Maybe it's a a bug that has to do with the specific sketch.

catarak:
My current steps to reproduce this bug are:

1. Open existing sketch
2. Don't save it
3. hit duplicate
4. Make changes to the duplicate sketch
5. When you save the duplicated sketch, your changes disappear

BenLuker:
You know what? I think I was just stupid and did something wrong. Thanks
for your concern though.

catarak:
No worries! I think that this helped me find a separate bug so 🎉 🎉 🎉

kjhollen:
Oohhhhh I think I saw this over the summer but didn't report it because I couldn't figure out the steps to reproduce it. Glad you could track this down, Cassie, and thanks for filing, Ben!

himanshuc3:
@catarak Repeating the steps you did to reproduce the issue, the page got reloaded but the changes did not get lost when I repeated the process.

catarak:
@shinytang6 does #546 fix this?

dhruvdutt:
@catarak Can this be closed?

PR #540, opened 2018-02-12
Fixes file saving problem

### LINKED ISSUE #470: Changes aren't saving
#### Nature of issue?

- [ X] Found a bug
- [ ] New feature request

#### Details about the bug: 

- Web browser and version: Version 61.0.3163.100 (Official Build) (64-bit)
- Operating System: Mac OS 10.12.6 (16G29)
- Steps to reproduce this:
1. make a change in code
2. use keyboard shortcut cmd-s to save ("project saved"dialog appears)
3. reload page and change has disappeared.

#### New feature details:

shinytang6:

Thank you! Fixes #470 #419 #437 ;

### DIFF
--- a/client/components/Nav.jsx
+++ b/client/components/Nav.jsx
@@ -167,6 +167,7 @@ class Nav extends React.PureComponent {
               <li className="nav__dropdown-item">
                 <button
                   onClick={() => {
+                    this.props.autosaveProject();
                     this.props.exportProjectAsZip(this.props.project.id);
                     this.setDropdown('none');
                   }}
@@ -487,6 +488,7 @@ class Nav extends React.PureComponent {
 Nav.propTypes = {
   newProject: PropTypes.func.isRequired,
   saveProject: PropTypes.func.isRequired,
+  autosaveProject: PropTypes.func.isRequired,
   exportProjectAsZip: PropTypes.func.isRequired,
   cloneProject: PropTypes.func.isRequired,
   user: PropTypes.shape({
--- a/client/modules/IDE/actions/project.js
+++ b/client/modules/IDE/actions/project.js
@@ -2,6 +2,7 @@ import { browserHistory } from 'react-router';
 import axios from 'axios';
 import objectID from 'bson-objectid';
 import each from 'async/each';
+import compare from 'just-compare';
 import * as ActionTypes from '../../../constants';
 import { showToast, setToastText } from './toast';
 import { setUnsavedChanges,
@@ -72,7 +73,7 @@ export function clearPersistedState() {
 
 export function saveProject(autosave = false) {
   return (dispatch, getState) => {
-    const state = getState();
+    let state = getState();
     if (state.user.id && state.project.owner && state.project.owner.id !== state.user.id) {
       return Promise.reject();
     }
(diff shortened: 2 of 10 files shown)


shinytang6:
@catarak l think l find where the problem lies

The steps to reproduce that:

1. open two identical projects on browser
2. make changes to one projects and **quickly** save it using **ctrl+s** (but you can't do that **too quickly**)
3. then you will see the project you changed seem to be saved(with the unsavedchange dot disappered, if exists, it means you press too quickly), but actually it is not saved.

The problems occurs because when you press ctrl+s, it triggers the `saveProject` :
[https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/actions/project.js#L73](https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/actions/project.js#L73)

But at that time the changes are not detected because of debounce, so what uploads to the server is the unchanged file, see : 
[https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/components/Editor.jsx#L114](https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/components/Editor.jsx#L114)

the difference between `quickly` and `too quickly` is whether 
[https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/actions/project.js#L84](https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/actions/project.js#L84)
happens before [https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/components/Editor.jsx#L115](https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/components/Editor.jsx#L115)

Hope l explain clearly :)

catarak:
i'm not sure the right solution is to remove the debounce. if a user has 'auto-refresh' enabled, then the debounce keeps the sketch from updating every time a character changes and therefore crashing all of the time. maybe a better solution would be when a user hits 'play', the file updates before generating the preview?

shinytang6:
>maybe a better solution would be when a user hits 'play', the file updates before generating the preview

@catarak   lt may still cause problem in some cases...
e.g. If after this situation happens, a user still makes a lot of changes and save many times, hiting `play` will lose all the changes as you say..

Maybe we can call `state.files` again and compare it with the `response.data.files`  
before [https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/actions/project.js#L86](https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/actions/project.js#L86)

If error happens, they are different.... what do you think?

catarak:
that's a good idea! maybe `setProject` shouldn't even be called there. the source of truth here should be what's in the CodeMirror editor unless a user has just opened a sketch.

shinytang6:
That’s true, l will work on it soon！btw, can l ask are there some other ways you know/use to test ssl？

catarak:
the directions in the readme are all i know—in production i'm using let's encrypt for the ssl certs, but i recently learned that ssl is also free if you use aws cloudfront.

shinytang6:
l work on a new laptop today:) it seems that l made something wrong with git...Sorry for that o(╥﹏╥)o, close this and refer to #546 :)

catarak:
all good!

PR #548, opened 2018-02-18
Fixes #508

lm-n:
fixes #508 
moved css for accessible output from files.js to PreviewFrame.jsx


MathuraMG:
@lm-n - yup, let's add this to be tracked on the library side

PR #563, opened 2018-02-21
fixes #384

### LINKED ISSUE #384: Update Toast styling to new design
screen shot 2017-07-11 at 10 48 10 am

More details on the [Zeplin project](https://zpl.io/Z2cihf1).

himanshuc3:
- [x] Correct Font size
- [x] Correct positioning to bottom

### DIFF
--- a/client/styles/components/_toast.scss
+++ b/client/styles/components/_toast.scss
@@ -3,12 +3,15 @@
   color: $toast-text-color;
   padding: #{10 / $base-font-size}rem;
   position: absolute;
-  bottom: 5vh;
-  left: 5vh;
+  bottom: 10%;
+  left: 5%;
+  min-width: #{375 / $base-font-size}rem; //500px
   font-size: #{20 / $base-font-size}rem;
   line-height: #{24 / $base-font-size}rem;
   display: flex;
   z-index: 9999;
+  align-items: center;
+  justify-content: space-between;
   box-shadow: 0 6px 6px 0 rgba(0,0,0,0.10);
 }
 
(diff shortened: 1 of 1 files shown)


client/styles/components/_toast.scss
```diff
@@ -3,8 +3,10 @@
   color: $toast-text-color;
   padding: #{10 / $base-font-size}rem;
   position: absolute;
-  top: 0;
-  right: 40%;
+  bottom: 5vh;
```
catarak: could you use consistent measurements here? for the site i've been using `rem`s, which are easy to convert with SCSS:
`#{<pixels> / $base-font-size}rem;`

catarak:
a few things:

1. I think the toast needs to be a little higher on the page (and a little more to the right). Could you use percentage instead of `vh`?
2. It needs a minimum width (in the mockups it's 500px).
3. Instead of centering the text with `line-height`, could you use flexbox? IMO this is more semantically correct. 

Thank you!

client/styles/components/_toast.scss
```diff
@@ -3,16 +3,22 @@
   color: $toast-text-color;
   padding: #{10 / $base-font-size}rem;
   position: absolute;
-  top: 0;
-  right: 40%;
+  bottom: 10%;
+  left: 5%;
+  min-width: #{375 / $base-font-size}rem; //500px
+  font-size: #{20 / $base-font-size}rem;
+  line-height: #{24 / $base-font-size}rem;
```
catarak: you don't need line height here 😄

PR #565, opened 2018-02-22
Adding basic meta tags.

himanshuc3:
- [x] MIME type
- [x] keywords
- [x] Description

### DIFF
--- a/client/modules/IDE/components/Editor.jsx
+++ b/client/modules/IDE/components/Editor.jsx
@@ -181,7 +181,7 @@ class Editor extends React.Component {
     if (prevProps.consoleEvents !== this.props.consoleEvents) {
       this.props.showRuntimeErrorWarning();
     }
-    for (let i = 0; i < 1000; i += 1) {
+    for (let i = 0; i < this._cm.lineCount(); i += 1) {
       this._cm.removeLineClass(i, 'background', 'line-runtime-error');
     }
     if (this.props.runtimeErrorWarningVisible) {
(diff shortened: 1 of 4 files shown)


catarak:
awesome! in the future, could you open an issue first so that the pull request can reference an issue? also this brings up a good question as to what we want these meta tags to be. thoughts @lmccart & @shiffman?

shiffman:
This is great! I might suggest avoiding the word "playground", I like the spirit of that very much, but it might be confusing as there is @yining1023's project which is actually called p5 playground! http://1023.io/p5-inspector/

I think for now: "p5.js", "p5.js web editor", "web editor" are good?

lmccart:
I would also add "processing", "code editor"

catarak:
could you change the description from "A web editor for sketching using p5.js library." to "A web editor for p5.js, a JavaScript library with the goal of making coding accessible to artists, designers, educators, and beginners."

himanshuc3:
:zap:

catarak:
could you also add these meta tags to `server/views/404page.js`? also `server/views/mailLayout.js`? emails also have meta tags, right?

himanshuc3:
Do you want me add the same or have a discussion about what their description should contain.

catarak:
i think they should be the same, thanks!

himanshuc3:
@catarak I added meta tags to 404error page. MailLayout page is made using mjml library and I could not find any existing meta tag support, though basic tags such as IE-edge support, content type are added by default. Adding meta tags to mailLayout could be made using another PR as a custom component could be made for it.

catarak:
okay! i can open an issue about it. maybe emails don't need meta tags, but we can investigate it!

PR #574, opened 2018-02-26
fixes #573

### LINKED ISSUE #573: Documentation Updates
- From the beginner perspective, it could help them start faster.
- It helps organize code, find things faster and therefore makes finding bugs easier.

Starting from the root. This Issue starts by adding some bits of comments in index.html page. More issues will follow regarding documentation from my side. It would be providing me with more depth understanding of code.

himanshuc3:


catarak:
honestly i don't think this file is even used—it was at first but i moved the index.html to a file called `server/views/index.js`.

himanshuc3:
Well, then a beginner can be directed to index.js view through comments because there is nowhere mentioned that the entry point is index.js file.

catarak:
rather than just add a comment that this file is no longer used, we should just delete it, right?

catarak:
also feel free to add that to the README or CONTRIBUTING—would love to improve the documentation!

himanshuc3:
Well, there are 2 reasons I've added comments to index.html file.

1.  Since there is no documentation right now in CONTRIBUTING i've added it till that is formed.
2. I thought maybe for some reason if we were ever to rebase the starting point to index.html from index.js.
If there's no question of rebasing to index.html, I shall then delete this and add the documentation to CONTRIBUTING file.

catarak:
i think it would be better to just add to CONTRIBUTING or README! when adding to either of these you don't need to add the perfect documentation. i think it's better to slowly add all of these ideas as they come up and eventually we will have some awesome documentation ✨

PR #577, opened 2018-02-27
Issue overlay

himanshuc3:
Solves:
- [x] Closing modal when clicking outside it
- [x] Closing modal when pressing escape

Do make sure that the code added does not create more bugs. I personally did not found any but i'm a little skeptical.

### DIFF
--- a/client/modules/App/components/Overlay.jsx
+++ b/client/modules/App/components/Overlay.jsx
@@ -39,6 +39,8 @@ class Overlay extends React.Component {
   }
 
   keyPressHandle(e) {
+    // escape key code = 27.
+    // So here we are checking if the key pressed was Escape key.
     if (e.keyCode === 27) {
       this.close();
     }
@@ -59,7 +61,7 @@ class Overlay extends React.Component {
       children
     } = this.props;
     return (
-      <a tabIndex={-42} onKeyDown={this.keyPressHandle} id="overlay__anchor">
+      <a tabIndex={-1} onKeyDown={this.keyPressHandle} className="overlay__anchor">
         <div className="overlay">
           <div className="overlay__content">
             <section
@@ -80,25 +82,6 @@ class Overlay extends React.Component {
           </div>
         </div>
       </a>
-    // <div className="overlay">
-    // <div className="overlay__content">
-    //   <section
-    //     tabIndex="0"
-    //     role="main"
-    //     aria-label={ariaLabel}
-    //     ref={(node) => { this.node = node; }}
-    //     className="overlay__body"
-    //   >
-    //     <header className="overlay__header">
-    //       <h2 className="overlay__title">{title}</h2>
-    //       <button className="overlay__close-button" onClick={this.close} >
-    //         <InlineSVG src={exitUrl} alt="close overlay" />
-    //       </button>
-    //     </header>
-    //     {children}
-    //   </section>
-    // </div>
-    // </div>
     );
   }
 }
--- a/client/modules/App/components/Overlay.jsx
+++ b/client/modules/App/components/Overlay.jsx
@@ -16,6 +16,7 @@ class Overlay extends React.Component {
 
   componentWillMount() {
     document.addEventListener('mousedown', this.handleClick, false);
+    document.addEventListener('keydown', this.keyPressHandle);
   }
 
   componentDidMount() {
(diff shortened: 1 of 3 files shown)


client/modules/App/components/Overlay.jsx
```diff
@@ -30,25 +59,46 @@ class Overlay extends React.Component {
...
+                  <InlineSVG src={exitUrl} alt="close overlay" />
+                </button>
+              </header>
+              {children}
+            </section>
+          </div>
         </div>
-      </div>
+      </a>
+    // <div className="overlay">
```
catarak: rather than just commenting out this code, could you remove it? if we want to look at the old version we can just revert this file to a previous commit ✨

client/modules/App/components/Overlay.jsx
```diff
@@ -9,10 +9,39 @@ class Overlay extends React.Component {
...
+
+    this.handleClickOutside();
+  }
+
+  handleClickOutside() {
+    this.close();
+  }
+
+  keyPressHandle(e) {
+    if (e.keyCode === 27) {
```
catarak: could you add a comment here that 27 == escape key?

client/modules/App/components/Overlay.jsx
```diff
@@ -30,25 +59,46 @@ class Overlay extends React.Component {
...
-          >
-            <header className="overlay__header">
-              <h2 className="overlay__title">{title}</h2>
-              <button className="overlay__close-button" onClick={this.close}>
-                <InlineSVG src={exitUrl} alt="close overlay" />
-              </button>
-            </header>
-            {children}
-          </section>
+      <a tabIndex={-42} onKeyDown={this.keyPressHandle} id="overlay__anchor">
```
catarak: i think there's no difference between setting the tab index to -1 and to any other negative integer, right?

client/styles/components/_overlay.scss
```diff
@@ -40,4 +40,8 @@
 .overlay__close-button {
   @include icon();
   padding: #{3 / $base-font-size}rem 0;
+}
+
+#overlay__anchor{
```
catarak: for this project i would like to apply styles only by classes (and sometimes tags for base styles), no id's.

client/modules/App/components/Overlay.jsx
catarak: also is this valid/semantic html? can you put a div inside of an a tag?

catarak:
can you also rebase this PR with the HEAD of master?

client/modules/App/components/Overlay.jsx
himanshuc3: ok, forgot about it :sweat_smile:

client/modules/App/components/Overlay.jsx
himanshuc3: sure

client/modules/App/components/Overlay.jsx
himanshuc3: Since i am giving the anchor a keypress property i guess it does have a weak semantic meaning. It's the only hack I know to put up with Handling events since divs can't have event handling in react.

client/styles/components/_overlay.scss
himanshuc3: ok,i'll change that

himanshuc3:
What time are you available for chatting. I mean development could go faster, cause right now,
I make a commit, the day after you reply in IST and then I make changes and the next day you merge, which takes minimum of 3 days to merge even the simplest of PRs. Please mention the time standard too.

catarak:
I'm in EST, so really anytime during regular working hours (10AM to 6PM). It's also totally cool that issues take a few days to merge in. There's no rush!

himanshuc3:
cool. I just want things to get done. I just feel there's just too much in life to get done. So get going is my scheme of things. Anyway, any other changes in this PR.

catarak:
this is working but i don't understand why the addition of the `<a>` tag is necessary. it's also not valid HTML (see https://stackoverflow.com/a/1828032/4086967, the overlays frequently have interactive content) so it needs to be removed.

himanshuc3:
I could remove anchor tag but that would also remove the escape key functionality. I tried giving the div a role to make it compatible it to be treated as a button but it was still showing an error.

catarak:
hmmmm there's a function in `IDEView.jsx` that captures key events... maybe that would be helpful to use?

catarak:
or perhaps some inspiration from other overlay examples/libraries?

himanshuc3:
I think, this one's fixed the overlay issue.

catarak:
works great!

Issue #468, opened 2017-11-04
Missing Console Errors


shiffman:
Starting over from #444, this github issue is to track error messages that show up in the native JS console, but not in the p5.js web editor console.  To be clear, there are likely some error messages that should be hidden in the web editor console so this thread is also for discussion around these questions. This first issue contains a list of all the errors and subsequent comments will include specific code examples to trigger the missing error. Anyone should feel free to jump in and add to the comment thread and I'll keep the list below up to date!

* [ ] undefined objects
* [ ] file not found errors with `loadJSON()`, `loadImage()`, etc.

shiffman:
Here's an example related to using an `undefined` string with `text()`:

```javascript
let txt;

function setup() { 
  createCanvas(400, 400);
  text(txt,100,100);
} 
```

JS console error: `Cannot read property 'toString' of undefined`

shiffman:
Here's an example related to "file not found":

```javascript
let img;

function preload() {
  img = loadImage('missingfile.png');	
}

function setup() {}
```

JS console error: `GET http://alpha.editor.p5js.org//missingfile.png 404 (Not Found)`

shiffman:
Incorrect renderer:

```javascript
function setup() { 
  createCanvas(400, 400);
} 

function draw() { 
  background(220);
  rotateZ(100);
}
```

Error should read: `Uncaught not supported in p2d. Please use webgl mode`

catarak:
this has mostly been fixed, except for the file not found error. not sure why that's not getting to the console!

GaurangTandon:
Is #868 a duplicate of this issue?

catarak:
yep! they are the same issue.

shakeabi:
@catarak I was looking through how errors are captured and thought [this](https://stackoverflow.com/a/28771916) might help.

catarak:
@shakeabi maybe! we're using the `console-feed` library, so i'm not sure if that error handler needs to be added in addition to `console-feed`.

w3cj:
Hey! I'm going to take a stab at the file not found issue. This is what I see happening in the editor:

![](https://i.imgur.com/eKUWKj5.png)

Just want to clarify, would the fix be, instead of showing the `Event` object log, show the actual error message `Failed to load resource: the server responded with a status of 404 ()`?

w3cj:
[...]

PR #579, opened 2018-02-28
[WIP] display the script error caused by external library

shinytang6:

Thank you!

### DIFF
--- a/client/modules/IDE/components/PreviewFrame.jsx
+++ b/client/modules/IDE/components/PreviewFrame.jsx
@@ -36,8 +36,8 @@ function getAllScriptOffsets(htmlFile) {
     } else {
       endFilenameInd = htmlFile.indexOf('.js', ind + startTag.length + 3);
       filename = htmlFile.substring(ind + startTag.length, endFilenameInd);
-      // the length of hijackConsoleErrorsScript is 31 lines
-      lineOffset = htmlFile.substring(0, ind).split('\n').length + 31;
+      // the length of hijackConsoleErrorsScript is 33 lines
+      lineOffset = htmlFile.substring(0, ind).split('\n').length + 33;
       offs.push([lineOffset, filename]);
       lastInd = ind + 1;
     }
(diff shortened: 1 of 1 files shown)


shinytang6:
partly relates to #468 :)
fixes e.g. 
```
let txt;

function setup() { 
  createCanvas(400, 400);
  text(txt,100,100);
} 
```

```
function setup() { 
  createCanvas(400, 400);
} 

function draw() { 
  background(220);
  rotateZ(100);
}

```

Still need to know more about how to display lineNumber when the error caused by external library.
Currently it only shows the error line in e.g p5.min.js :)

catarak:
this is interesting and i'm not sure how to handle it. the issue is that the user is using the library improperly. i think the ideal case is that we would get the full stack trace of the error, figure out the line that is in one of the files in sketch itself, and put the error on that line, and show the full stack trace in the console. can you get the whole stack trace? there's a script called `hijackConsole.js` that sends the console messages through a message channel to the web editor console.

shinytang6:
The first one seems to be working properly now:)
```
function setup() { 
  createCanvas(400, 400);
} 

function draw() { 
  background(220);
  rotateZ(100);
}
```
this one doesn't show whole stack trace, only reports `Uncaught not supported in p2d. Please use webgl mode`, l think it is relevant to the p5.js's console?

catarak:
cool, i like where this is going! i wonder if it would make sense to send the whole stack back to the in-window console? is it too confusing for a coding newbie? 
screen shot 2018-03-01 at 1 43 31 pm

vs.

screen shot 2018-03-01 at 1 43 07 pm

is there potentially a happy medium? as someone with coding experience i found the error confusing because i was like, "toString isn't being called on this line," and i think it's important to say that the error is being thrown in another file, but is a result of the user not using the library correctly, and therefore there still is in error. thoughts @lmccart & @shiffman?

i also just changed the p5 version to the non-minified version to see if the friendly errors system would catch it, and it doesn't.

shiffman:
I think the ideal outcome would be for the error to read something like:

```
p5 cannot render undefined text. sketch: line 5
```

So maybe this is a matter of adding this particular scenario to the friendly errors system?

If rewriting the error is not an option I think I would stick with the single error over the full stack trace for the web editor? We could reconsider this based on any future work on the console itself #7 -- for example, maybe the full stack trace could "unfold"?  But my instinct says that the full stack trace is more scary than helpful, and it would be available in the chrome developer tools should someone want to look.

catarak:
makes sense to me!

shinytang6:
@catarak @shiffman so what is the verdict on this one currenly:)

shiffman:
For now let's go with "Uncaught TypeError: Cannot read property 'toString' of undefined (sketch: Line 5)". I can add an issue regarding a friendly error for this separately in the p5.js repo.

shinytang6:
That will be fine :smile:

catarak:
cool, then this PR is good to be merged.

PR #580, opened 2018-02-28
🚧 WIP - add support for p5.js version check

### LINKED ISSUE #144: p5.js version
@catarak @lmccart and I discussed this on a call yesterday.  At some point in the future we'd like to have a library management system (#4), but for now we need a simple way to keep the p5.js version referenced in `index.html` up-to-date.   We would not change `index.html` in existing projects (or "duplicated" projects) but would simply use the current `p5.js` version (0.5.4 at the time of filing this issue) whenever a "new" project is created.  

@lmccart can you remind us what the url is to check the current p5  version?  It could be checked something like once a day or each time "new" is clicked. 

Did I get this right?

dhruvdutt:
This PR is initial WIP to add support p5.js version check as described in #144 

Server: :cloud: 
- Exposes p5.js latest version at new API endpoint: `/api/p5version/`

Client: :computer: 
- Check for latest p5.js version during initialization using above API

TODO 👷
- [ ] Fix re-rendering issue
- [ ] Version checks during "new project" action

Fixes #144 :bug: 

Thank you!


dhruvdutt:
@catarak Can you please look into this issue?

After fetching p5 version from api, I'm setting in into [files reducer](https://github.com/dhruvdutt/p5.js-web-editor/commit/adba986cf5b3c1bd8999a24b100c2d76a3a589a9#diff-a9bf519846ef971fe08e38defb744624R238) and the state gets updated (check screenshot: redux dev tools section).

The issue is that the html file markup inside editor is not getting updated.
Possibly, this could be the issue
https://github.com/processing/p5.js-web-editor/blob/ea02d7692d52a5bbefe4b6dcf1f1aa41cdff872a/client/modules/IDE/components/Editor.jsx#L147-L156

Also, a note that I'm setting default one to 0.5.0 in the [initial state](https://github.com/dhruvdutt/p5.js-web-editor/commit/adba986cf5b3c1bd8999a24b100c2d76a3a589a9#diff-a9bf519846ef971fe08e38defb744624R12) so that we can see the updated one (0.6.0) from the API.

image

catarak:
i think the code you've pieced out is probably the issue—it's somewhat hacky code i wrote to figure out if there are any changes the the files, which is obviously not robust. would be awesome to update it!

catarak:
i also just had an idea—since the p5 version isn't changing that often, maybe it would make sense to store it in mongo or something, and add a script that checks that URL every day or every hour or something? or maybe add a CI task that every time there's a tagged release in the p5 repo the editor gets updated? what do you think?

dhruvdutt:
> i think the code you've pieced out is probably the issue

I think it will take a while to investigate it. We should certainly fix the hack but is this version check a priority? Fixing the hack is a priority for sure.
Did we get stuck in any others issues because of this hack?

> store it in mongo or something

Makes complete sense to do it. I also thought to cache it somewhere as this API call takes up at least 200 ms to complete which directly impacts our startup time/new project creation time. We should definitely deploy a cron job at server or setup CI based on tagged release but things will ultimately boil down to where we're stuck in the component lifecycle.

catarak:
i don't know if there are any other issues, but yeah, it would be great to make the code more robust and clean. 

to your second point, one thing i've been thinking about is stuff that can be done server side. it would be cool if certain data could be bootstrapped (something like this: https://medium.com/@baphemot/understanding-reactjs-data-hydration-initialization-bacbb790c7cb)

catarak:
closeed this PR since it seems to be stale. feel free to reopen if you're still working on it!

PR #586, opened 2018-03-02
fixes #555

### LINKED ISSUE #555: Move examples username and password to .env file
#### Nature of issue?

- [X] Found a bug

#### Details about the bug: 

- Web browser and version:  N/A
- Operating System:  N/A
- Steps to reproduce this:

the file `server/examples.js` contains the p5 username and password. this should be moved to environment variables so that any user can't login to this account!

himanshuc3:


README.md
```diff
@@ -24,6 +24,9 @@ This project is currently in development! It will be announced when there is a (
   GITHUB_ID=<your-github-client-id>
   GITHUB_SECRET=<your-github-client-secret>
   MAILGUN_KEY=<your-mailgun-api-key>
+  EXAMPLES_USERNAME=<your-name>
```
catarak: these variables need to be added to the production installation .env instructions too!

himanshuc3:
but isn't the `.env` file ignored by git. So it cannot be commited anyway? Should i force it or something?

catarak:
don't check in the `.env` file—on the README there's an example for what the `.env` file should look like in production and development. the idea of the .env file is that it's a place for secret variables that shouldn't be checked in, like API keys and stuff like that.

PR #599, opened 2018-03-11
fixes #441

### LINKED ISSUE #441: Canvas preview in editor doesn't show full range of mouse postions
function draw() {
print(mouseX);  // Never shows 0 when run in editor, lowest value appears to be 6
}

MacOS Chrome Version 61.0.3163.91 (Official Build) (64-bit)

himanshuc3:


catarak:
no sure if the dragging element should have a background color or something, but this works. we can save that for another issue ✨

Issue #569, opened 2018-02-23
Add authorization to specific routes


catarak:
#### Nature of issue?

- [X] Existing feature enhancement

#### Feature enhancement details:
Certain routes should not be accessible by all users! One that comes to mind is a user's settings (`/<username>/settings`) and a user's assets (`/<username/assets`). Maybe these routes should not include the username and just use the user that it stored in the cookie? Or they should redirect to `/` if a user who is not logged in as the `<username>` tries to navigate to one of these routes. The former appears to be what most websites do.

himanshuc3:
If I may ask, is the file structuring done according to mern.io.

catarak:
yes, that is the inspiration!

catarak:
server-side, this project uses a library called `passport` which adds some `node`/`express` middleware magic 🔮and if a user is logged in, the user is stored in `req.user`. so you could compare the `<username>` in the route to `req.user.username`. another option would be to remove the username from the route, and determine the user from the cookie (just check `req.user.id`).

migom6:
@catarak 
what's the progress? should I take this issue?

catarak:
no progress, feel free to take it!

migom6:
@catarak @shiffman 
Does this **_/\<username>\/settings_** exists ? if not should this route be created ?

catarak:
it does exist. i was thinking that this route should be moved to just `/settings` and the assets one should be moved to just `/assets` and then we can avoid this authorization issue all together.

PR #615, opened 2018-04-07
Solving issue #569 authorised routes. 

migom6:
This PR aims to solve issue #569 i.e authorization of routes **_/assets_** and **_/settings_** and using cookie in the client to access the username.

Hi @catarak, can you review this code and tell me if my approach is correct or it needs changes. I have changed the route  ` /<username>/settings` to  `/assets`   only as of now.

**_Thanks_** 😄

### DIFF
--- a/client/components/Nav.jsx
+++ b/client/components/Nav.jsx
@@ -475,7 +475,7 @@ class Nav extends React.PureComponent {
                 </li>
                 <li className="nav__dropdown-item">
                   <Link
-                    to={`/${this.props.user.username}/assets`}
+                    to={'/assets'}
                     onFocus={this.handleFocus.bind(this, 'account')}
                     onBlur={this.handleBlur}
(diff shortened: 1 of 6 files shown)


catarak:
thanks for submitting this! the `username` isn't actually stored in the cookie—it contains an id which maps to a session storage in mongodb. therefore, you can only grab the username using the cookie server side. client-side, when the application loads, it makes an ajax request to `/api/session` which fetches the current user if they are logged it and stores it in the redux store. therefore, if the website is loading `/assets` client-side, you only need to check the username in `req.user.username`, and if the website is loading server-side, you can grab the username from `req.user.username` after the ajax request from `/api/session` returns, and if the user is not logged in, redirect to `/`.

migom6:
@catarak 
Thanks for the elaborate description and my approach was completely off. I will change the code and let you know soon.

migom6:
I'll be sending a new PR. Closing this.

PR #639, opened 2018-05-18
[WIP] Improve current console

shinytang6:

* [ ] there are no linting errors -- `npm run lint`
* [ ] your code is in a uniquely-named feature branch and has been rebased on top of the latest master. If you're asked to make more changes, make sure you rebase onto master then too!

Thank you!
Ref #7


shinytang6:
Still work in progress. Need more style work and tests

shinytang6:
Some screenshots of the improved console.






There are currently two issues that I cannot figure out:

1. The console's color does not render correctly when the theme is switched.
2. The logged times label is now a float element which means that there will be something wrong when the number of times is large, l think it's better to be packaged into the console-feed library, but l cannt do that now :worried: 

@catarak Can you plz take a look? l am not good at designing styles :thinking:

shiffman:
So excited about this, great work!

machinic:
@catarak if I understand the scope of this correctly, we want to harmonize the particular values used in the console, like in the following examples derived from `client/styles/components/_console.scss`:

```
preview-console__log, 
.preview-console__info, 
.preview-console__log__logged-times, .preview-console__info__logged-times 
	background-color: $lightsteelblue;
```
```
.preview-console__error,
.preview-console__error__logged-times
	background-color: $red;
```
```
.preview-console__warn,
preview-console__warn__logged-times
background-color: $orange;
```
```
.preview-console__debug,
.preview-console__debug__logged-times
	background-color: $dodgerblue;
```


with core colors from `client/styles/abstracts/_variables.scss` e.g.

```
$p5js-pink: #ed225d;
$white: #fff;
$black: #000;
$yellow: #F5DC23;
$orange: #ffa500;
$red: #ff0000;
$lightsteelblue: #B0C4DE;
$dodgerblue: #1E90FF;
$primary-text-color: #333;
$icon-color: #8b8b8b;
$icon-hover-color: #333;
```

I'd be happy to take a crack at this from a design perspective and post an update in a few days or so if that timeframe works. I'd be interested to know if there's any opposition to using Sass functions to darken and/or lighten the core colors for one or both of the themes? I think that approach might produce more pleasing results within a theme for the console which effectively has a different background color than the editor overall. 

I'm also thinking some check into the syntax highlighting colors for appropriate parallels might be worth doing. 

I don't see `$console-warn-color: #ffbe05;` or `$console-error-color: #ff5f52;` in play here and would be interested to know if those are essentially relics of a previous approach.

shinytang6:
@machinic l removed the `$console-warn-console` and `$console-error-color` to distinguish the color of times logged and the log content. If this method is accepted, these few variables should be deleted :)

machinic:
@shinytang6 — perfect, thanks for the info!

catarak:
@shinytang6 the theming is done using a SCSS mixin called `themify` and a SCSS function called `getThemifyVariable()`. if you look in the file `_variables.scss`, you'll notice a SCSS map called `$themes`.  you'll notice that in files like `_console.scss`, you'll see
```scss
.preview-console__log {
	@include themify() {
		color: getThemifyVariable('secondary-text-color');
	}
}
```
this is the magic that generates the different styles necessary to create themes. in order to define a color variable, it needs to be added to every single theme in the `$themes` map, and then used within the `themify` mixin, and retrieved using the `getThemifyVariable` function. there's lots of examples of this in the SCSS code to follow, but let me know if you're still confused. 

@machinic it would be awesome to get your input on the console theming! i think the approach of lightening or darkening from the core syntax theme could be nice, to distinguish from the code editor. though maybe that's overthinking it, and it could be as simple as differentiating the background color?

shinytang6:
Yes I have noticed that. But in this case I need to pass styles to the console-feed library instead of defining it in `_console.scss`, that is the problem

shinytang6:
Thanks, I will investigate!

machinic:
@catarak — cool. The reason I'd advocate for using some Sass function to adjust either the Lightness or the Saturate in the console related base colors for the theme is less about distinguishing the console from the code editor (those two spaces should have different backgrounds already) and more about how the colors interact with a dark or light background in general. 

Here's an example, these aren't the current error, warning colors in the editor right now — it's an attempt to mitigate some of what I'm talking about without making an adjustment for one theme or another (though a bit more accommodating it still fails, imo): 

console-color-studies 2x

When we did the syntax colors we dialed in specific sets of hex values for each theme for this reason, too. I thought that it might be a nice reprieve from that approach to lean into what we can get out of Sass.

shinytang6:
@catarak After some trials, l think it's difficult to export different styles based on the theme class in scss because webpack bundle all the files into a static file before running. 

Re-rendering is working now by adding some tricky code

catarak:
@machinic okay, got it! 

@shinytang6 glad you got it (somewhat) working! we can do some investigation of applying styles to the logs.

client/styles/components/_console.scss
```diff
@@ -18,28 +18,53 @@
...
+	.preview-console__debug {
+		flex: 1 0 auto;
+		position: relative;
+
+		.preview-console__debug__logged-times {
+			background-color: $dodgerblue;
+		}
+	}
+
+	.theme-with-dark-background, .theme-with-contrast-background {
```
catarak: if, as above, you change the class of the console messages to `.preview-console__message--${method}`, then you can change this to
```scss
.preview-console__message {
  @include themify() {
     color: getThemifyVariable('console-bg-color');
  }
}
```

And then you need to add this variable to the `$theme` map in `_variables.scss`.

client/styles/components/_console.scss
```diff
@@ -18,28 +18,53 @@
...
+		color: $black;
+	}
+
+	.theme-with-light-background {
```
catarak: if you make the change on line 62 you can delete this.

client/styles/components/_console.scss
```diff
@@ -18,28 +18,53 @@
 	}
 
 	// assign styles to different types of console messages
-	.preview-console__log {
-		@include themify(){
-			color: getThemifyVariable('secondary-text-color');
-		}
+	.preview-console__log, .preview-console__info{
```
catarak: Since now all of the console messages have the same class in common (`.preview-console__message`), you don't have to repeat these styles a bunch of times. You can do
```SCSS
.preview-console__message {
  flex: 1 0 auto;
  position: relative;
}
```

client/modules/IDE/components/Console.jsx
```diff
@@ -48,8 +85,23 @@ class Console extends React.Component {
...
             }
             return (
-              <div key={consoleEvent.id} className={`preview-console__${method}`}>
-                {Object.keys(args).map(key => <span key={`${consoleEvent.id}-${key}`}>{args[key]}</span>)}
+              <div key={consoleEvent.id} className={`preview-console__${method} theme-with-${theme}-background`}>
+                { times > 1 &&
+                  <div className={`preview-console__logged-times preview-console__${method}__logged-times`}>{times}</div>
+                }
+                <ConsoleFeed
+                  // logs={[
```
catarak: could you delete this comment?

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -33,12 +33,33 @@ class PreviewFrame extends React.Component {
...
         const args = message.arguments;
         Object.keys(args).forEach((key) => {
-          if (args[key].includes('Exiting potential infinite loop')) {
+          if (typeof args[key] === 'string' && args[key].includes('Exiting potential infinite loop')) {
             this.props.stopSketch();
             this.props.expandConsole();
           }
         });
       });
+      // console.log(pickBy(messageEvent.data[0], (value, key) => startsWith(key, 'arguments') || startsWith(key, 'method')));
```
catarak: could you delete this comment?

client/modules/IDE/components/Console.jsx
```diff
@@ -48,8 +85,23 @@ class Console extends React.Component {
               );
             }
             return (
-              <div key={consoleEvent.id} className={`preview-console__${method}`}>
-                {Object.keys(args).map(key => <span key={`${consoleEvent.id}-${key}`}>{args[key]}</span>)}
+              <div key={consoleEvent.id} className={`preview-console__${method} theme-with-${theme}-background`}>
+                { times > 1 &&
+                  <div className={`preview-console__logged-times preview-console__${method}__logged-times`}>{times}</div>
```
catarak: you should only need one class here: `.preview-console__logged-times`. i'll explain below how to adjust the SCSS.

client/styles/components/_console.scss
```diff
@@ -18,28 +18,53 @@
...
 		flex: 1 0 auto;
+		position: relative;
+
+		.preview-console__log__logged-times, .preview-console__info__logged-times{
```
catarak: if you change the logged times class to `.preview-console__logged-times` only, you can write this style code pretty concisely:
```
.preview-console__logged-times {
  .preview-console__message--info &, .preview-console__message--log & {
    background-color: $lightsteelblue;
  }
  .preview-console__message--warn & {
    background-color: $orange;
  }
  .preview-console__message--debug & {
    background-color: $dodgerblue;
  }
  .preview-console__message--error & {
    background-color: $red;
  }
}
```

client/styles/components/_console.scss
```diff
@@ -102,4 +127,15 @@
 	.preview-console--collapsed & {
 		display: none;
 	}
+}
+
+.preview-console__logged-times {
+	font-size: 10px;
```
catarak: This project doesn't use pixel measurements but uses REM's! Therefore this should be
```SCSS
  font-size: #{10 / $base-font-size}rem;
```

client/styles/components/_console.scss
```diff
@@ -102,4 +127,15 @@
...
+	font-weight: bold;
+	margin: 2px 0 0 6px;
+	border-radius: 100px;
```
catarak: Does this work if it's 100%?

client/styles/components/_console.scss
```diff
@@ -102,4 +127,15 @@
 	.preview-console--collapsed & {
 		display: none;
 	}
+}
+
+.preview-console__logged-times {
+	font-size: 10px;
+	font-weight: bold;
+	margin: 2px 0 0 6px;
```
catarak: This should also be in REMs instead of pixels.

client/styles/components/_console.scss
```diff
@@ -102,4 +127,15 @@
...
+	border-radius: 100px;
+	padding: 1px 4px;
```
catarak: Again, REMs and not pixels.

client/styles/components/_console.scss
```diff
@@ -102,4 +127,15 @@
...
+	z-index: 100;
+	float: left;
```
catarak: I don't like to use `float` because it's annoying. You should be able to change this to `left: 0` and i think you'll get the same effect.

catarak:
@shinytang6 let me know if you have any questions about my requested changes! If you're not familiar with BEM (Block Element Modifier) I'd recommend taking a look at this: http://getbem.com/naming/. Or if you're confused about any of the SCSS syntax, the SASS documentation is really good and I'd recommend taking a look at it.

client/styles/components/_console.scss
shinytang6: It's better to use px

client/styles/components/_console.scss
shinytang6: Thanks :) This is great

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -33,12 +33,32 @@ class PreviewFrame extends React.Component {
...
         const args = message.arguments;
         Object.keys(args).forEach((key) => {
-          if (args[key].includes('Exiting potential infinite loop')) {
+          if (typeof args[key] === 'string' && args[key].includes('Exiting potential infinite loop')) {
             this.props.stopSketch();
             this.props.expandConsole();
           }
         });
       });
+      for (let i = 0; i < messageEvent.data.length;) {
```
catarak: i can see what's going on here but this is a pretty dense block of code here. is there any way to clean this up a bit? break it out into its own separate function or functions? or maybe this could be done in the `messageEvent.data.forEach` loop above? You can use `Array.prototype.every` instead of `Array.prototype.forEach` break out of the loop early.

shinytang6: Yes, the code here is a bit redundant，l will try to clean it up

shinytang6:
Thanks for that @machinic ! I tried to change the background color of the console-logged-times according to you, it looks great and I can help change it. But there is a problem that the background color of the icons are immutable, we need embedded base64 URIs for the icons, e.g. the warning icon's URI is `data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs+9AAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAAAJcEhZcwAADsMAAA7DAcdvqGQAAACkSURBVChTbY7BCoJQFERn0Q/3BX1JuxQjsSCXiV8gtCgxhCIrKIRIqKDVzXl5w5cNHBjm6eGinXiAXu5inY2xYm/mbpIh+vcFhLA3sx0athNUhymEsP+10lAEEA17x8o/9wFuNGnYuVlWve0SQl7P0sBu3aq2R1Q/1JzSkYGd29eqNv2wjdnUuvNRciC/N+qe+7gidbA8zyHkOINsvA/sumcOkjcabcBmw2+mMgAAAABJRU5ErkJggg==`

catarak:
@shinytang6 i dug into the `console-feed` code a bit, and I don't think the icons need to be base64 URIs. if you look at the default values in https://github.com/samdenty99/console-feed/blob/d6c1b9fec2b352cbf622ca7f9c965e9c518491e0/src/Component/theme/default.ts, you'll see that the URI's are wrapped in a `url(...)`. This means that you could link to them also with an absolute or relative path. i think it would make sense to save all of the icons as images (or even better as svg's), and then import them to JS using webpack (by just doing `import logIcon from '../../images..'` in the Console component). if you're not comfortable preparing these icons, let me know! i'm happy to do it since it's outside of programming work.

shinytang6:
@catarak Yes, that's my original intention. I just gave an example because I saw that @machinic wanted to change the color of the icons with css. I think it would be better to provide two different sets of svg patterns to import.

machinic:
@shinytang6 @catarak Glad to hear we can make these moves. 

I'm most interested in working out colors that are both appropriate and systematic for each theme to form a coherent design whole — whatever is the most sensible implementation for icons and for the text strings I'm all for. 👍🏻

shinytang6:
@machinic That would be awesome. Thank you very much! I am preparing the icons according to your suggestion. l think they will be out soon

catarak:
This is coming along great! I noticed a couple of things:

1. There's a bunch of linting warnings:
screen shot 2018-06-08 at 12 43 52 pm
could you fix these?
2. Don't know if you're still working on this, but I think the colors are still off:
screen shot 2018-06-08 at 12 46 20 pm
screen shot 2018-06-08 at 12 46 35 pm
I know the colors are not totally finished, but I think for now it would be good if the text matches the icon color (except for when the text needs to be the primary text color because it's not legible otherwise).

client/utils/consoleUtils.js
```diff
@@ -16,11 +25,17 @@ export const hijackConsole = `var iframeWindow = window;
   methods.forEach( function(method) {
     iframeWindow.console[method] = function() {
       originalConsole[method].apply(originalConsole, arguments);
-  
       var args = Array.from(arguments);
+      var toString = Object.prototype.toString;
```
catarak: to do deep copying, could you use the `lodash` method?

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -86,9 +78,43 @@ class PreviewFrame extends React.Component {
   }
 
   componentWillUnmount() {
+    window.removeEventListener('message', this.handleConsoleEvent);
     ReactDOM.unmountComponentAtNode(this.iframeElement.contentDocument.body);
   }
 
+  handleConsoleEvent(messageEvent) {
+    if (Array.isArray(messageEvent.data)) {
+      messageEvent.data.forEach((message) => {
```
catarak: i think this loop could be merged with the second one—put everything in an `every` loop by piecing out the body of the `forEach` and putting it in the `every` loop, and `return false` if there's an infinite loop.

client/styles/components/_console.scss
```diff
@@ -102,4 +85,28 @@
...
+	font-weight: bold;
+	margin: #{2 / $base-font-size}rem 0 0 #{8 / $base-font-size}rem;
+	border-radius: 10px;
+	padding: #{1 / $base-font-size}rem #{4 / $base-font-size}rem;
+	z-index: 100;
+	left: 0;
+	position: absolute;
+
+	.preview-console__message--info &, .preview-console__message--log & {
+		background-color: $lightsteelblue;
```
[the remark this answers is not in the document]
shinytang6: For now, l just put them inside the `$theme`, l will modify them if a new design is out :)

client/utils/consoleUtils.js
shinytang6: I think we may not be able to do this because the code here will be run directly in the browser and cannot use the third-party libraries, right?

client/modules/IDE/components/PreviewFrame.jsx
shinytang6: done :+1:

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -84,22 +84,19 @@ class PreviewFrame extends React.Component {
 
   handleConsoleEvent(messageEvent) {
     if (Array.isArray(messageEvent.data)) {
-      messageEvent.data.forEach((message) => {
-        const args = message.arguments;
+      messageEvent.data.every((message, index, arr) => {
+        const { arguments: args } = message;
         Object.keys(args).forEach((key) => {
           if (typeof args[key] === 'string' && args[key].includes('Exiting potential infinite loop')) {
             this.props.stopSketch();
```
catarak: if this is true, you'll probably want to return false, right? maybe you could have some variable outside of the forEach loop, and if it gets set because there is an infinite loop, you could exit from the `every` loop early.

catarak:
thanks for making all of these changes! i'm going to test this a little and make sure it's all working but this looks great.

shinytang6:
Hi @catarak 
Yes, these methods can be added, but I thought that some of these methods will lose functionality because we have rewritten these methods.

> Uncaught DataCloneError: Failed to execute 'postMessage' on 'Window': #<Promise> could not be cloned.

The reason for this error is that trying to pass an object with methods to window.postMessage.

Previous implementation of that is simply `return (typeof i === 'string') ? i : JSON.stringify(i)`, this will lead to converting non-string variables, e.g Promise will convert to "{}", objects with methods will convert to undefined. It can be avoided with more type detection.

I also considered this question when I first explored this library. But l thought that there would be more adjustments to the code structure to implement it and l wondered if we need all these methods and these types of console information. But I can try if I decide to do that.

catarak:
I think that if we use the `Hook()` API, we can remove our `hijackConsole` and we will be able to handle all of the different methods and types properly. See https://github.com/samdenty99/console-feed#serialization.

shinytang6:
@catarak I had some troubles when I tried to use `Hook` to bind the iframe in `PreviewFrame.jsx`.
A simplified situation is as follows: https://codesandbox.io/s/jjz07mv94y
l tried to add 
```
Hook(this.iframeElement.contentWindow.console, (log) => {
        const decoded = Decode(log);
        this.setState(state => update(state, { logs: { $push: [decoded] } }));
 });
```` 
in `PreviewFrame.jsx` to catch iframe's console info, but it seems not working.

l tried to dynamically add some script to iframe using `this.iframeElement.contentWindow.document.write("<script>console.log('Hello!')</script>");` and they can be caught inside `Hook`. 

So it seems that the console info of **srcdoc** cannot be obtained from `iframe.contentWindow.console`? Do you have any ideas about it?

catarak:
@shinytang6 that's really annoying. i'm gonna do a little research and see what comes up.

catarak:
i was looking at how CodeSandbox handles this, and it appears that `Hook` is used on `window.console` (since the `iframe` is contained within `window`, its console messages apear in `window.console`). (see https://github.com/CompuIves/codesandbox-client/blob/b889de2713e654a8b41f23b9a65970609e6cd588/packages/app/src/sandbox/console/index.js) 

I'm not totally clear on how this doesn't catch errors that are in the `window` though. Hopefully that's a good start to making this work!

catarak:
That sounds good, try `react-frame-component` and see if that works. Another potential option is adding to `hijackConsole()` to make it work with the different data types, though I'm not sure how difficult that would be.

catarak:
closing this, as #656 starts over and takes this in a different direction.

Issue #640, opened 2018-05-20
npm install errors - v10.1.0 compatibility problem?


awarua:
#### Nature of issue?

- [X ] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Details about the bug: 

- Operating System: MacOSX 10.13.4
- node version: v10.1.0
- Steps to reproduce this:
``` sh
$ git clone git@github.com:processing/p5.js-web-editor.git
$ cd p5.js-web-editor.git
$ npm install
```

When I follow the installation instructions above (from the README), the npm installation fails with the following error (this is an extract of a much longer output): 

```sh
internal/modules/cjs/loader.js:573
    throw err;
    ^

Error: Cannot find module 'mkdirp'
```

I then try installing that module manually with `npm i mkdirp`, which leads to similar error for the `inherits` module. After that, I get an error for the `once` module. Then `inherits` again.

I'm not too familiar with npm/node, so maybe I'm doing something fundamentally wrong. I'm just trying to follow the README instructions.

awarua:
I kept adding modules to the command line until it finally installed. This is what I ended up using:
```sh
$ npm install inherits  core-util-is once pinkie-promise string-width object-assign strip-ansi which
$ npm install
```

catarak:
i'm not sure what the issue is, but there seems to be a problem with `package-lock.json`. i ran into this and i was able to fix it by removing `package-lock.json` and then running `npm install` again. Also, I'm using node 8.11.1.

awarua:
A further update. Although the `npm install` command was able to complete above, when I tried to run the server, it had an error.

I tried downgrading to node v8.11.2 with `brew install node@8` and it worked. So it seems like this might be a compatibility problem with the latest version of node? I will update the title of the issue to reflect this.

catarak:
i think there's also a bigger question here, which is how to handle enforcing contributors to use a certain node version (maybe just using Docker?) or how to handle different versions of node?

shinytang6:
I also encountered installation problems, maybe we should add `package-lock.json` to `.gitignore`?

catarak:
i think if `package-lock.json` is functioning properly it's supposed to be checked in. i ran into a really similar issue on a separate project i was woking on (unfortunately it's a private repo) and we fixed it by switching from npm to yarn. maybe that's the thing to do? i think there's some research to do.

awarua:
[...]

PR #642, opened 2018-05-24
Remove authorization from specific routes

### LINKED ISSUE #569: Add authorization to specific routes

shinytang6:

Thank you! Fixes #569

### DIFF
--- a/server/controllers/aws.controller.js
+++ b/server/controllers/aws.controller.js
@@ -108,6 +108,11 @@ export function copyObjectInS3(req, res) {
   });
 }
 
+export function listObjectsInS3(req, res) {
+  req.params = Object.assign({}, req.user);
+  listObjectsInS3ForUser(req, res);
+}
+
 export function listObjectsInS3ForUser(req, res) {
   const { username } = req.params;
   findUserByUsername(username, (user) => {
--- a/server/controllers/aws.controller.js
+++ b/server/controllers/aws.controller.js
@@ -108,13 +108,13 @@ export function copyObjectInS3(req, res) {
   });
 }
 
-export function listObjectsInS3(req, res) {
-  req.params = Object.assign({}, req.user);
-  listObjectsInS3ForUser(req, res);
-}
-
 export function listObjectsInS3ForUser(req, res) {
-  const { username } = req.params;
+  let username;
+  if (req.params) {
+    username = req.params.username;
+  } else {
+    username = req.user.username;
+  }
   findUserByUsername(username, (user) => {
     const userId = user.id;
     const params = {
--- a/server/controllers/aws.controller.js
+++ b/server/controllers/aws.controller.js
@@ -109,44 +109,41 @@ export function copyObjectInS3(req, res) {
 }
 
 export function listObjectsInS3ForUser(req, res) {
-  if (req.user) {
-    findUserByUsername(req.user.username, (user) => {
-      const userId = user.id;
-      const params = {
-        s3Params: {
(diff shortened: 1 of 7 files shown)


server/controllers/aws.controller.js
```diff
@@ -108,6 +108,11 @@ export function copyObjectInS3(req, res) {
   });
 }
 
+export function listObjectsInS3(req, res) {
+  req.params = Object.assign({}, req.user);
```
catarak: i think overwriting req.params is confusing—maybe `listObjectsInS3ForUser` could check if req.params.username is defined, and if not, use req.user?

catarak:
woohoo thanks for this! sorry about the travis build fail, it's related to #640.

shinytang6:
@catarak l am thinking that maybe `/:username/sketches` should also be removed, what do you think?

catarak:
@shinytang6 we should keep that as it's currently the only way to see another user's sketches.

server/controllers/aws.controller.js
shinytang6: done :)

catarak:
for this PR, you also need to change the routes, client-side, and change the routes in `client/components/Nav.jsx`.

shinytang6:
Do you mean removing all `/:username/assets` and `/:username/account` on client side? l think there's no need to remove that cause they are only valid when users logged in click 'My assets/settings'

catarak:
Yes, those routes should be moved to `/assets` and `/account` client side, and the `<Link>`s should be changed in the navigation as well.

server/routes/server.routes.js
```diff
@@ -57,7 +57,15 @@ router.get('/verify', (req, res) => {
 });
 
 router.get('/sketches', (req, res) => {
-  res.send(renderIndex());
+  req.user ? res.send(renderIndex()) : get404Sketch(html => res.send(html));
+});
+
+router.get('/assets', (req, res) => {
+  req.user ? res.send(renderIndex()) : get404Sketch(html => res.send(html));
```
catarak: After using this, I think it's confusing that this renders a 404 page if the user is not logged in—I think this should redirect to `/login`.

server/routes/server.routes.js
```diff
@@ -57,7 +57,15 @@ router.get('/verify', (req, res) => {
...
+});
+
+router.get('/account', (req, res) => {
+  req.user ? res.send(renderIndex()) : get404Sketch(html => res.send(html));
```
catarak: After using this, I think it's confusing that this renders a 404 page if the user is not logged in—I think this should redirect to `/login`.

server/routes/server.routes.js
```diff
@@ -57,7 +57,15 @@ router.get('/verify', (req, res) => {
 });
 
 router.get('/sketches', (req, res) => {
-  res.send(renderIndex());
+  req.user ? res.send(renderIndex()) : get404Sketch(html => res.send(html));
```
catarak: This should also redirect to login rather than be a 404!

catarak:
This is working great! Just one last thing to fix.

PR #652, opened 2018-06-16
Fix needs saving mark for renamed: Fixes #576

awarua:
Before your pull request is reviewed and merged, please ensure that:

* [ ] there are no linting errors -- `npm run lint` (there are linting errors, but not in the files I have touched.)
* [X] your code is in a uniquely-named feature branch and has been rebased on top of the latest master. If you're asked to make more changes, make sure you rebase onto master then too!
* [X] your pull request is descriptively named and links to an issue number, i.e. `Fixes #576`

This is an attempt at fixing the problem described in #576, which is that after a file in a project has been renamed, the 'needs-saving' marker is not cleared after saving. I'm mainly adding this pull request to show that it fixes the problem, but I don't think it's the best code. If I was more familiar with the lodash library, I would try and use one of the functions provided by that rather than writing my own custom comparison function. 

Another option is to look at the parts of the code that set the 'isEditingName' and 'isOptionsOpen' properties after files have been renamed. These are the properties that cause the comparision between the local copy of the project and that returned by the server to be different.

Issue #576, opened 2018-02-27
renaming files don't mark as saved


shiffman:
- Found a bug
- tested chrome, mac os

Steps to reproduce:
1. Create a new sketch, add some code.
2. Save sketch.
3. Rename `sketch.js` to `sketch1.js`.
4. Change some code.
5. Save sketch.

After renaming the dot to mark a sketch as saved never goes away and a warning pops up if you attempt to leave or reload the page. However, as far as I can tell the sketch is actually saved (when I force reload my latest changes are there.)

catarak:
thanks for submitting this!

dhruvdutt:
@catarak I would love to try this. Can you please describe the fix?

catarak:
i don't know off the top of my head how to fix this!

himanshuc3:
@catarak Since the popup appears, it means the state variable probably would be affected. Can it be some change in `scss` files that is required?

catarak:
@himanshuc3 i think it's more likely that the state isn't getting set properly—there's some logic that determines whether or not a project needs to be saved based on a sketch being changed by a user. i assume this logic is imperfect!

awarua:
I was looking at it and didn't get very far, but for what it's worth, it looks like some of the relevant properties and functions might be `unsavedChanges`, `setUnsavedChanges()`, `warnIfUnsavedChanges()` and `unsavedChangesDotUrl`.

Hopefully, that saves someone else a little bit of time to find out what to search on.

awarua:
I have been looking at this again. I believe that the piece of code that matters is [/client/modules/IDE/actions/project.js#L87](https://github.com/processing/p5.js-web-editor/blob/9958619fc4454104f55dc3f9a67726c8e4637150/client/modules/IDE/actions/project.js#L87)

```javascript
if (!isEqual(currentState.files, response.data.files)) {
  savedProject.files = currentState.files;
  dispatch(setUnsavedChanges(true));
} else {
  dispatch(setUnsavedChanges(false));
}
```

I have stepped through the code in the debugger and when you are saving a project with a renamed file, it takes the first branch, so it seems that the currentState.files are not the same as the files that the response has...

Because of this, I checked that if *any* of the files has been renamed, the unsaved changes dot will not disappear – even if the file that is altered is *not* the one that was renamed.

awarua:
[...]

PR #658, opened 2018-06-27
Fix needs saving mark 2 - for #576

awarua:

* [X] there are no linting errors -- `npm run lint`
* [X] your code is in a uniquely-named feature branch and has been rebased on top of the latest master. If you're asked to make more changes, make sure you rebase onto master then too!
* [X] your pull request is descriptively named and links to an issue number, i.e. `Fixes #123`

This is a work in progress pull request to address #576 based on advice from @catarak and @shinytang6 on my previous pull request (#652).

I am trying to move the properties `isOptionsOpen` and `isEditingName` to be stored as state in the `FileNode` component, rather than going into the Redux state. This PR just does this for `isOptionsOpen`. I want to check that I'm on the right track and am not doing anything too heinous...

### DIFF
--- a/client/constants.js
+++ b/client/constants.js
@@ -65,8 +65,9 @@ export const CLEAR_LINT_MESSAGE = 'CLEAR_LINT_MESSAGE';
 
 export const UPDATE_FILE_NAME = 'UPDATE_FILE_NAME';
 export const DELETE_FILE = 'DELETE_FILE';
-export const SHOW_EDIT_FILE_NAME = 'SHOW_EDIT_FILE_NAME';
-export const HIDE_EDIT_FILE_NAME = 'HIDE_EDIT_FILE_NAME';
+// Constants for show/hide file name editing
+// export const SHOW_EDIT_FILE_NAME = 'SHOW_EDIT_FILE_NAME';
+// export const HIDE_EDIT_FILE_NAME = 'HIDE_EDIT_FILE_NAME';
 
 export const SET_AUTOSAVE = 'SET_AUTOSAVE';
 export const SET_LINT_WARNING = 'SET_LINT_WARNING';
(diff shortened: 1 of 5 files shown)


catarak:
yeah this is looking great! the only suggestion i have is to remove all of the commented out code.

awarua:
Thanks @catarak - I've completed the changes for `isEditingFileName` too. Have also removed commented out code. 

One question I had, which you can maybe advise on: Should I be binding the event handlers I've added to the `FileNode` component to this in the constructor for that?

This appears to work fine on my local installation and solves the problem of the needs saving mark not being removed.

catarak:
Yes! The event handlers should be bound to `this` in the constructor. The reason for this is so that in the event handler, you can access the state of the React component (which is `this`). If you don't bind it, then I think by default `this` is the HTML element that the event handler is added to, though I'm not 100% sure on that.

awarua:
@catarak - great! I actually did do that in the latest changes, so I think that means that this PR is complete and will leave it from you for here.

catarak:
this looks great! i noticed a bug, which is if you rename a file, it doesn't save the sketch, or mark it as unsaved. i think that when you rename a file, the sketch should save.

bendman:
@catarak Is this PR just waiting for a fix to this last bug?

> [...] i noticed a bug, which is if you rename a file, it doesn't save the sketch, or mark it as unsaved. i think that when you rename a file, the sketch should save.

catarak:
@bendman yes! feel free to pick up and fix. i don't think it makes sense to merge this in as it is now since the bug i found is a regression.

bendman:
@catarak I'm working on this one now, but I'm not understanding how to reproduce the regressive bug. Right now on the live site I also see the bug where renaming a file doesn't save the sketch or mark it as unsaved.

This happens on the live site and this branch for me: If I rename a file, it doesn't save it or mark it as unsaved. If I reload the project, the file will have the original name.
p5-save-bug

The existing behavior _is_ kind of weird though. Should I try to include a fix in this branch by saving the project once a file is renamed?

catarak:
@bendman you're right! i think i was mixed up since when you change the sketch name, the sketch automatically saves, and now i'm wondering if that should be the same behavior for files, or if the sketch should set its state to having unsaved changes. i'll put this in a separate ticket.

let me take a second pass at this PR and see if there are any other changes to be made, otherwise i will _actually_ merge it in.

catarak:
just tested again, and this is working great. hopefully this also fixes #675. going to merge this!

PR #685, opened 2018-09-08
Fixes #680 by adjusting canvas and iframe layout

### LINKED ISSUE #680: `windowHeight` is 30px too big using the editor
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

#### Details about the bug: 

`windowHeight` doesn't match the height of the window and is off by about 30px.

- Web browser and version: Safari 12.0 (14606.1.36.1.9) & Chrome
- Operating System:  MacOSX
- Steps to reproduce this:
use `createCanvas(windowWidth, windowHeight);` in the editor, it is 30px too big to fit the screen.
You can do `createCanvas(windowWidth-10, windowHeight-40);` to fix it but it is still a bug.
[Here is a test sketch you can use](https://editor.p5js.org/belgiancoder/sketches/BytOy7eu7), You should be able to see where the lines intersect.

screenshot 2018-09-07 at 18 29 22

#### New feature details:

Just correct the value of `windowWidth` and `windowHeight`. `windowWidth` is correct but it would be nicer if there was the same padding on both sides (10px).
Here is the expected result. 
screenshot 2018-09-07 at 18 32 29

#### Update: Chrome has this too.
screenshot 2018-09-07 at 18 53 38

bendman:
- Set the preview canvas to `display:block` to remove vertical spacing
caused by the default value, which was making a full-height canvas cause
vertical scrolling, and the vertical scrollbar caused horizontal
scrolling
- Wrap the preview content in a container with `position:relative` and
no other visible content, so that the full-height iframe excludes the
height of the preview frame header, preventing it from going beyond the
height of the page

Fixes #680

image

>Before your pull request is reviewed and merged, please ensure that:
>
>* [x] there are no linting errors -- `npm run lint`
>No additional errors
>* [x] your code is in a uniquely-named feature branch and has been rebased on top of the latest master. If you're asked to make more changes, make sure you rebase onto master then too!
>* [x] your pull request is descriptively named and links to an issue number, i.e. `Fixes #123`
>
>Thank you!

### DIFF
--- a/client/modules/IDE/pages/IDEView.jsx
+++ b/client/modules/IDE/pages/IDEView.jsx
@@ -370,41 +370,43 @@ class IDEView extends React.Component {
                 <header className="preview-frame__header">
                   <h2 className="preview-frame__title">Preview</h2>
                 </header>
-                <div className="preview-frame-overlay" ref={(element) => { this.overlay = element; }}>
+                <div className="preview-frame__content">
+                  <div className="preview-frame-overlay" ref={(element) => { this.overlay = element; }}>
+                  </div>
+                  <div>
+                    {(
+                      (
+                        (this.props.preferences.textOutput ||
+                            this.props.preferences.gridOutput ||
+                            this.props.preferences.soundOutput
+                        ) &&
+                            this.props.ide.isPlaying
+                      ) ||
+                        this.props.ide.isAccessibleOutputPlaying
+                    )
+                    }
+                  </div>
+                  <PreviewFrame
+                    htmlFile={this.props.htmlFile}
+                    files={this.props.files}
+                    content={this.props.selectedFile.content}
+                    isPlaying={this.props.ide.isPlaying}
+                    isAccessibleOutputPlaying={this.props.ide.isAccessibleOutputPlaying}
+                    textOutput={this.props.preferences.textOutput}
+                    gridOutput={this.props.preferences.gridOutput}
+                    soundOutput={this.props.preferences.soundOutput}
+                    setTextOutput={this.props.setTextOutput}
+                    setGridOutput={this.props.setGridOutput}
+                    setSoundOutput={this.props.setSoundOutput}
+                    dispatchConsoleEvent={this.props.dispatchConsoleEvent}
+                    autorefresh={this.props.preferences.autorefresh}
+                    previewIsRefreshing={this.props.ide.previewIsRefreshing}
+                    endSketchRefresh={this.props.endSketchRefresh}
+                    stopSketch={this.props.stopSketch}
+                    setBlobUrl={this.props.setBlobUrl}
+                    expandConsole={this.props.expandConsole}
+                  />
                 </div>
-                <div>
-                  {(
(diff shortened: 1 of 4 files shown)


client/modules/IDE/reducers/files.js
```diff
@@ -31,6 +31,9 @@ const defaultCSS =
   margin: 0;
   padding: 0;
 }
+canvas {
```
catarak: just curious, is there a way to fix this without adding this CSS? the problem i have with this is that all sketches made before this PR would still be broken, which maybe is inevitable and that's okay, but if there's another way that would be coo.

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -370,41 +370,43 @@ class IDEView extends React.Component {
                 <header className="preview-frame__header">
                   <h2 className="preview-frame__title">Preview</h2>
                 </header>
-                <div className="preview-frame-overlay" ref={(element) => { this.overlay = element; }}>
+                <div className="preview-frame__content">
+                  <div className="preview-frame-overlay" ref={(element) => { this.overlay = element; }}>
```
catarak: this is probably a task for a separate issue/PR but i was looking at this and maybe this overlay thing isn't necessary anymore. i added it to make the pane resizing work. the dragging resizer thing used to be positioned over the iframe, and the overlay was necessary for the iframe not to capture the mouse events, but then #441 happened, and it got moved to be between the code editor and the preview.

client/modules/IDE/reducers/files.js
bendman: Not that I know of. Without some CSS (`display: block` or `vertical-align: middle`) there is a gap below the canvas because it's inline. It makes it so people can see the whole canvas, but they need to scroll a bit.

It works pretty well without the CSS if you have scrollbars disabled like the default on OSX:

image

But when scrollbars are visible it gets worse. I don't have a windows machine, so I haven't tried it out, but I suspect Windows always has scrollbars visible 😞 

image

catarak: got it! sorry for the delay on this one, but just had a thought—say a user specifically made the height of the canvas larger than the width/height of the preview pane. i think it would make sense for the scrollbars to appear, right? if we included this css, would that make the scrollbars disappear for that case? i feel like it would make sense to keep them in that case?

catarak: just an update—i've been working on getting a browserstack account (which is free for open source projects) so i can test this.

catarak: just tested this on windows—scrollbar does appear in the use case when you'd want it to appear.

catarak:
apologies for this taking so long to merge in—it works great!

client/modules/IDE/reducers/files.js
bendman: Oh great! Sorry I never got back to this, I couldn't get this to run locally (an issue on my end with some dependency), so I wasn't able to test it again.

PR #713, opened 2018-10-04
Fixes #692

### LINKED ISSUE #692: include a link to the sketch in editor from the default fullscreen url?
Not sure how this relates to #208, but an issue I see a lot in class is that students will link to the full-screen version of their sketch in their blog posts, and then when they're presenting in-class and I ask to see their source code, there's no easy way for them to get back to the editor view.

A possible solution might be to include an inconspicuous link (say, a 50% transparent "Back" icon in the lower-right corner?) that only appears in the full screen view.

Another possible solution might be to make the mapping between the full-screen URL and the editor URL for a sketch more clear, so that it would be easy to modify the URL to get back to the original sketch. (As it stands, the full-screen URL doesn't include the username of the creator, so you can't just get rid of `full/` in the URL to see the sketch in the editor.)

I suspect both of these solutions have significant drawbacks, but I wanted to make the suggestion in case other people have the same issue and it turns out that there's an easy way to implement a fix. Thanks!

Vertmo:

* [x] all the changes you wanted to make are done

Thank you!

I'm making some progress on the #692 issue : I've changed the routes, and I now only have to style the "full" header, and to modify the "Share" window.
However I also have a question : with the current modification, I totally delete the old full and embed routes (the one not using the username in the url). This WILL break existing embed links. Should I reintroduce the old routes as an option to avoid this problem ?


catarak:
re: your question, good call! i think the `/full/<sketch_id>` links should not change for backwards compatibility, and for the sake of reducing confusion, `/username/full/<sketch_id>` should also not change.

okay. after thinking about it, here's my updated opinion on what all of these urls should be:
1. `/<username>/sketches/<sketch_id>` is the edit view
2. `/<username>/full/<sketch_id>` is the full screen view, without a header. `/full/<sketch_id>`
3. `<username>/present/<sketch_id>` is the view which also contains a header with the username, the sketch name, and a link to the edit view, as well as the rendered sketch in an iframe.

Vertmo:
So the embed is renamed to full and the full is rename to present ? That would also break existing embeddings right ?I think the simplest solution would be to keed the old urls functional, and to add the new complete ones on top.
Good thought ! I will see what I can do about rendering client size.

catarak:
sorry to change my mind, but i think i misinterpreted your question—i thought you were saying that the 'full' links would break because folks would expect them not to have a header. now i get it. my _updated_ thoughts:

1. `/<username>/sketches/<sketch_id>` is the edit view
2. `/<username>/embed/<sketch_id>` is the full screen view, without a header. `/embed/<sketch_id>` also works for backwards compatibility but is not populated in the share modal.
3. `/<username>/full/<sketch_id>` is the view which also contains a header with the username, the sketch name, and a link to the edit view, as well as the rendered sketch in an iframe. `/full/<sketch_id>` also works for backwards compatibility but is not populated in the share modal.

Vertmo:
No problem ! That's duly noted, I'll work on that tomorrow ^^

Vertmo:
This should be good this time ^^ Tell me what you think !

client/modules/IDE/components/CopyableInput.jsx
```diff
@@ -42,7 +43,12 @@ class CopyableInput extends React.Component {
           onMouseLeave={this.onMouseLeaveHandler}
         >
           <label className="copyable-input__label" htmlFor={`copyable-input__value-${label}`}>
-            {label}
+            <div style={{ 'display': 'flex', 'justify-content': 'space-between' }}>
```
catarak: can you add these styles to the scss? all of the scss lives in `client/styles`. you'll need to add a class to this div!

client/modules/IDE/components/CopyableInput.jsx
```diff
@@ -42,7 +43,12 @@ class CopyableInput extends React.Component {
...
+              {label} {hasPreviewLink ? (
+                <a style={{ 'text-align': 'right' }} target="_blank" href={value}>
```
catarak: same here—can you add this style to the scss?

catarak: you could also make this react logic simpler:
```js
{{ hasPreviewLink && <a>...</a>}}
```

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -317,23 +317,31 @@ class PreviewFrame extends React.Component {
...
+    doc.width = `${matches[1]}px`;
+    doc.height = `${matches[2]}px`;
   }
 
   render() {
     return (
       <iframe
+        id="canvas_frame"
         className="preview-frame"
+        width="720px" // TODO
```
catarak: do these values (width and height) have to be set? did you mean to return to this?

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -317,23 +317,31 @@ class PreviewFrame extends React.Component {
...
+        height="400px" // TODO
         aria-label="sketch output"
         role="main"
+        scrolling="no"
```
catarak: there's maybe an (edge) case where we'd want to leave scrolling in—for example if someone made a purposefully very large sketch.

client/styles/layout/_fullscreen.scss
```diff
@@ -5,13 +5,33 @@
...
+.fullscreen-preview__header {
+  text-align: center;
+  font-size: 16px;
+  font-family: Montserrat, sans-serif;
+  margin-bottom: 5px;
+}
+
 .fullscreen-preview__title {
   width: 100%;
+  color: #ed225d;
```
catarak: rather than adding these colors inline, you'll need to pull them from variables in `client/styles/abstracts/_variables.scss`. this should also pull from whatever the user has set as their theme! and by user i mean person logged into the editor, rather than the owner of the sketch.

client/modules/IDE/components/CopyableInput.jsx
Vertmo: Actually, the text-align attribute was not even necessary, so I simply removed it ^^

client/modules/IDE/components/PreviewFrame.jsx
Vertmo: Oh yeah, these were only temporary and I forgot to remove them. My bad !

client/modules/IDE/components/PreviewFrame.jsx
Vertmo: I tried, but this kind of yield weird results, since the iframe is slightly bigger than the canvas it encloses. That means that either I allow an error margin when computing the size of the iframe from the size of the canvas (but then the canvas will be slightly off-center) or the scroll bars are there permanently. Neither seem like good solutions. Do you have another idea ?

client/styles/layout/_fullscreen.scss
Vertmo: I'm currently trying to resolve this, but I'm kind of having a problem with sass : the `@include` that I'm adding to the `_fullscreen.scss` file are not taken into account in the developpement environment (whereas if I add them to another file, it works perfectly). Has anyone ever had a similar problem ?

client/modules/IDE/components/PreviewFrame.jsx
catarak: i think this issue is fixed by #685—let me look into merging it!

catarak:
This is working pretty well! I have some design feedback, but I think instead of writing it up, I'll make some quick wireframes for you.

Also, there's a lot of React errors in the console, from `props` not being passed to the `PreviewFrame` component. Can you fix these?

Vertmo:
I think these errors were "already there" before, only unseen because the component wasn't really use. I'll fix them then !

Vertmo:
Should the canvas still be centered inside the fullscreen view ? And as for the </> icon, should I use a free one and credit the author (where ?) or do you already have one I can use ?

catarak:
- The canvas should *not* be centered.
- For now, you can use the [IcoMoon free](https://icomoon.io/#preview-free) code icon (go to that link and then search "code").

Vertmo:
Done ! Tell me what you think ^^

catarak:
this is coming along great! a few more things:
- in the fullscreen view, the sketch title should link to the edit view of that sketch, and it should not be bold, should be the same font size as the header in the edit view, and the colors should match the title in the edit view.
- in the fullscreen view, the username should link to `/<username>/sketches`, and it should not be bold, should be the same font size as the header in the edit view, and the colors should match the title in the edit view.
- the "code" icon should be the same size as the play/stop/settings icons. It should be #a0a0a0 and on hover, should be the p5.js pink. 
- the header should match the theme of the logged-in user—this means the background color and text color should change.
- in the share modal, i don't think there needs to be a link to the edit view, since you can just close out of the share modal and it's there. 
- Rather than "Preview in new tab", i think the copy can just be "Open". I'm not quite sure this is the right placement for that link, but i think it's good for ow.

Vertmo:
There you go !

catarak:
yay! getting closer! a couple more things:
- i didn't mean take out the "edit" url from the share modal completely, just the link to open it in a new tab.
- the "code" icon still seems really big—could you make it smaller?

Vertmo:
Oh right ! Sorry about the "edit" url. Reverting now.
About the "code" icon, are you sure ? It seems big to me too, but it should be the same size as the play/stops/settings logos, as you asked. I think it's because it's not in a "square" box that it appears this way (it's wider than it's tall).

catarak:
I think with icons, for them to be the "same size", it means that they fit into the same square area. So even though the "code" icon is the same height, I think it's too wide! Does that make sense?

client/modules/IDE/components/PreviewFrame.jsx
catarak: could you also remove this stuff that tries to fix #680? i want to keep this PR contained to #692 if possible.

client/styles/components/_preview-frame.scss
```diff
@@ -1,8 +1,7 @@
 .preview-frame {
-  min-height: 100%;
-  min-width: 100%;
   position: absolute;
   border-width: 0;
+  overflow: hidden;
```
catarak: same with this, this is related to #680, can you take this out?

client/styles/layout/_fullscreen.scss
```diff
@@ -3,10 +3,29 @@
...
   height: 100%;
   flex-flow: column;
+  @include themify() {
+    background-color: getThemifyVariable('background-color');
+  }
+}
+
+.fullscreen-preview__header {
+  text-align: center;
+  font-size: 16px;
```
catarak: could you use `rem`s instead of pixel values?

client/modules/IDE/components/PreviewFrame.jsx
Vertmo: Right, I'm on it. Turns out my code was pretty dirty and non standard on that point. I'm correcting it right now ^^

client/styles/layout/_fullscreen.scss
Vertmo: Oh right, actually these classes were not even used anymore, so I'll just remove them

Vertmo:
Okay, I juste made all of the requested changes but one : I'm still puzzled about the logo, because it appears (in my browser anyway) to be the same width as the circle bounding the play, pause and settings logo (44px wide to be precise), so I don't really now what to do with it...
Oh and I also rebased.

catarak:
thanks! i'll take a close look at it.

client/styles/components/_preview-nav.scss
```diff
@@ -0,0 +1,14 @@
+.preview-nav__editor-svg {
+  & svg {
+    width: #{44 / $base-font-size}rem;
```
catarak: okay. i think i figured it out! i tried changing the width and height here from 44 to 22, and i think it looks better. i think we were comparing the size of different things! i meant the play/stop icon as in just the triangle/square icon, and i think that you thought i meant the icon + the circular button surrounding it.

Vertmo: That must be it ! Fixed !

catarak:
yay! i'm going to merge this in ✨

PR #718, opened 2018-10-04
Fixes #712

### LINKED ISSUE #712: Highlight matching brackets in code editor
#### Nature of issue?

- [ ] Found a bug
- [X] New feature request

#### New feature details:
In the code editor, it would make it easier to debug code if when you have selected or have the cursor next to a bracket, it would highlight the matching one. 

This is a feature built into CodeMirror. To add this feature, the steps are
1. Include the CodeMirror `edit/matchbrackets.js` file.
2. Add styling for `CodeMirror-matchingbracket`, `CodeMirror-nonmatchingbracket`.

You can read more about these in the [CodeMirror manual](https://codemirror.net/doc/manual.html).

Janglee123:


client/styles/vendors/_codemirror.scss
```diff
@@ -133,8 +133,6 @@
 
 /* Default styles for common addons */
 
-div.CodeMirror span.CodeMirror-matchingbracket {color: #0f0;}
```
shinytang6: I think there's no need to modify this file, right? This is the default style file for CodeMirror, it will be overwritten by the custom style.

client/styles/components/_p5-contrast-codemirror-theme.scss
```diff
@@ -97,8 +97,13 @@ $p5-contrast-activeline: #999999;
 }
 
 .cm-s-p5-contrast .CodeMirror-matchingbracket {
-  outline: 1px solid $p5-contrast-darkgray;
-  color: black !important;
+  color: $p5-contrast-green !important;
```
catarak: can you add these styles without using `!important`? in this project i prefer to not use it unless absolutely necessary.

catarak:
@Janglee123 also, I appreciate you trying out a design for this! I looked at some other code editors (i.e. what I have set up right now in Visual Studio Code):
screen shot 2018-10-05 at 5 04 14 pm
I'd prefer for the styling of the matching brackets to put a 1px border on the brackets, the same color as the comments for the respective theme, i.e. for the light theme, that's #A0A0A0. I was looking at how to do this, and you can't use `border` because that changes the size of the `<span>`. Instead, you'll need to add a `box-shadow`:
```css
.cm-s-p5-light .CodeMirror-matchingbracket {
    box-shadow: inset 0 0 0 1px #A0A0A0;
}
```
then, for the `.CodeMirror-nonmatchingbracket`, I would like for there to be no added style.

Janglee123:
@catarak I used `outline` instead of `box-shadow` as it is defined by default in style and using  `!important`  seems to be required because there are 2 more files([this](https://github.com/processing/p5.js-web-editor/blob/e83654f6ad2593c06bfbb169e636f86fef7be68a/client/styles/build/css/main.css#L556) and [this](https://github.com/processing/p5.js-web-editor/blob/e83654f6ad2593c06bfbb169e636f86fef7be68a/client/styles/vendors/_codemirror.scss#L136)) where style is defined for matching brackets. and I updated style just like you said. Hope you like it.

client/styles/components/_p5-light-codemirror-theme.scss
```diff
@@ -21,7 +21,7 @@ $p5-light-white: #FDFDFD;
 $p5-light-orange: #EE9900;
 $p5-light-lightgray: #E0D7D1;
 $p5-light-darkgray: #666666;
-$p5-light-green: #58a10b; 
+$p5-light-green: #58a10b;
```
catarak: could you take out this variable since we're not using it?

catarak:
@Janglee123 this looks great! one way to get around using `!important` would be to make the matching bracket style have a higher [css specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity). however, i think this would make the theme files less consistent, so i think leaving the `!important` styles is okay.

Issue #681, opened 2018-09-07
When changing a file name, sketch does not mark itself as unsaved


catarak:
#### Nature of issue?

- [X] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Details about the bug: 

- Web browser and version:  Google Chrome | 68.0.3440.106 (Official Build) (64-bit)
- Operating System:  Mac OSX
- Steps to reproduce this:

As discussed in #658,
1. Open an existing sketch, or save a new sketch
2. Open the sidebar
3. Click on the "v" next to a file name, and edit the file name
4. Hit enter or click outside of the text input
5. If you refresh the page, the old file name is still there. And since the sketch is not marked as unsaved, there is no warning that the sketch is unsaved.

Expected Behavior:
1. Open an existing sketch, or save a new sketch
2. Open the sidebar
3. Click on the "v" next to a file name, and edit the file name
4. Sketch is marked as "unsaved"

catarak:
@bhawesh96 i think that i agree—i think the behavior matches how a desktop code editor works, and i like that.

Janglee123:
@catarak  I want to work on this but can not find where to start, can you help me ?

catarak:
Sure! If you look at the [input for a file in the sidebar](https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/components/FileNode.jsx#L138), when a user blurs the input (i.e. clicks out of it or hits enter), it calls a function called [validateFileName()](https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/components/FileNode.jsx#L51), which dispatches an action by calling [updateFileName()](https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/components/FileNode.jsx#L62). `updateFileName()` only updates the file client-side, in the redux store, so this data would have to be propagated to the server. 

There's a few different ways to approach this:
1. Save the sketch after updating the file name. This could be hard, since there's no way to tell when the redux store has finished updating, and I think this is not best redux practices.
2. Create a new API endpoint, for updating just the file name. This would mean making an AJAX request, creating an endpoint in express server-side, and figuring out the MongoDB query to make this work! There's a lot of steps here, but you could look at [saveProject()](https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/actions/project.js#L67) for reference.

-

EOstling:
@catarak I would like to help with this issue, what needs to be done?

[...]

PR #739, opened 2018-10-23
[WIP] #681 Fixes 

Janglee123:
Renaming file will save file on server side with new name.


client/modules/IDE/actions/files.js
```diff
@@ -156,6 +156,25 @@ export function updateFileName(id, name) {
...
 }
 
+export function saveFile(id) {
+  return (dispatch, getState) => {
+    const state = getState();
+    const UpdatedFile = state.files.find(file => file.id === id);
+    if (UpdatedFile && state.project.id) {
+      axios.put(`${ROOT_URL}/projects/${state.project.id}/files/${id}`, UpdatedFile, { withCredentials: true })
+        .then(() => {
+          dispatch(setUnsavedChanges(false));
```
Janglee123: There is a new issue with a dot appearing after file name that used to indicate unsaved changes. File will be saved after renaming file thus dot should have to disappear. This behaviour controlled by setUnsavedChanges()`. But this function is used for all files in project. If  I don't use it here there are two problems, one with a dot and second is warning message show when leaving site. When use of it cause problem for other unsaved files.

catarak: yeah, i've noticed this! it's been super annoying. i'm not sure what's going on here. i may revert a commit that seems to have made this worse.

catarak:
@Janglee123 is this ready for feedback/testing?

Janglee123:
@catarak yes

client/modules/IDE/components/FileNode.jsx
```diff
@@ -61,6 +61,7 @@ export class FileNode extends React.Component {
     ) {
       this.props.updateFileName(this.props.id, this.originalFileName);
     }
+    this.props.saveFile(this.props.id);
```
catarak: One issue is that if the file name is invalid (if one of the above if statements is true), then the file name will erroneously get updated.

catarak: also, i think _only_ the file name should be saved, and not the entire file.

server/controllers/file.controller.js
```diff
@@ -119,3 +119,25 @@ export function getFileContent(req, res) {
     res.send(resolvedFile.content);
   });
 }
+
+export function updateFileName(req, res) {
```
catarak: i think that the file name should be extracted client side, rather than server side. therefore, the file should be updated based on whatever is in `req.body`, and if you only want to update the name, you should only send the name.

this also means i think this function should be named `updateFile`.

client/modules/IDE/actions/files.js
```diff
@@ -156,6 +156,24 @@ export function updateFileName(id, name) {
   };
 }
 
+export function saveFileName(id) {
+  return (dispatch, getState) => {
+    const state = getState();
+    const selectedFile = state.files.find(file => file.id === id);
```
catarak: i think that here, you should extract only the file name, and only `PUT` that to the API, rather than doing this server-side.

Janglee123:
I tried that but at server side I got this error message  when request is sent to server
 ```
in requests of type error
 [object Object]
```
and this is a line to send put request to server side 
```
axios.put(`${ROOT_URL}/projects/${state.project.id}/files/${id}`, selectedFile.name, { withCredentials: true })
```
Googling didn't help me, @catarak  I will do it but again a little help is needed.

server/controllers/file.controller.js
Janglee123: I think `updateFileName` is more suitable as we are only saving file name not whole file. `updateFile` may be misleading.

catarak: yeah i hear you! i'm thinking that this api endpoint should update the whole file, rather than just the name, and then it can be used to update just the file name, or the file content, or both, depending on what is in the body of the request. does that make sense?

Janglee123:
done it. @catarak You can suggest changes.

PR #746, opened 2018-10-28
Fix async validation in signup form (fixes #742)

### LINKED ISSUE #742: Sign up form doesn't validate username uniqueness
#### Nature of issue?

- [X] Found a bug
- [ ] New feature request

#### Details about the bug: 

- Web browser and version: 
- Operating System: 
- Steps to reproduce this:

I noticed that in the web editor logs, the node application has been crashing because of an attempt to create a new user with an existing username. In the sign up form, there should be validation to prevent this, but it appears to not be working. Would be nice to fix this soon as it's causing crashes in the application!

anaplian:
I fixed the exception caused by signing up with a username that is already in use, and added support for multiple asynchronous errors in the sign up form (previously, moving focus from the email field would erase errors in the username field.)

### DIFF
--- a/client/modules/User/pages/SignupView.jsx
+++ b/client/modules/User/pages/SignupView.jsx
@@ -59,10 +59,26 @@ class SignupView extends React.Component {
   }
 }
 
+function asyncErrorsSelector(formName, state) {
+  const form = state.form[formName];
+  if (!form) {
+    return {};
+  }
+
+  const fieldNames = Object.keys(form).filter(key => !key.startsWith('_'));
+  return fieldNames.reduce((asyncErrors, fieldName) => {
+    if (form[fieldName].asyncError) {
+      return { ...asyncErrors, [fieldName]: form[fieldName].asyncError };
+    }
+    return asyncErrors;
+  }, {});
+}
+
 function mapStateToProps(state) {
   return {
     user: state.user,
-    previousPath: state.ide.previousPath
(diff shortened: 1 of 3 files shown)


client/modules/User/pages/SignupView.jsx
```diff
@@ -71,21 +87,29 @@ function mapDispatchToProps(dispatch) {
...
-        if (response.data.exists) {
-          const error = {};
-          error[fieldToValidate] = response.data.message;
-          throw error;
-        }
-      });
-  }
-  return Promise.resolve(true).then(() => {});
+  const errors = {};
+  return Promise.resolve(true)
```
catarak: is there a better way to write this code? the promises are nested here, and usually with promises it makes sense to have them only one level deep, if that make sense.

client/modules/User/pages/SignupView.jsx
```diff
@@ -59,10 +59,26 @@ class SignupView extends React.Component {
   }
 }
 
+function asyncErrorsSelector(formName, state) {
```
catarak: is this is code necessary for keeping around around multiple asyncErrors? if yes, that's cool, i'm just curious if there's a way to do this that feels slightly less hacky, that's built into redux form.

catarak:
took a deeper look at this. tbh i think the way that promises mix with conditionals is super confusing! however, i couldn't think of another way to structure this, so i'm going to merge it 🎉

Issue #734, opened 2018-10-18
Updating assets


slowizzm:
- [x ] Existing feature enhancement

Instead of adding a `-#` to the end of a file that is being re-uploaded (or more specifically a file that has the same name as an existing file in the project), it would be nice if it would just overwrite that existing file.

catarak:
thanks for reporting! i think it would be confusing if it just overwrote the old asset. however, maybe this can function like in an operating system—the editor could show an alert, asking if the user wants to overwrite the file or keep both.

slowizzm:
Yes, I agree. An alert would be best to prevent user error.

LakshSingla:
@catarak @slowizzm What are the different ways in which one can add files ? Are there any other ways than clicking on the project name to reveal the add file/folder dropdown ?

LakshSingla:
@catarak I would like to work on this issue. Can you please help me with which functions needs refactoring. Also, the only method I know to add a file in the ide is to click on the new file or new folder dragdown, in the sidepane. Are there any other methods to achieve it?

slowizzm:
Yes, that is the only method I know of to upload files. I have also noticed that you can only upload 6 files at a time.

LakshSingla:
@catarak @slowizzm I found the redux action responsible for producing the unique filename. Is there any pre existing modal I can reuse for issuing the warning?

catarak:
yeah! i would use the component `client/modules/App/components/Overlay.jsx`. Maybe it would make sense for the overlay to have a few buttons which say "Keep Both", "Replace", "Cancel", and then the "X" would also cancel.

catarak:
in my opinion, the amount of effort it would take to work on this issue exceeds its benefit, and therefore i think it makes sense to close this issue. feel free to reopen if you think otherwise!!

slowizzm:
@catarak That makes sense. I would like to say that I have students every semester frustrated with this because they often need to re-edit their assets and when some of those assets are image sequences this can be a tedious process. One solution for this is for me to just have them work in vs code, but then that takes them out of the p5 environment and I love using this editor as a teaching tool for art and media students new to code. 

I will let you be the judge to re-open, as I am sure there are higher-priorities needing attention.

catarak:
got it! maybe the solution here is not being able to edit assets, but for the flow for uploading/renaming/deleting assets to be easier.

slowizzm:
[...]

PR #759, opened 2018-11-09
[WIP]Rough draft of file/folder name collision resolution #734

LakshSingla:
- Add a check in the createFile action for repeated files

 - Add a NameCollisionModal.jsx[WIP] and corresponding reducers and actions to show and hide it

@catarak This is a very rough draft regarding issue #734. A check is applied on collision of file names only. What I had in mind was that instead of directly calling the `createFile` action we call some other action which would in turn call `createFile` if there was no name collision, else it would display the `NameCollisionModal.jsx` which would in turn call `createFile` if the user wants to keep both the files and `returns` if the user presses cancel. (I am thinking whether we should keep replace file option as we are only adding empty files and it would be redundant).
Please confirm if the logic to achieve this issue is fine, or it needs refinement.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha1-zgBCgEX7t9XrwOp7+DV4nxU2arI=",
       "requires": {
         "@babel/types": "7.0.0-beta.51",
-        "lodash": "^4.17.5"
+        "lodash": "4.17.10"
       },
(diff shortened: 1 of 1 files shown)


LakshSingla:
@catarak Did you review the changes ?

catarak:
@LakshSingla i have not yet, i'll do it soon!

PR #766, opened 2018-11-17
Serve up an empty 404 page if no examples user

francisli:

Thank you!

If there is no p5 examples user, the server will crash hard when serving up a 404. This change will serve up the page without an embedded sketch. Also, it fixes the generated HTML so that the inserted message is inside the <body></body> tags.


catarak:
i think this points at a bigger issue which is that it would be cool if the examples were handled in a different way from regular users.

i'm okay with merging this since it prevents the editor from crashing, but i wonder if there should be an error (client side?) that's like, hey, the database hasn't been initialized!

francisli:
If you want to write up some copy I can put it into the no-user error page...

Otherwise, the reason I encountered this is because I opted not to install all the default examples for our deployment- I didn't want our curious third and fourth graders distracting themselves by finding and playing around with all the cool examples during our sessions...

francisli:
@catarak I've just rebased/squashed this branch onto the latest from master, is there anything else you'd like me to do here?

catarak:
Rather than just inserting an empty HTML page, could the 404 page just be rendered without the example embedded in the bottom?

francisli:
@catarak sorry for not being clear, but that's exactly what this already does- it refactors the existing uses the existing page

francisli:
@catarak sorry, hit enter too quickly- it refactors the existing 404 code so that a) the HTML is valid and b) it doesn't add an example into it. It doesn't just render a blank page, you see the message and styling above it.

catarak:
this is working great. just a thought—would it also make sense to remove the "examples" link in the navigation? is that not important for you? the only thing that's slightly annoying is that if `/p5/sketches` is rendered client-size, which you would get to from clicking "examples" in the nav, then it renders the sketch list view with no sketches, but if you refresh the page, it shows a 404 error, which is inconsistent.

francisli:
Pushed a fix for the bug you just reported...

catarak:
the `userExists` function should catch if the p5 user is not defined, so i don't think the change you made actually changes anything. i'm talking about when you open the web editor, and then click file -> examples, which renders an empty list of sketches.

francisli:
@catarak yes, I understand the bug, and this patch _does_ fix it- when the userExists function returns false, the additional boolean expression will override that if the user is 'p5' so that the index page will still be returned instead of the 404. Let me fix the linter error first, though...

catarak:
my bad, i see what you're saying! 

this is, perhaps, a small issue, but this _should_ return a 404, since the user doesn't exist, rather than returning a list of empty sketches, which implies the user exists but has no sketches. and it's also inconsistent—`/<non-existent-user>/sketches` renders a 404.

francisli:
@catarak this fix is only for the special-case p5 user (/p5/sketches), which is used for the Examples list dialog. If you don't have that user, it will show an empty list when you select Examples from the menu, which I think is fine.

Since the p5 user is special (and hardcoded all throughout the codebase), I think it is fine to handle it differently here per your original comment that refreshing the window after you've opened the Examples dialog from the menu will not show the same result...?

Any other non-existent user will still return a 404, as is expected.

francisli:
In other words, there is no "consistent" solution here because the p5 user is already a special-case user that is handled differently. So we need to do the same here if you don't want the browser to return 404 when you refresh after opening the Examples dialog (or navigate to it directly)...

francisli:
@catarak I've reverted the "fix" for handling the p5 user and added another environment variable for enabling/disabling support for showing Examples in the File menu...

catarak:
okay! i think this is a better solution, at least in the short term.

PR #778, opened 2018-12-10
Disable login/save

francisli:

Thank you!
---------
This PR introduces a new environment variable LOGIN_ENABLED, which defaults to TRUE if not defined. Setting LOGIN_ENABLED=false will remove the Log in/Sign up links from the nav bar and the Save option from the drop-down menu.

The reason I've implemented this is because we're using the editor with elementary school students, and they've already started poking around and creating accounts. However, we don't want to deal with COPPA (Children's Online Privacy Protection Act) issues right now, so I'm opting to just hide/disable this...


catarak:
Thanks for this! We have discussed making the web editor COPPA/FERPA compliant, though I think that it would be a few months out.

Issue #806, opened 2019-01-15
Web Editor welcome experience


daaain:
#### Nature of issue?

- New feature request

#### Feature enhancement details:

I found P5.js via a Mozilla blog post: https://blog.mozilla.org/blog/2019/01/03/moss-2018-year-in-review/

In which it's described as:

> The Processing Foundation maintains p5.js, an open-source JavaScript framework that makes creating visual media with code on the web accessible to anyone, especially those without traditional computer science backgrounds. p5.js enables users to quickly prototype interactive applications, data visualizations, and narrative experiences, and share them easily on the web.

So I opened https://p5js.org/ and after reading the intro clicked on the link on the bottom to try the Web Editor.

What I was expecting is maybe some tutorial but definitely something a visual way of working, but what it ended up seeing is a minimal example and not much idea what to do:

<img width="1127" alt="image" src="https://user-images.githubusercontent.com/69962/51213680-7fc1ef80-1913-11e9-9fb4-f3785da7f9db.png">

So what I'm proposing is creating some sort of welcome / beginner experience, be it a modal window with a short video embedded, an interactive tutorial, or a (moving) visual library of examples, but please make it not be just a blank grey background 🙏

catarak:
@daaain i think this is a great idea! the web editor has mostly been used in conjunction with curricula/workshops/classes/etc. which has served this purpose. i made a [video](https://www.youtube.com/watch?v=x1rJJRVTpAI) that walks through the basic features, and i know @shiffman has made a few one-off shorter videos that show off features of the editor as well. 

if you have ideas/reference please add them here!

LakshSingla:
I totally second this part. I also completely missed that arrow. 
> I totally missed the power behind the > left of sketch.js above the line numbers.
> I thought i was just some kind of graphical highlighter that had no function.
> Maybe it should be a different colour and have better tip text.
> ![screenshot_7](https://user-images.githubusercontent.com/121805/51218809-e1756e80-1992-11e9-8422-8203f602e992.jpg)

shiffman:
I agree with all of the thoughts above! A model for this might be the Processing "Getting Started" tool created by @jaewhyun as part of last year's Google Summer of Code. For more, you can read her medium post [Making the Processing Development Environment Better for Beginners](https://medium.com/processing-foundation/making-the-processing-development-environment-better-for-beginners-a43302952f18). 

[...]

PR #815, opened 2019-01-22
Changes to SVG images & added titles to images for tooltips & css +contrast

hellonearthis:
The changes make to the SVG files are mostly removing the TITLE as it over road the tooltips when hovering over a graphic. ALT text doesn't work with SVG so I added TITLE= to the buttons connected with these inline SVG images so I friendly tooltip is given.  #806 
I changed the left-arrow.svg and right-arrow.svg to give a better representation of its function.
screenshot_3

I increased the high contrast css to boost the contrast of the line numbers, it's know white.

The docker environment has issues with the popup windows not loading images or sounds and that wasn't me :D but I hope it's just a docker build issue.

I'm new to working with repos, so I hope i did the whole branch pull thing right.


catarak:
thanks for starting to work on this! i have a few thoughts:
1. this pull request is for three separate issues: (1) adding tooltips, (2) icon change suggestions, (3) changing colors in the high contrast theme. these are unrelated issues and should be three separate pull requests.
2. there are unnecessary changes to `package.lock` and line ending changes that should be removed.
3. it's inconsistent how the tooltips are being added. sometimes they are added to the SVGs, sometimes to the span that's holding the svg, sometimes to a button. where is it important that they be added? 
4. i don't think the titles should be directly added to the SVG files. what if you want to use an SVG in multiple places? since the inline SVG library this project is using doesn't allow you to change the SVG attributes, i'm thinking maybe it should change to a different one, [react-svg-inline](https://github.com/MoOx/react-svg-inline), which allows you to remove all props and specify accessibility ones.

hellonearthis:
I will put the icon change and high contrast into to other branches and restore those to master version along with the package.lock changed, so just the tool tips will be here.

I stated added the <title> to the buttons but discovered that some areas that needed a tool top didn't have a button to add it too.  Or that adding it to the button isn't sufficient as the space spanning the area needed a tooltip.  Going to review how I did them again and see if there is a consistent way to add these.

Yes, adding titles to SVG gets messy fast when it comes to updates if tol tips are hidden in the image files. 
SVGO is used by webpack to minify* the SVG images,  it has options to remove <title> <desc> from SVG files at build time.  So that would be the easiest way to fix the current SVG files.

catarak:
feel free to split up this PR as well if that's easier—maybe make one that just updates the webpack svg library, and then one that changes the tooltips.

PR #817, opened 2019-01-25
[WIP] Add friendly errors pref, fixes #417

### LINKED ISSUE #417: friendly error system preferences item
@almchung has implemented the "friendly error system" in the p5.js, that is, functionality that will (optionally) check the parameters passed into p5 functions when they're called and send helpful messages to console. we're about to merge it this week. this feature can be toggled on and off by setting a boolean variable (`p5.disableFriendlyErrors = true/false;`) in the sketch. is it possible to add this as a preferences item that can be toggled from there, too? 

if it sounds ok, I can work on implementing this.

catarak:


catarak:
this issue is deceptively hard to fix—upon revisiting this, i'm not totally happy with this approach. if a user is open to another person's sketch, it would overwrite their index.html, which i don't think is a good idea. i'll have to think about this more...

PR #834, opened 2019-02-16
Prevent crash when console.error is used - Fixes #833

### LINKED ISSUE #833: Page Crashes Under certain console.error situations
#### Nature of issue?
- Found a bug

#### Details about the bug: 

Any call to `console.error` where the first argument is not a string will cause a crash.
This includes all of the following:
`console.error()`
`console.error({})` // any object that doesn't support `indexOf`
`console.error({}, 'A string here')`

In particular, some parts of p5 will call `console.error` with an Error object (or sometimes generic other objects) as the first argument.

This is a more generic version of one of the problems in #831

meiamsome:


catarak:
awesome, thanks for this!

Issue #790, opened 2018-12-17
Save/refresh issue


nanaosakisan:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Details about the bug: 

- Web browser and version: Firefox Version 64.0 (and the lastest version of Chrome)
- Operating System: MacOs Sierra Version 10.12.6
- Steps to reproduce this: don't save/refresh code's modifications when I try to run the program.

catarak:
thanks for reporting @nanaosakisan! can you be more specific about how to reproduce this bug? for example, by listing out a few steps? i don't understand what the bug is.

nanaosakisan:
I can't tell you step by step how to reproduce the bug. If I change my code (like add a variable named "test" for example), and save the change, when I want to run the program, an error notify appears and tells me that the variable "tes" doesn't exist. If I refresh the page, I can see that the variable I created before is named "tes", so not correctly.

If I don't save the change, the bug doesn't appear.

Télécharger Outlook pour Android<https://aka.ms/ghei36>



De : Cassie Tarakajian
Envoyé : lundi 17 décembre à 20:08
Objet : Re: [processing/p5.js-web-editor] Save/refresh issue (#790)
À : processing/p5.js-web-editor
Cc : nanaosakisan, Mention


thanks for reporting @nanaosakisan<https://github.com/nanaosakisan>! can you be more specific about how to reproduce this bug? for example, by listing out a few steps? i don't understand what the bug is.

shiffman:
I have noticed some odd behavior like this after the fix for #675 was published. I think this may be related? I meant to file this a few days ago, but I forgot. You can see the trouble I was having here:

https://youtu.be/pKyU92cza0Y?t=1978

I will investigate a bit more!

shiffman:
I encountered this bug again this morning. I still cannot determine what exactly is causing it and I don't see anything in the console output. But in case this is helpful,  here is the sketch:

https://editor.p5js.org/codingtrain/sketches/SJMl3u5xN

This is what  I am doing:

1. Editing lines 35 and 38 to switch `add()` to `push()`.
2. Manually saving with command-s.
3. Run sketch.
4. Error: `Uncaught TypeError: next.add is not a function`
5. Refresh page (no warning) and code is back to `add()`.

shiffman:
(Update I restarted browser and was able to save the change)

catarak:
[...]

PR #838, opened 2019-02-20
Prevent Race Condition when Updating Project - Fixes #837

### LINKED ISSUE #837: Race condition when updating a project
#### Nature of issue?
- Found a bug

#### Details about the bug: 
Whilst looking through the code for other things, I discovered there is multiple `res.json` calls when a project is updated to have a different amount of files. In particular,
https://github.com/processing/p5.js-web-editor/blob/master/server/controllers/project.controller.js#L80 should be in an else clause.

I wasn't able to cause this to exhibit any weird behaviour locally, but I suspect it could be buggy when the mongodb server is remote. I am not sure what affect this might have

meiamsome:

### DIFF
--- a/server/controllers/project.controller.js
+++ b/server/controllers/project.controller.js
@@ -76,8 +76,9 @@ export function updateProject(req, res) {
             }
             res.json(savedProject);
           });
+        } else {
+          res.json(updatedProject);
         }
-        res.json(updatedProject);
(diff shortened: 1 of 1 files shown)


catarak:
thanks! maybe this fixes (or at least partially fixes) #790 too?

PR #844, opened 2019-02-24
Use File id in Redux Actions - Fixes #842

### LINKED ISSUE #842: Multiple Files of the Same Name in folders
#### Nature of issue?
- Found a bug

#### Details about the bug: 
When there are two files with the same name but in different folders, the frontend UI can get confused and overwrite the wrong file contents.

1. Create a new project
2. Add a folder
3. Create file 'style.css' inside the folder, and add some contents and save
4. The contets of 'style.css' in the root folder are the same as those you just created in the subfolder

meiamsome:
Prevents file name collisions that are possible with the folder
system.

### DIFF
--- a/client/modules/IDE/actions/files.js
+++ b/client/modules/IDE/actions/files.js
@@ -29,10 +29,10 @@ function createUniqueName(name, parentId, files) {
   return testName;
 }
 
-export function updateFileContent(name, content) {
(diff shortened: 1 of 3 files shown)


catarak:
yay thanks for fixing this! this is working great for me.

Issue #836, opened 2019-02-19
add soft wrap for long text 


bmoren:
#### Nature of issue?

- New feature request

#### Feature enhancement details:

in the examples, and when writing examples for my classes, there tends to be long descriptive comments. It'd be amazing to have the ability to toggle softwrap or even have a setting for it in the editor so these comments can automatically remain within the editor window and prevent excess scrolling.

perhaps we could implement the wrap/hardwrap.js addon for code mirror to accomplish this?
https://codemirror.net/demo/hardwrap.html

I'd prefer that the line numbers are incremented with the soft wrap in account but this would still be a welcome addition: eg: 
![screen shot 2019-02-19 at 11 38 50 am](https://user-images.githubusercontent.com/1385996/53035408-f5513c00-343a-11e9-8ffa-f4e9249ca301.png)

#### New feature details:

bmoren:
comment/continuecomment.js

comtinue Comment addon could be an alternative  to achieve a similar ease of commenting

bmoren:
This is maybe also as easy as adding a toggle to the settings for lineWrapping here? : https://github.com/processing/p5.js-web-editor/blob/master/client/modules/IDE/components/Editor.jsx#L82

catarak:
i, personally, prefer not to have soft wrapping when writing code, as i find it confusing to look at. i see the value of it for comments, and i think perhaps the best approach would be the codemirror continuecomment addon, or a different codemirror addon that i don't know about (it seems there are a lot of these).

bmoren:
I agree that having softwrap, especially for functions with a lot of parameters, gets really confusing. 
eg.
```
ley x = map(mygreatvalue, startingVariable, 
endingVariable, 1,1000); //very confusing!
````


It seems like the continue comment addon would be the ticket. If it's a welcome thing to try I can maybe have a go at it this weekend and see if it's a good behavior

catarak:
go ahead and play around with it! in my experience, including codemirror plugins isn't super hard, it's just a matter of including the file and then updating the CodeMirror configuration when you instantiate it.

bmoren:
can someone point me to the places I'll need to work in order to add a feature in the settings.  I'm new with react, and there's a lot going on here!

LakshSingla:
[...]

Issue #263, opened 2017-01-11
Add closing curly brace automatically


avocadoboi:
(^:

catarak:
This issue already exists as #55. It's still open for debate, but for now we've decided that it's more confusing to beginners. Another option would be to add it as a preference.

PR #859, opened 2019-03-01
adding in continue comment codemirror addon adresses #836

bmoren:
I'm not sure that this behavior is actually a good idea, it's a bit tricky to get out of the continuation of the comments, it might be better to add a real soft wrap option into the settings or edit menu.

* [?] is from a uniquely-named feature branch and has been rebased on top of the latest master. (If I was asked to make more changes, I have made sure to rebase onto master then too) <-- I think?

### DIFF
--- a/client/modules/IDE/components/Editor.jsx
+++ b/client/modules/IDE/components/Editor.jsx
@@ -14,12 +14,14 @@ import 'codemirror/addon/fold/foldcode';
 import 'codemirror/addon/fold/foldgutter';
 import 'codemirror/addon/fold/indent-fold';
 import 'codemirror/addon/comment/comment';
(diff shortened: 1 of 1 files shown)


catarak:
i agree, i can't tell if this is helpful or annoying. i feel like this is similar to the discussion around #263, in which it seems kinda confusing for new programmers.

bmoren:
agreed, I feel like it is likely pretty confusing to have to click or shift+enter to get out of the comment block. In light of this, I suppose my preference would be to have the soft wrap option in the settings, but to have it disabled by default.

bmoren:
going to close this, since It seems like a bad idea in comparison to some other options

Issue #865, opened 2019-03-04
resizing console only comes into effect after re-engaging with the code editor window


bmoren:
#### Nature of issue?

The console must be tied into the update of the codemirror editor. when resizing the console, it will only accept the resize when re-engaging with the editor. 

- Found a bug

#### Details about the bug: 

- Web browser and version:  FF 65.0.1 (64-bit)
- Operating System:  MAC OS 10.13.4
- Steps to reproduce this bug: 
1. drag up the console to a new size
1. it will snap back to its original position
1. make a line return in the code editor
1. the console will snap to the position you chose previously.

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/1385996/53742568-8de5b400-3e5e-11e9-8c5a-65f958981a4e.gif)

catarak:
i've noticed this, but haven't figured out steps to reproduce, so thanks for this!

siddhant1:
Will investigate and fix !

PR #866, opened 2019-03-05
Fixed Console Pane Not Resizing (#865)

siddhant1:

### DIFF
--- a/client/modules/IDE/pages/IDEView.jsx
+++ b/client/modules/IDE/pages/IDEView.jsx
@@ -138,8 +138,8 @@ class IDEView extends React.Component {
   _handleConsolePaneOnDragFinished() {
     this.consoleSize = this.consolePane.state.draggedSize;
     this.consolePane.setState({
-      resized: false,
-      draggedSize: undefined,
+      resized: true,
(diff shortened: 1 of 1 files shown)


siddhant1:
Fixes #865

siddhant1:
I think the changes are based on the fact that how this code https://github.com/tomkp/react-split-pane/blob/master/src/SplitPane.js works,

So in this code it check if we have the size prop coming and if yes it assigns the value to the resizer , if not it calculates by a method named geDefaultSize what is the next value .

In our code , we are passing in default value and value of ref , now it calculates the new value of resizer using the getDefaultSize method and it sets it , now in the next line we set dragged size to undefined which again runs getDefaultState and now `draggedSize` is undefined so it returns minValue.

An easy fix to this solution would be not passing `draggedSize` value undefined which I have done now .
Another fix could be to pass size and not defaultSize because `getDefaultSIze` checks for this value first.

@catarak can you suggest which implementation is better?

catarak:
thanks for making this PR! i took a look at this, and then realized that this code needed to be cleaned up, so i made #869. i'm gonna merge that one in, but thanks for your help on this issue.

Issue #597, opened 2018-03-08
Navigation dropdown menus are not accessible by keyboard


catarak:
#### Nature of issue?

- [X] Found a bug

#### Details about the bug: 

- Web browser and version:  Chrome
- Operating System:  Mac OSX
- Steps to reproduce this:

1. Hit `tab` to focus on one of the navigation menu options (e.g. 'File')
2. Hit `enter` to open the dropdown
3. Hit `tab` again to cycle through the options of the dropdown, except the dropdown closes. 

I'm pretty sure at one point this worked, as I spend a lot of time researching how to make dropdowns keyboard accessible and perfecting it.

himanshuc3:
It's working in my local environment! Reproduced it as per your instructions @catarak.

catarak:
hmm, as in you're able to tab through and select "new", "open", etc.?

himanshuc3:
No, i am able to tab from between menu items but not within sub-menu items.

catarak:
Okay, same for me. The issue is not being able to tab into submenu items.

Neilshweky:
I would like to tackle this issue, or a similar issue that pressing `tab` cycles through all of the different tabs

catarak:
go for it!

Neilshweky:
I submitted a pull request! Let me know what you think!

himanshuc3:
Is this solved @Neilshweky @catarak

Neilshweky:
Hi! I'm sorry I haven't forgotten about this I just have so much work right now (as the semester is nearly over).... I definitely should be able to work on it this week though, I'll keep you posted! Thanks!

catarak:
no worries ✨

migom6:
what's the status?

Neilshweky:
Hey! I'm looking into other ways of fixing this issue!

dhruvdutt:
@catarak This can be closed I guess.

PR #876, opened 2019-03-09
Remove file menu on blur

### LINKED ISSUE #874: Multiple file options
#### Nature of issue?

- Found a bug

#### Details about the bug: 

- Web browser and version: Chrome 70.0.3538.77 (Official Build) (64-bit)
- Operating System: Ubuntu 18.04.1 LTS
    - Right click on mutiple files

On losing focus file menu needs to be closed. Otherwise it looks ugly when no option is selected from file menu and another file is right clicked.

screenshot from 2019-03-08 22-44-30

nimishagarwal76:

* [x] is descriptively named and links to an issue number, i.e. `Fixes #874`

Fixes #874 

This PR resolves the issue of multiple file menu as on blurring, they will be hidden.
But this work in all cases except when we right click on the current file as it automatically changes the focus to the ```˅``` resulting in losing of focus from the file name and disappearing of menu. If you right click on the ```˅``` then it would work properly.

Screenshot from 2019-03-09 12-55-56

@catarak why is a down arrow required specifically for the current file, whereas it is not availaible for other files.

Doesn't work if person uses TAB to navigate as it loses Focus. I will find another approach.

### DIFF
--- a/client/modules/IDE/components/FileNode.jsx
+++ b/client/modules/IDE/components/FileNode.jsx
@@ -154,6 +154,7 @@ export class FileNode extends React.Component {
                   tabIndex="0"
                   onClick={this.toggleFileOptions}
                   onBlur={() => setTimeout(this.hideFileOptions, 200)}
+                  style={{ display: (this.state.isOptionsOpen) ? 'inline-block' : '' }}
                 >
                   <InlineSVG src={downArrowUrl} />
                 </button>
--- a/client/styles/components/_sidebar.scss
+++ b/client/styles/components/_sidebar.scss
@@ -125,7 +125,6 @@
   @include icon();
   @include themify() {
     padding: #{4 / $base-font-size}rem 0;
-    background-color: map-get($theme-map, 'file-selected-color');
     padding-right: #{6 / $base-font-size}rem;
   }
   display: none;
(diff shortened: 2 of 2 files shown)


nimishagarwal76:
@catarak  the issue is resolved, Regarding not able to navigate by TAB #597 deals with it. While solving this I found that each file node should have a ```˅``` arrow but it is only visible for current file (I think its a bug. Do correct me if its not).
 So I made ```˅``` arrow display as ```none``` cause upon clicking the name focus went on the arrow and onBlur event use to trigger.
Can you please test it and tell your views regarding it.

catarak:
the selected file _should_ show a caret, which, when testing your pr, i noticed that it doesn't anymore.

nimishagarwal76:
Yeah I wanted to know that each Filenode has a caret but it gets displayed only for currently selected file node (before my PR) . Is it a bug or some condition is set for displaying it only for current file.

> the selected file _should_ show a caret, which, when testing your pr, i noticed that it doesn't anymore.

Yeah currently I set its visibility to none.

hellonearthis:
If there is no caret then how do you rename or delete a file.  Having is show on the currently selected file is good.

catarak:
no, it's on purpose! i think it makes sense to only show it for the selected file for ui-clutter reasons. the inspiration was the glitch.com editor, and i noticed that they also show the caret when hovering over a file, which could also be nice.

nimishagarwal76:
@catarak I fixed it. I also implemented the caret upon hovering. Please test it and do let me know if something else is required.

catarak:
@nimishagarwal76 i ended up making another fix, so i'm gonna merge this in with my fix!

Issue #853, opened 2019-03-01
Remove redundant console.logs


siddhant1:
I think we should remove console.logs from the code 

Links :- https://dev.to/mornir/-dont-leave-console-logs-in-production-14na
https://hackernoon.com/please-stop-using-console-log-its-broken-b5d7d396cf15

I can send in a PR 

Have a look @catarak

catarak:
totally! it would be amazing to add a logging library. maybe this ticket could be that—investigate and add logging library?

siddhant1:
Yeah , I will look into it and start refactoring

siddhant1:
As I was looking through the code , I found out that most of the console statements were used for debugging and some for handling errors.
There are 2 ways to solve this.
1. Remove console.logs used for debugging
2.find a way to disable console.logs on the prod mode and enable in dev mode which agains boils down to adding a logging library

catarak:
sounds great to me!

siddhant1:
@catarak I was thinking of a solution to this ticket.
Should I monkey patch window.console.log which checks if we are in dev mode or prod mode and logs only in dev mode?

catarak:
`process.env.NODE_ENV` is already passed to the front end (e.g. in [App.jsx](https://github.com/processing/p5.js-web-editor/blob/master/client/modules/App/App.jsx)), so removing console logs when in prod should be pretty simple!

GaurangTandon:
@siddhant1 Could I please work on this issue while you're working on the other ones? Thank you!

siddhant1:
Hey Gaurang , this issue was already solved and needs to be closed. You can work on other ones :)

catarak:
@siddhant1 / @GaurangTandon this issue hasn't been solved. no logging library has been added.

siddhant1:
The issue was originally to remove logs which was done , for adding logging library , I will send a pr soon

sagar-joshi:
I am working on this

vivekbopaliya:
@catarak 
![Screenshot (177)](https://github.com/processing/p5.js-web-editor/assets/122638553/febba2d2-f313-4de3-900b-750da01bcb58)
 
as there are bunch of console logs for debugging only, we can remove them---and add Winston logging in relevant msgs only

PR #878, opened 2019-03-10
Patched console.log to only log in dev mode (#853)

siddhant1:

I don't think this is the mode elegant solution , but we can add another logger that logs with time and everything.
Also there is a need to remove console.logs from error callbacks and actually handle errors.
I will look into it next

### DIFF
--- a/client/index.jsx
+++ b/client/index.jsx
@@ -6,6 +6,8 @@ import { Router, browserHistory } from 'react-router';
 import configureStore from './store';
 import routes from './routes';
 
+// Load the patched console.log
+require('./utils/logger.js');
 require('./styles/main.scss');
 
 // Load the p5 png logo, so that webpack will use it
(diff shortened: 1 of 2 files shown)


siddhant1:
Also I feel the errors are handled at the backend but there are many logs too.
I think we should remove those logs and actually log into a file (winston or morgan) or a datastore.
That'll be a great improvement.

catarak:
rather than overwriting `console.log` in production, is there a way with webpack to just remove all console.log statements from the javascript? i think that is a better solution.

siddhant1:
https://www.npmjs.com/package/babel-plugin-transform-remove-console

I think terser and uglifyJs also do it , but I found this babel plugin which is handy too .

@dhruvdatt @catarak

siddhant1:
The console.logs were not removed from the bundle and I opened a new pr with a better approach( #884 ) , babel-plugin will remove logs from the bundle in the prod.
Closing this for now/

PR #880, opened 2019-03-11
Added Loader to indicate loading of sketches

siddhant1:


catarak:
is there a reason you put the loader in a separate reducer? i think it makes sense to put a flag in the projects reducer.

client/styles/components/_loader.scss
```diff
@@ -0,0 +1,36 @@
+.lds-ring {
```
catarak: also, the scss should use BEM and should not use pixel measurements (should use rem's).

siddhant1:
I think putting the loader in a different reducer will enable to start loading from any component , otherwise it will be bound to the projects

siddhant1:
Changed to rem , also I have changed the loader to a single div , is there any need of bem syntax now?

client/styles/components/_loader.scss
```diff
@@ -0,0 +1,46 @@
+$color_1: #000000;
```
catarak: you should load the color from the appropriate theme, see `client/styles/abstracts/_variables.scss`, and use the SASS function `getThemifyVariable`

client/styles/components/_loader.scss
```diff
@@ -0,0 +1,46 @@
...
+	background: #000000;
+	animation: load1 1s infinite ease-in-out;
+	width: 1rem;
+	height: 4rem;
+	color: $color_1;
+	margin: 8rem auto;
+	position: relative;
+	font-size: 11px;
+	transform: translateZ(0);
+	-webkit-animation-delay: -0.16s;
```
catarak: webpack has an autoprefixer so you shouldn't need the prefix here

client/styles/components/_loader.scss
```diff
@@ -0,0 +1,46 @@
...
+	}
+	40% {
+		box-shadow: 0 -2rem;
+		height: 5rem;
+	}
+}
+.loader {
+	background: #000000;
+	animation: load1 1s infinite ease-in-out;
+	width: 1rem;
```
catarak: when using `rem`s, calculate them by doing `#{16 / $base-font-size}rem`, so that it's easy to still see what the pixel measurement would be. this is for readability.

siddhant1: Should I put them as a comment alongside?

catarak:
i hear you about the loader being able to be used for different API requests. i was doing some research about how other projects use loaders with redux, and i stumbled across this neat [loader reducer example](https://medium.com/stashaway-engineering/react-redux-tips-better-way-to-handle-loading-flags-in-your-reducers-afda42a804c6). i think, however, for now, that example is more complicated than this project needs, and we can leave the reducer you wrote as it is now. just thinking out loud here 😸

siddhant1:
Also I was thinking to add an avg loader instead of this , what do you think?

catarak:
i don't have a preference whether the loader is an svg or done in css. whatever is easier, just as long as you can change its color with respect to the theme. i'm not crazy about the loader's design, as the color is not right (it should be `logo-color`!), and the rectangles don't feel right. i'm thinking something circular, maybe something like https://editor.p5js.org/cassie/sketches/QGPxNuA9z?

siddhant1:
Looks good , I think I need to do that with scss

siddhant1:
http://tobiasahlin.com/spinkit/ (2nd and 4th)
These look kinda similar to me , what do you think about them?

catarak:
i like the second loader there!

client/modules/IDE/actions/projects.js
```diff
@@ -8,6 +8,7 @@ const ROOT_URL = __process.env.API_URL;
 
 export function getProjects(username) {
   return (dispatch) => {
+    dispatch({ type: ActionTypes.START_LOADING });
```
catarak: rather than putting the action payload inline, this should call a function. same with the other times dispatch is called.

client/modules/IDE/components/SketchList.jsx
```diff
@@ -32,7 +33,8 @@ class SketchList extends React.Component {
         <Helmet>
           <title>{this.getSketchesTitle()}</title>
         </Helmet>
-        { this.props.sketches.length === 0 &&
+        {this.props.loading && !this.props.sketches.length && <Loader />}
```
catarak: can this logic be moved into a function so it's easier to read?

client/styles/components/_loader.scss
```diff
@@ -0,0 +1,43 @@
...
+	@include themify() {
+		background-color: getThemifyVariable('logo-color');
+	}
+
+	opacity: 0.6;
+	position: absolute;
+	top: 0;
+	left: 0;
+
+	-webkit-animation: sk-bounce 2.0s infinite ease-in-out;
```
catarak: you shouldn't need this prefix

client/styles/components/_loader.scss
```diff
@@ -0,0 +1,43 @@
...
+	animation: sk-bounce 2.0s infinite ease-in-out;
+}
+
+.loader__circle2 {
+	-webkit-animation-delay: -1.0s;
```
catarak: or this prefix either!

siddhant1:
Already made changes with the 2nd one

client/utils/loader.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
```
catarak: i think this file should be in client/modules/App/components

catarak:
just saw that 😄 requested some changes!

siddhant1:
Working on em :)

client/modules/IDE/actions/projects.js
siddhant1: Something like startLoading and stopLoading ? Also should I put these functions in a separate file?

client/modules/IDE/components/SketchList.jsx
```diff
@@ -26,14 +26,18 @@ class SketchList extends React.Component {
     return `p5.js Web Editor | ${this.props.username}'s sketches`;
   }
 
+  startLoader() {
```
catarak: could you call this function `renderLoader`?

siddhant1:
Not sure if this is intentional , but i think we should clear the list when the close the SketchList modal, its very confusing because on slow internet if I open examples and then open my sketches , examples show up first and then my sketches show up.
According to me a better UI would be clearing and showing a loader.
A benefit they might give is if the user is accessing examples (or his sketches) again and again but for exmaple I add a sketch , previous sketches show up first and then the new sketch shows.

client/modules/IDE/actions/projects.js
catarak: yes to both questions! every reducer should have its own actions/reducer file.

siddhant1:
I'm not sure if we want this implementation , we are calling start and stopLoader from another action creator , and not from a component so we can't use bindActionCreators.
Are you thihnking of any other approach?

siddhant1:
image
I'm also not sure if we should do this , this is basically the same but the dispatch method is moved to another file.
In current implementation , we make action creators and dispatch from other action creator

client/modules/IDE/components/SketchList.jsx
```diff
@@ -25,14 +26,19 @@ class SketchList extends React.Component {
...
   render() {
     const username = this.props.username !== undefined ? this.props.username : this.props.user.username;
     return (
       <div className="sketches-table-container">
         <Helmet>
           <title>{this.getSketchesTitle()}</title>
         </Helmet>
-        { this.props.sketches.length === 0 &&
+        {this.renderLoader()}
+        { !this.props.loading && this.props.sketches.length === 0 &&
```
catarak: last small thing—could you turn this into another helper method, which you could also call in `renderLoader()`?

siddhant1:
https://github.com/processing/p5.js-web-editor/pull/880#issuecomment-477067060

@catarak can you also look at this

catarak:
> Not sure if this is intentional , but i think we should clear the list when the close the SketchList modal, its very confusing because on slow internet if I open examples and then open my sketches , examples show up first and then my sketches show up.
> According to me a better UI would be clearing and showing a loader.
> A benefit they might give is if the user is accessing examples (or his sketches) again and again but for exmaple I add a sketch , previous sketches show up first and then the new sketch shows.

i agree! it's confusing behavior. i think what should actually happen is that the list of sketches passed to the sketchlist should be a selector.

siddhant1:
What I have read about selectors (I am very new to them) is that selectors are a temporary UI of the original lists , in this case we are making an ajax request based on the url , can you explain how does it solves the problem?

catarak:
ahh i think you are right! i was thinking that because the ajax request would be the same, with a different username, so this could function like a selector, but the ajax part means it can't be a selector. i think, rather than deleting the list sketches, you could just render the loader if `this.props.loading` is true, and not worry about the length of the sketches.

siddhant1:
Yes ! This feels like a good solution , will push changes today

siddhant1:
@catarak I think this one is complete now!

client/modules/IDE/components/SketchList.jsx
```diff
@@ -25,17 +26,21 @@ class SketchList extends React.Component {
...
       <div className="sketches-table-container">
         <Helmet>
           <title>{this.getSketchesTitle()}</title>
         </Helmet>
-        { this.props.sketches.length === 0 &&
-          <p className="sketches-table__empty">No sketches.</p>
-        }
-        { this.props.sketches.length > 0 &&
+        {this.renderLoader()}
+        {!this.props.loading && this.props.sketches.length > 0 &&
```
catarak: could you turn `!this.props.loading && this.props.sketches.length > 0` into a helper function?

siddhant1: Should I do something like arePages() && <JSX/>

siddhant1: I am sorry for delay , I will finish this today.

siddhant1:
@catarak  Sorry for the delay, had some exams going on! So was unable to code for some time.
Will close this and other prs in few days

catarak:
let me know if you need help with the remaining changes on this!

siddhant1:
Sure!

catarak:
cool, i added another commit on top of this, and just merged ✨

PR #884, opened 2019-03-13
Patched console.logs to work only in dev mode (#853)

siddhant1:

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,8 @@
       "plugins": [
         "transform-react-remove-prop-types",
         "transform-react-constant-elements",
-        "transform-react-inline-elements"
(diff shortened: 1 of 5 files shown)


siddhant1:
@catarak , the package-lock will again conflict :p , maybe we can setup .nvmrc for consistent node version across the project

catarak:
thanks for working on this! it seems like you can remove console.* statements using uglify, which is already installed. so why not do that instead of adding another dependency?

siddhant1:
Yes , I think we can also do it with webpack and uglifyJs and since we are already using uglifyJs plugin that'd be a better approach

siddhant1:
Moved to #890

PR #887, opened 2019-03-14
Added hover effect on sidebar elements

### LINKED ISSUE #886: Hover effect on sidebar elements
#### Nature of issue?

#### Details about the bug: 

- Operating System:  Windows

add hover effect on side bar elements
Capture

Earlier there was no hover effect on sidebar elements like "remane" or "delete"

plxity:
fixes #886

### DIFF
--- a/client/modules/IDE/components/FileNode.jsx
+++ b/client/modules/IDE/components/FileNode.jsx
@@ -203,7 +203,7 @@ export class FileNode extends React.Component {
                     {(() => { // eslint-disable-line
                       if (this.props.fileType === 'folder') {
                         return (
-                          <li className="sidebar__project-option">
+                          <li>
                             <button
                               aria-label="add folder"
                               onClick={() => {
@@ -220,7 +220,7 @@ export class FileNode extends React.Component {
                         );
                       }
                     })()}
-                    <li className="sidebar__project-option">
+                    <li>
                       <button
                         onClick={() => {
                           this.originalFileName = this.props.name;
@@ -235,7 +235,7 @@ export class FileNode extends React.Component {
                         Rename
                       </button>
                     </li>
-                    <li className="sidebar__project-option">
+                    <li>
                       <button
                         onClick={() => {
                           if (window.confirm(`Are you sure you want to delete ${this.props.name}?`)) {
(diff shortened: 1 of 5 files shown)


client/modules/IDE/components/Sidebar.jsx
```diff
@@ -69,12 +69,12 @@ class Sidebar extends React.Component {
               <InlineSVG src={downArrowUrl} />
             </button>
             <ul className="sidebar__project-options">
-              <li>
+              <li className="sidebar_file">
```
catarak: following [BEM](http://getbem.com/) conventions, i would name this class "sidebar__project-option"

client/modules/IDE/components/FileNode.jsx
```diff
@@ -162,7 +162,7 @@ export class FileNode extends React.Component {
                     {(() => { // eslint-disable-line
                       if (this.props.fileType === 'folder') {
                         return (
-                          <li>
+                          <li className="sidebar_file">
```
catarak: following BEM conventions, i would name this class "sidebar__file-item-option"

client/styles/components/_sidebar.scss
```diff
@@ -286,8 +285,42 @@
   background-color: transparent;
   border: none;
 }
+.sidebar_file{
```
catarak: since this is using the same code as the Nav, maybe this should be moved into a SASS placeholder or mixin?

plxity:
@catarak I have made the following changes. Please review it again.

catarak:
could you remove changes to `package-lock.json`? also, you didn't move the styles to a mixin or placeholder even though they are duplicate styles.

plxity:
@catarak Okay, I will remove changes from `package-lock.json`.
Actually, There is a difference in the styles, I have removed a line or two to make it look proper.

catarak:
you can still use a placeholder or mixin! you can extend it and then override those styles.

plxity:
@catarak Okay, I will fix that.

plxity:
@catarak  I have used a different approach now. and have implemented the changes. Can you please review this? :)
Thank you.

catarak:
here's some stuff i'm noticing: 
Screen Shot 2019-04-17 at 2 49 26 PM
the highlight doesn't reach the full width/height of each item
Screen Shot 2019-04-17 at 2 49 37 PM
it looks a little weird without the border radius. i think the style should match the navigation dropdown
Screen Shot 2019-04-17 at 2 49 44 PM
when you hover over the first item for a folder dropdown, it doesn't highlight

i think that this should use the same styles as the navigation dropdown, so i think this should be moved into a placeholder, rather than repeating the same code twice.

plxity:
@catarak Okay, I will look into it once. and make the following changes :)

PR #890, opened 2019-03-14
Patched console.logs to work only in dev mode (#853)

siddhant1:

Also why are we still using webpack 3 , should I do the migration?


catarak:
i think i may have _tried_ to update webpack and got stuck. feel free to open an issue about it and make a pr, but i'd make that a separate issue.

catarak:
tested, and this is working for me!

Issue #789, opened 2018-12-15
Sorting sketches


CranShawdy:
#### Nature of issue?

- [ ] Found a bug
- [ ] Existing feature enhancement
- [x] New feature request

#### New feature details:

How about sorting sketches which are present in file/open/ via Sketch(name), Date created or Date updated values?

catarak:
thanks for reporting, and great idea! i think this is covered in #390 but i think it also makes sense to break this out into its own ticket as you have done.

GaurangTandon:
@catarak I can work on this issue. Can you please describe what files need to be edited? Also, do you have a UI in mind?

siddhant1:
Is this still being worked at? @catarak can i take this one?

GaurangTandon:
Yeah, @siddhant1 you can take this issue. Sorry I was unable to work on it.

siddhant1:
Alright ! I will work on this one

siddhant1:
@catarak @GaurangTandon  Can you have a look at https://github.com/siddhant1/p5.js-web-editor/tree/feature
I am not able to point at the problem , in my sense it should work .
Can you help me fix this

siddhant1:
Fixed it ! Sending a PR soon

siddhant1:
@GaurangTandon @catarak What icons should I use for carat symbol.
Should I include font-awesome?

catarak:
in the design there's not currently a carat for the sorting; however, i do think that there should be one to the right of each heading. this project uses modern pictograms, so that's a good place to pull icons from. i'm thinking these icons for now:
<img width="98" alt="screen shot 2019-03-05 at 5 21 17 pm" src="https://user-images.githubusercontent.com/6063380/53841585-20737980-3f6b-11e9-9067-d1dd5c3f4f08.png">
modern pictograms is not free so i can send you those icons.

siddhant1:
Yeah ! Please share them

PR #910, opened 2019-03-16
Add sorting to sketches #789

siddhant1:


client/modules/IDE/actions/projects.js
```diff
@@ -54,3 +55,7 @@ export function deleteProject(id) {
       });
   };
 }
+
+export function sortProjects(order, orderBy) {
```
catarak: you don't need to import the store to dispatch this action. you should use `bindActionCreators` and `mapDispatchToProps` as part of `redux` and `react-redux`

client/modules/IDE/components/SketchList.jsx
```diff
@@ -9,12 +9,19 @@ import { bindActionCreators } from 'redux';
...
+import orderListByDate from '../../../selectors/Sort-Sketches';
 
 const trashCan = require('../../../images/trash-can.svg');
+const arrowUp = require('../../../images/arrow5_up.svg');
+const arrowDown = require('../../../images/arrow5_down.svg');
 
 class SketchList extends React.Component {
   constructor(props) {
     super(props);
+    this.state = {
```
catarak: why is this in the state and not in the redux store?

client/selectors/Sort-Sketches.js
```diff
@@ -0,0 +1,23 @@
...
+
+const orderListByDate = createSelector(
+  sketchListSelector,
+  orderSelector,
+  orderBySelector,
+  (sketchList, order, orderBy) => {
+    const sortedList = sketchList.sort((a, b) => {
+      const orderResult =
+        order === 'Asc'
+          ? new Date(a[orderBy]) - new Date(b[orderBy])
```
catarak: this should use `date-fns`

client/modules/IDE/reducers/sorting.js
```diff
@@ -0,0 +1,12 @@
+import * as ActionTypes from '../../../constants';
+
+const sorting = (state = { orderBy: 'createdAt', order: 'Dsc' }, action) => {
```
catarak: i'm not sure if it makes sense to make sorting into its own reducer. i would put it in the `ide` reducer, but i could also be convinced that it's too big and needs to be broken up.

client/modules/IDE/components/SketchList.jsx
siddhant1: This is in the redux store but I had to show svgs based on createdAt and updateAt fields and didn't wanted to store both in the store, now the store just has order and orderbyy

client/modules/IDE/reducers/sorting.js
siddhant1: I feel adding sorting to anything needs a new reducer , I can change accordingly.

client/selectors/Sort-Sketches.js
siddhant1: Sure

client/modules/IDE/actions/projects.js
```diff
@@ -57,5 +56,5 @@ export function deleteProject(id) {
 }
 
 export function sortProjects(order, orderBy) {
-  return () => store.dispatch({ type: ActionTypes.SET_SORT_PARAM, payload: { orderBy, order } });
+  return dispatch => dispatch({ type: ActionTypes.SET_SORT_PARAM, payload: { orderBy, order } });
```
catarak: because the action isn't asynchronous, the action doesn't need to return a function. it can simply return an object, and you can bind it to dispatch using `bindActionCreators`/`mapDispatchToProps`

siddhant1:
Made the requested changes, but do these colors also need to update according to the theme?
sort

client/modules/IDE/components/SketchList.jsx
catarak: i'm still confused as to why this is stored in the component state—isn't just the order and order by enough?

catarak:
also, this is missing order by sketch name.

client/modules/IDE/components/SketchList.jsx
siddhant1: I'll try to explain this.
I have put in just 2 things in the redux store , order and orderBy , Say for example I click on `createdAt` and `Ascending` and then the redux store values change to order = Ascending and orderBy = Descending , Now I click on `updatedAt` and `Ascending`  and now the redux store values change to orderBy = updatedAt and order = `Ascending` , now I have to render svgs based on the current order of createdAt , but redux store holds values of updatedAtorder , so there is no way of knowing what is the current order of updatedAt.
There can be 2 solutions for this.
1. Making isolated state.
2. Maintaing local state (which I have done in this case)

catarak: i'm not sure i totally understand what you mean here. you should only ONE field, ONE direction at a time. The expected behavior is:
1. By default, the sort order is "Created Date, Descending". ONLY the "Created Date" field, since it is selected, renders the "ASC/DSC" arrows
2. If you click on another field, then simply apply a default value to order (maybe ASC? maybe DSC if it's date, ASC if it's a string?)

catarak:
> Made the requested changes, but do these colors also need to update according to the theme?
> sort

yes the arrows should match the theme!

catarak:
@andrewn i just made some changes to get this PR in working order. would you mind reviewing?

client/modules/IDE/selectors/projects.js
```diff
@@ -0,0 +1,34 @@
...
+
+const getSortedSketches = createSelector(
+  getSketches,
+  getField,
+  getDirection,
+  (sketches, field, direction) => {
+    if (field === 'name') {
+      const sortedSketches = sortBy(sketches, 'name');
+      if (direction === DIRECTION.DESC) {
+        return reverse(sortedSketches);
```
andrewn: This potentially sorts the `sketches` list twice. In practice this probably doesn't matter too much for performance. [`lodash.orderBy`](https://lodash.com/docs/#orderBy) allows the sort order to be specified.

andrewn:
This works really well!

There's a small visual glitch when you change the order. I think it's due to the dashed button border causing the button to jump up when it's added (see gif below).

catarak:
thanks for reviewing! i'll make these small changes ✨

catarak:
i'm going to merge this one!

PR #927, opened 2019-03-17
fixes #924

### LINKED ISSUE #924: Unable to type indentation amount and text size.
#### Nature of issue?

- Found a bug

#### Details about the bug: 
When tried to manually enter Text size, it lands up on 8, 16 or 36. Also, the indentation amount can only be changed by selecting the entire number and typing a new one. Hitting backspace to the existing value automatically sets it to 2 and on typing directly, it sets up to 6.
- Web browser and version: On Chrome
- Operating System: MacOSX
For both, just type in the textbox selecting and without selecting.

@catarak @siddhant1   What do you think?

GaurangTandon:

**Problem identification:** the input element was using an `onChange` event instead of an `onBlur` event. So typing `20` yielded `36` because the first input of `2` yielded `8` (as 8 is the lowest element) and the subsequent `0` made it `80` therefore we will get `36` (since 36 is upper limit).

**Problem fix:** using `onBlur` and `defaultValue`. Now, the value only updates when the user finalizes their input value and leaves (`blur`s) the input field.

### DIFF
--- a/client/modules/IDE/components/Preferences.jsx
+++ b/client/modules/IDE/components/Preferences.jsx
@@ -32,6 +32,7 @@ class Preferences extends React.Component {
       value = 8;
     }
     this.props.setFontSize(value);
+    event.target.value = value;
   }
 
   handleUpdateIndentation(event) {
@@ -46,6 +47,7 @@ class Preferences extends React.Component {
       value = 0;
     }
     this.props.setIndentation(value);
+    event.target.value = value;
   }
 
   handleUpdateAutosave(event) {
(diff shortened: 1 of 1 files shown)


catarak:
i think that these inputs should also update when you hit "Enter". so i think it makes sense to wrap them in a form and have an `onSubmit` handler.

client/modules/IDE/components/Preferences.jsx
```diff
@@ -32,6 +32,7 @@ class Preferences extends React.Component {
       value = 8;
     }
     this.props.setFontSize(value);
+    event.target.value = value;
```
catarak: what if, rather than doing this, the value is stored on the component state, and then onSubmit/onBlur it gets propagated to the redux state?

PR #930, opened 2019-03-17
Fix signup form password matching

GaurangTandon:

I fixed the problem here by:

- simply checking if the confirm password field has at least one or more characters
- displaying the error message below the confirm password field instead of the password field. I think this is an additional improvement for intuitivity, as it is the second field that says "Confirm password", so it is the one that should say that passwords do not match. (GMail signup also works the same way)

Hope it is good. Please let me know of additional suggestions for improvement @catarak


GaurangTandon:
closed as accidentally got another commit in
will make new pr

PR #926, opened 2019-03-17
change to acceptable contrast colors (fixes #925)

GaurangTandon:
I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] is from a uniquely-named feature branch and has been rebased on top of the latest master. (If I was asked to make more changes, I have made sure to rebase onto master then too)
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

@catarak I have verified that the colors that I have now selected are well above the 4.5AA guideline. Is the new scheme fine?

One question though: `secondary-text-color` and `inactive-text-color` are now both the same. Would you want me to change them slightly? Secondary text color could be a bit bolder perhaps, or just simply an even darker shade of grey? Something like `hsl(0, 0, 10%)`?

Issue #925, opened 2019-03-17
Very low contrast ratio for several heading texts in light theme


GaurangTandon:
#### Nature of issue?

- Found a bug

#### Details about the bug: 

- Web browser and version: Chrome latest
- Operating System: Ubuntu 18.04
- Steps to reproduce this bug: visit editor.p5js.org

![image](https://user-images.githubusercontent.com/6308683/54494171-85c95200-48fd-11e9-9a9a-bcb555b4b17a.png)

Notice the contrast ratio value. It is much below the acceptable 4.5 AA benchmark set by the WCAG. I personally also found it difficult to view the navbar items easily. The same goes for labels on the signup page, and several other sub-header texts. They could probably use a slightly darker shade of grey color.

GaurangTandon:
PR: #926

catarak:
This is addressed in some design changes on Zeplin that have no yet been integrated. Can you use the colors from there?

catarak:
Fixed by #1406.

PR #931, opened 2019-03-17
fixes #906

### LINKED ISSUE #906: Passwords do not match error in signup page is not intuitive
#### Nature of issue?

- Found a bug

The passwords do not match error in signup page is not intuitive.

All other error messages popup only after the respective fields are filled in.
However, 'passwords do not match' error pops up immediately after the focus of the password field is blurred. This is slightly non-intuitive and irritating. I think this error should come up only after,
1. the user has entered the confirm password field and
2. removed the focus of the same field.

#### Details about the bug: 

- Operating System: Ubuntu 18.04

@catarak what are views on this?
PS: Would like to work on this...

GaurangTandon:

fixes #906

I fixed the problem here by:

- simply checking if the confirm password field has at least one or more characters
- displaying the error message below the confirm password field instead of the password field. I think this is an additional improvement for intuitivity, as it is the second field that says "Confirm password", so it is the one that should say that passwords do not match. (GMail signup also works the same way)

Hope it is good. Please let me know of additional suggestions for improvement @catarak

### DIFF
--- a/client/utils/reduxFormUtils.js
+++ b/client/utils/reduxFormUtils.js
@@ -73,8 +73,8 @@ export function validateSignup(formProps) {
     errors.confirmPassword = 'Please enter a password confirmation';
   }
 
-  if (formProps.password !== formProps.confirmPassword) {
-    errors.password = 'Passwords must match';
+  if (formProps.password !== formProps.confirmPassword && formProps.confirmPassword) {
+    errors.confirmPassword = 'Passwords must match';
   }
 
   return errors;


plxity:
Hey @GaurangTandon, Give suitable title to your PR. As it would make easier for everyone to know what is it about. :)

GaurangTandon:
@plxity Hey, indeed I agree with you, but I have noticed that I do not get the sweet little reference box that I get otherwise right now. Notice the:

> GaurangTandon referenced a pull request that will close this issue 18 hours ago

does appear on #931 but does not appear on #925, for which I made the PR #926, where I did pay attention to the title :(

I think those references are important but it's fine if titles are important too. Is there any other way to get those dynamic references?

plxity:
@GaurangTandon oh okay, I got your point. Let's see if there is any other way for dynamic references. :).

catarak:
@plxity / @GaurangTandon you can have both the ticket linking and a descriptive name! you would name the PR something like 'Fixes #906, "Passwords much match" error only appears if both password and confirm password are filled in'

catarak:
this looks great though, merging!

PR #938, opened 2019-03-18
[WIP] fixes #934

### LINKED ISSUE #934: Merge together the exact same console events
#### Nature of issue?

Consider the following code:

```
function draw() {
	print("hello");
}
```

```
(12) hello
(5) hello
(6) hello
...
```

GaurangTandon:

----

fixes #934 

@catarak a question: Why is the `PreviewFrame` componenet handling updates to the `Console`? Shouldn't the `handleConsoleEvent` method be part of `Console` instead? I am asking because if it was a part of `Console` then I would be able to re-render the console, and not have to extraneously pass `props.consoleEvent` down to `PreviewFrame` as well.

Thoughts?

### DIFF
--- a/client/modules/IDE/components/PreviewFrame.jsx
+++ b/client/modules/IDE/components/PreviewFrame.jsx
@@ -86,7 +86,11 @@ class PreviewFrame extends React.Component {
           source: message.source
         }));
 
-      decodedMessages.every((message, index, arr) => {
+      console.log(JSON.stringify(decodedMessages));
+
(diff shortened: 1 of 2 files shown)


catarak:
you know, there actually is no good reason for `handleConsoleEvent` to be in `PreviewFrame`. i think it originally got put there because the iframe is sending messages to the parent app, so it felt like the handler for the console messages should be near the iframe. since the event handler is on the global window, you are right, it does make sense to put it with the console component.

catarak:
Closing this as it conflicts a lot with the changes I made to make the interactive console work.

Issue #698, opened 2018-09-20
Auto-refresh bug


F4Tornado:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Details about the bug: When auto-refresh is enabled and I'm typing a for loop, the editor crashes.

- Web browser and version: Google Chrome | 69.0.3497.92 (Official Build) (64-bit) (cohort: Stable)
- Operating System: Windows 10

catarak:
thanks for reporting! there's a library included in this project to prevent infinite loops from crashing the editor, which should prevent this. however, i noticed that
```js
for (var i = 0; ;) {

}
```
is caught and stopped by the `loop-protect` code, but 
```js
for (let i = 0; ; ) {

}
```
is not caught and handled by the loop protect code.

catarak:
i was looking at the `loop-protect` library and it seems like the version that i made the fork from ([jsbin/loop-protect](https://github.com/jsbin/loop-protect)) has updated the project a lot, and i think it would make sense for this project to switch from the fork i made to the jsbin library. i also just tested the two for loops on jsbin, and it seems to catch the infinite loop for both of them, so this gives me confidence that this change would fix this bug.

catarak:
i started to work on this, following the instructions [here](https://github.com/jsbin/loop-protect#example-client-implementation), but keep getting the following error from Babel: `cannot read property 'registerPlugin' of undefined`. for some reason when i do `import Babel from '@babel/standalone';`, `Babel` is still undefined.

catarak:
switching this to high priority—seems like it's causing a lot of bugs and frustration.

siddhant1:
@catarak can you help me with this.
I think we have to make changes in the previewEntry.js file?

siddhant1:
I have made some progress but not able to catch the infinite loop , I have asked the developer of this library to help me with it . I will send a PR soon to close this.

siddhant1:
I have made some progress and its running with var and while loops but for (let) loops are still not working

siddhant1:
Not sure if this is happening but I feel that we need to transpile down to es5 and then createa the transformed code , I converted it to use jsbin/loop-protect but it still stucks on for(let) loop #947

goldwasser:
[...]

PR #947, opened 2019-03-18
[ WIP ] Added new implementation of loop-protect  (#698)

siddhant1:

### DIFF
--- a/client/utils/loopProtect.js
+++ b/client/utils/loopProtect.js
@@ -1,0 +1,101 @@
+const generateBefore = (t, id) =>
+  t.variableDeclaration("var", [
+    t.variableDeclarator(
+      id,
+      t.callExpression(
+        t.memberExpression(t.identifier("Date"), t.identifier("now")),
+        []
+      )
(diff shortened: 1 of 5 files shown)


siddhant1:
This code is successfully catching for(var) , while and do while infinite loops but stucks on for(let) loop. I have raised an issue , @catarak any possible implementation that you can think of rn?

siddhant1:
@catarak , I got a lead for his one , codesandbox is implementing loop-protection , I will have a look at their code and make progress with this one.
Meanwhile , have you got any leads , I might need some help with this one

client/utils/loopProtect.js
```diff
@@ -0,0 +1,101 @@
+const generateBefore = (t, id) =>
```
catarak: where did you pull this file from?

catarak:
just took a look at this... you are definitely on the right track! are you pulling in `loop-protect` from npm or github? it seems like you are not, and i think that's the right approach.

PR #952, opened 2019-03-19
Updated server.js

### LINKED ISSUE #951: No condition for checking if only username or password is provided or if neither is provided
#### Nature of issue?

- Found a bug

#### Details about the bug: 

- Web browser and version: Google Chrome
- Operating System: Windows 10
- Steps to reproduce this bug: Inappropriate action when user just types username and forgets to type password or vice versa or neither of the two.

If a user forgets to type in any of the two ie username or password then no action is happening. Rather user should be made aware of the fact that he has missed to type username or password or both.

abhinavsagar:
[Fixes #951](https://github.com/processing/p5.js-web-editor/issues/951)

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -44,6 +44,35 @@ if (process.env.BASIC_USERNAME && process.env.BASIC_PASSWORD) {
   }));
 }
 
+else if (process.env.BASIC_USERNAME && !process.env.BASIC_PASSWORD) {
+  app.use(basicAuth({
+    users: {
+      [process.env.BASIC_USERNAME]: null
+    },
+    challenge: false
+  }));
+}
+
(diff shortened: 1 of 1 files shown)


siddhant1:
Can you fix the linting errors

abhinavsagar:
@siddhant1 Actually i am new to this. How should i go about fixing the linting errors?

siddhant1:
Hey , there is a .eslintrc file which dictates all the lint rules , we have to code according to that standards .
Now manually maintaining these standards is hard , i will suggest you get an editor (like vscode or atom) and install eslint plugin , this plugin will read . eslintrc and give you red underlines , you have to read them and fix them.

An example would be like found 4 space expected 2 , this means that you have to omit out 2 spaces .

Tell me if you need any further help.
Cheers!

Issue #568, opened 2018-02-23
Add character limit to sketch name - Back end validation


shinytang6:
#### Nature of issue?

- [x] Found a bug
- [ ] Existing feature enhancement
- [ ] New feature request

#### Details about the bug: 

name a project with special characters and try to donwload.
some of them will cause problem, e.g. &,|...

catarak:
this is two things, right? 1. sketch names need a length limit (i don't know what an appropriate character limit is) and 2. sketch names need to be... url encoded? escaped somehow? not sure.

himanshuc3:
@shinytang6 Can you provide an example. I tried a couple of them including &, | and when downloading the sketch, & is named and, | - or .

shinytang6:
@himanshuc3 What you listed is the problem:)

himanshuc3:
But what's the problem if the name of the file changes. It downloaded without any errors.

shinytang6:
it doesn't matter with the content, should download the project in what you name.

GaurangTandon:
@catarak Is anyone working on this issue? If not, then I'd like to take it up :) I plan to follow the advice given on [this post](https://superuser.com/questions/358855/what-characters-are-safe-in-cross-platform-file-names-for-linux-windows-and-os) for generating file names.

I plan to either:

1. generate file names based on the platform the user is using when they download the sketch (so same sketch can have different file names when downloaded on Windows/Linux) (also fall back to safer Windows naming when OS is not Windows/Mac/Linux)
2. generate safe cross platform file names (i.e. like Windows) every time for consistency's sake.

Would you suggest anything else? I prefer the second option. I can also display a toast message every time a restricted character is trimmed out from a file name.

catarak:
i suggest the latter. i don't think it makes sense to change the sketch titles as they are in the editor, other than giving them a length limit; rather, i think the sketch titles should be escaped only when they are downloaded and translated to the name of a zip file, as this is the only time the sketch title affects this.

catarak:
The zip file is generated on the fly, so when a browser window requests that url, it generates the zip file. The entry point, server-side, is here: https://github.com/processing/p5.js-web-editor/blob/master/server/controllers/project.controller.js#L332

GaurangTandon:
[...]

PR #962, opened 2019-03-20
Fix: Restrict sketch name length in backend

shakeabi:
- Add sketch name validation in backend
- Add respective error toasts in the frontend
- Do lint-fix

Fixes Issue #568 
@catarak What do you think?

### DIFF
--- a/client/modules/IDE/actions/project.js
+++ b/client/modules/IDE/actions/project.js
@@ -35,22 +35,24 @@ export function setProjectName(name) {
 export function getProject(id) {
   return (dispatch, getState) => {
     dispatch(justOpenedProject());
-    axios.get(`${ROOT_URL}/projects/${id}`, { withCredentials: true })
+    axios
+      .get(`${ROOT_URL}/projects/${id}`, { withCredentials: true })
       .then((response) => {
         dispatch(setProject(response.data));
         dispatch(setUnsavedChanges(false));
       })
-      .catch(response => dispatch({
-        type: ActionTypes.ERROR,
-        error: response.data
-      }));
+      .catch(response =>
+        dispatch({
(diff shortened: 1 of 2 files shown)


siddhant1:
Hey , @catarak I don't think we have any validation at the backend , I think we need a complete validation at the backend , should I open a separate issue for this?

meiamsome:
I think I would suggest we should prefer to [validate at the model level](https://mongoosejs.com/docs/validation.html) instead of ad-hoc, so as to enforce it across all current and new routes, but am open to other opinions

siddhant1:
Validating at the model level should be done but for example I send in wrong info or wrong type the server will throw an error when trying to insert in mongodb but the frontend won't get any 400 status , so I feel there should be 3 level validations , one at front end (can be disabled by changing html) , one at server level and one at db level .

meiamsome:
@siddhant1 Generally I've seen it done where you catch the save errors and forward it to the client if it is an instance of `mongoose.Error.ValidationError` as a 400 error, which also keeps the format for validation errors consistent as well as meaning you only have to implement the validation twice (Once for the model, once for the form), though I guess you can argue that you end up implementing the error flow in multiple places

siddhant1:
I agree , I think we can send mongoose error to the client , I have personally used Hapi.js Joi to validate my queries mostly but this makes sense too.

shakeabi:
@meiamsome I agree with you. Model level validation removes redundancy of validation across various routes. I think its the best solution. Will do it if its confirmed.... @catarak

catarak:
i also think that the validation should be done at the model level. would it make sense for the names to just be truncated rather than a 400 error/the save failing?

shakeabi:
@catarak Closing this as these changes are implemented in #1025

Issue #965, opened 2019-03-21
Remove indentation options from user preferences


catarak:
#### Nature of issue?

- Existing feature enhancement

#### Feature enhancement details:

4.3% of registered users have changed the indentation amount. 3% have changed space indentation to tab indentation. This seems pretty small. Since one goal of this project is to keep configuration options down, let's get rid of this and use a smart default. Since the airbnb eslint configuration uses "two spaces", I say we stick with that.

qualitymanifest:
I'll take a crack at this, it looks like it mainly just involves reverting most of the changes from this PR https://github.com/processing/p5.js-web-editor/pull/9/files

PR #9, opened 2016-07-11
Add indentation to preferences

MathuraMG:
This PR 
* adds indentation to preferences
* allows indentation and font to be a input
* changes the return syntax in preferences reducer

PR #966, opened 2019-03-22
Remove indentation configuration options

qualitymanifest:
Fixes https://github.com/processing/p5.js-web-editor/issues/965

Removed indentation configuration options using https://github.com/processing/p5.js-web-editor/pull/9 as a guide. `test` and `lint` both pass, and everything seems to run as expected, using the airbnb configuration for indentation.


client/modules/IDE/components/Editor.jsx
```diff
@@ -137,9 +137,6 @@ class Editor extends React.Component {
     });
 
     this._cm.getWrapperElement().style['font-size'] = `${this.props.fontSize}px`;
-    this._cm.setOption('indentWithTabs', this.props.isTabIndent);
```
catarak: because you removed these three options, now they are not set at all! these values should be
```js
{
  indentWithTabs: false,
  tabSize: 2,
  indentUnit: 2
}
```

client/modules/IDE/components/Editor.jsx
```diff
@@ -254,8 +244,8 @@ class Editor extends React.Component {
 
   tidyCode() {
     const beautifyOptions = {
-      indent_size: this.props.indentationAmount,
-      indent_with_tabs: this.props.isTabIndent
+      indent_size: 2,
```
catarak: also, because these values appear here too, why not put them into constants?

client/modules/IDE/components/Editor.jsx
catarak: because you removed these three options, now they are not set at all! these values should be
```js
{
  indentWithTabs: false,
  tabSize: 2,
  indentUnit: 2
}
```

catarak:
removing code feels so good  💪

qualitymanifest:
Ah, sorry, I'll take a look at this tonight. I'm new to this project and assumed that `tidyCode` would handle this functionality and that the `setOption`s were for the user-configurable aspect - didn't look into it enough!

qualitymanifest:
@catarak some notes on the changes:

When using the following:
```
  indentWithTabs: false,
  tabSize: 2,
  indentUnit: 2
```
When you press the tab key, you get an indentation that is the size of two spaces, however it is just a narrow-looking tab. You can replicate this on https://editor.p5js.org/ by choosing 2 spaces, and then pressing tab and trying to select it - you select the whole thing in one go. I'm assuming this is not the desired behavior, and that you actually want two spaces. When you start a new line inside of a block you get two actual spaces.

Looking at the CodeMirror docs it looks like the way to achieve this is by setting a key map for the tab key. This is the exact use case they use as the example for key maps, so I assume it is the only way to get two actual spaces. I couldn't find a way to make it happen with the settings above.

Also, I included the constants in `Editor.jsx` since it seems like `constants.js` is just for redux stuff - hopefully that is the correct move.

qualitymanifest:
I think I figured out why the settings above don't make the tab key use spaces - the codemirror docs are a bit vague here:
`indentWithTabs` and `indentUnit` are only for the defining the indentation the editor automatically handles, not for redefining the behavior of the tab key.
`tabSize` changes the size of a tab, but it is still a tab.

catarak:
@qualitymanifest thanks for taking a careful look at the CodeMirror docs! your explanation is really helpful.

catarak:
@qualitymanifest i also think your last commit is clearer!

PR #973, opened 2019-03-22
Refactor to async-await (#521)

### LINKED ISSUE #521: Promisify all mongoose queries
#### Nature of issue?

Right now, all of the mongoose queries use callbacks instead of the new ES6 promise syntax. This should be updated as nested callbacks are difficult to read.

siddhant1:

* [x ] is descriptively named and links to an issue number, i.e. `Fixes #521`

### DIFF
--- a/server/controllers/embed.controller.js
+++ b/server/controllers/embed.controller.js
@@ -1,12 +1,12 @@
-import jsdom, { serializeDocument } from "jsdom";
-import Project from "../models/project";
+import jsdom, { serializeDocument } from 'jsdom';
+import Project from '../models/project';
 import {
   injectMediaUrls,
   resolvePathsForElementsWithAttribute,
   resolveScripts,
   resolveStyles
-} from "../utils/previewGeneration";
-import { get404Sketch } from "../views/404Page";
+} from '../utils/previewGeneration';
+import { get404Sketch } from '../views/404Page';
 
 export async function serveProject(req, res) {
   try {
@@ -24,13 +24,13 @@ export async function serveProject(req, res) {
     jsdom.env(htmlFile, (innerErr, window) => {
       const sketchDoc = window.document;
 
-      const base = sketchDoc.createElement("base");
-      const fullUrl = `https://${req.get("host")}${req.originalUrl}`;
+      const base = sketchDoc.createElement('base');
+      const fullUrl = `https://${req.get('host')}${req.originalUrl}`;
       base.href = `${fullUrl}/`;
       sketchDoc.head.appendChild(base);
 
-      resolvePathsForElementsWithAttribute("src", sketchDoc, files);
-      resolvePathsForElementsWithAttribute("href", sketchDoc, files);
+      resolvePathsForElementsWithAttribute('src', sketchDoc, files);
+      resolvePathsForElementsWithAttribute('href', sketchDoc, files);
       resolveScripts(sketchDoc, files);
       resolveStyles(sketchDoc, files);
 
--- a/server/controllers/embed.controller.js
+++ b/server/controllers/embed.controller.js
@@ -38,7 +38,6 @@ export async function serveProject(req, res) {
     });
   } catch (err) {
     get404Sketch(html => res.send(html));
-    return;
   }
 }
 
--- a/server/controllers/embed.controller.js
+++ b/server/controllers/embed.controller.js
@@ -1,42 +1,45 @@
-import jsdom, { serializeDocument } from 'jsdom';
-import Project from '../models/project';
+import jsdom, { serializeDocument } from "jsdom";
+import Project from "../models/project";
 import {
   injectMediaUrls,
   resolvePathsForElementsWithAttribute,
   resolveScripts,
   resolveStyles
-} from '../utils/previewGeneration';
-import { get404Sketch } from '../views/404Page';
+} from "../utils/previewGeneration";
+import { get404Sketch } from "../views/404Page";
 
-export function serveProject(req, res) {
-  Project.findById(req.params.project_id)
-    .exec((err, project) => {
-      if (err || !project) {
-        get404Sketch(html => res.send(html));
-        return;
-      }
-      // TODO this does not parse html
-      const { files } = project;
-      const htmlFile = files.find(file => file.name.match(/\.html$/i)).content;
-      const filesToInject = files.filter(file => file.name.match(/\.(js|css)$/i));
-      injectMediaUrls(filesToInject, files, req.params.project_id);
+export async function serveProject(req, res) {
+  try {
+    const project = await Project.findById(req.params.project_id);
+    if (!project) {
+      get404Sketch(html => res.send(html));
+      return;
+    }
+    // TODO this does not parse html
+    const { files } = project;
+    const htmlFile = files.find(file => file.name.match(/\.html$/i)).content;
+    const filesToInject = files.filter(file => file.name.match(/\.(js|css)$/i));
+    injectMediaUrls(filesToInject, files, req.params.project_id);
 
-      jsdom.env(htmlFile, (innerErr, window) => {
-        const sketchDoc = window.document;
+    jsdom.env(htmlFile, (innerErr, window) => {
+      const sketchDoc = window.document;
 
-        const base = sketchDoc.createElement('base');
-        const fullUrl = `https://${req.get('host')}${req.originalUrl}`;
-        base.href = `${fullUrl}/`;
-        sketchDoc.head.appendChild(base);
+      const base = sketchDoc.createElement("base");
+      const fullUrl = `https://${req.get("host")}${req.originalUrl}`;
+      base.href = `${fullUrl}/`;
+      sketchDoc.head.appendChild(base);
 
-        resolvePathsForElementsWithAttribute('src', sketchDoc, files);
-        resolvePathsForElementsWithAttribute('href', sketchDoc, files);
-        resolveScripts(sketchDoc, files);
-        resolveStyles(sketchDoc, files);
+      resolvePathsForElementsWithAttribute("src", sketchDoc, files);
+      resolvePathsForElementsWithAttribute("href", sketchDoc, files);
+      resolveScripts(sketchDoc, files);
+      resolveStyles(sketchDoc, files);
 
-        res.send(serializeDocument(sketchDoc));
-      });
+      res.send(serializeDocument(sketchDoc));
     });
+  } catch (err) {
+    get404Sketch(html => res.send(html));
+    return;
+  }
 }
 
 export default serveProject;
--- a/server/controllers/file.controller.js
+++ b/server/controllers/file.controller.js
@@ -7,68 +7,90 @@ import { deleteObjectsFromS3, getObjectKey } from './aws.controller';
 // Bug -> timestamps don't get created, but it seems like this will
 // be fixed in mongoose soon
 // https://github.com/Automattic/mongoose/issues/4049
-export function createFile(req, res) {
-  Project.findOneAndUpdate(
-    {
-      _id: req.params.project_id,
-      user: req.user._id
-    },
-    {
-      $push: {
-        files: req.body
+export async function createFile(req, res) {
+  try {
+    const updatedProject = await Project.findOneAndUpdate(
+      {
+        _id: req.params.project_id,
+        user: req.user._id
+      },
+      {
+        $push: {
+          files: req.body
+        }
+      },
+      {
+        new: true
       }
-    },
-    {
-      new: true
-    }, (err, updatedProject) => {
-      if (err || !updatedProject) {
-        console.log(err);
-        res.status(403).send({ success: false, message: 'Project does not exist, or user does not match owner.' });
+    );
+
+    if (!updatedProject) {
+      res.status(403).send({
+        success: false,
+        message: 'Project does not exist, or user does not match owner.'
+      });
+      return;
+    }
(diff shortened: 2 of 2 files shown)


siddhant1:
@catarak I am working on working on all the controllers , Should I make separate prs or add commits here?

siddhant1:
I am currently having some linting error because of my global prettier settings, I will fix them at the end.

siddhant1:
ref #521

server/controllers/file.controller.js
```diff
@@ -7,68 +7,90 @@ import { deleteObjectsFromS3, getObjectKey } from './aws.controller';
...
-        if (innerErr) {
-          console.log(innerErr);
-          res.json({ success: false });
-          return;
-        }
-        res.json(updatedProject.files[updatedProject.files.length - 1]);
+      res.json(updatedProject.files[updatedProject.files.length - 1]);
+    });
+  } catch (err) {
+    if (err) {
```
meiamsome: If statement is not required. The return after the error message can also be removed.

server/controllers/file.controller.js
```diff
@@ -7,68 +7,90 @@ import { deleteObjectsFromS3, getObjectKey } from './aws.controller';
...
 }
 
 function getAllDescendantIds(files, nodeId) {
   const parentFile = files.find(file => file.id === nodeId);
   if (!parentFile) return [];
-  return parentFile.children
-    .reduce((acc, childId) => (
-      [...acc, childId, ...getAllDescendantIds(files, childId)]
-    ), []);
+  return parentFile.children.reduce(
```
meiamsome: Just as a heads up, these stylistic changes should really not be included, it breaks the git blame for this line and is generally unhelpful when looking at diffs.

That said, making an additional commit to revert them would be worse in general, so only fix that if you're going to be doing it by altering commit history. I just thought I'd point out that it is something to look out for when you're creating a PR: If you have diffs on lines irrelevant to what you're doing, something usually has gone wrong :)

server/controllers/file.controller.js
```diff
@@ -77,44 +99,68 @@ function deleteChild(files, parentId, id) {
...
+    project.files = deleteChild(
+      project.files,
+      req.query.parentId,
+      req.params.file_id
+    );
+    project = await project.save();
+    res.json(project.files);
+  } catch (err) {
+    // NOT SO GREAT ERROR HANDLING
+    console.log(err);
```
meiamsome: We should probably respond with a 500 error here so that the API user always gets a response - though don't send the user the error

server/controllers/file.controller.js
```diff
@@ -77,44 +99,68 @@ function deleteChild(files, parentId, id) {
...
+      res.status(404).send({
+        success: false,
+        message: 'File with that name and path does not exist.'
+      });
       return;
     }
     res.send(resolvedFile.content);
-  });
+  } catch (err) {
+    if (err) {
```
meiamsome: Another unnecessary if here & return below

server/controllers/file.controller.js
```diff
@@ -7,68 +7,90 @@ import { deleteObjectsFromS3, getObjectKey } from './aws.controller';
...
+    if (!updatedProject) {
+      res.status(403).send({
+        success: false,
+        message: 'Project does not exist, or user does not match owner.'
+      });
+      return;
+    }
+    const newFile = updatedProject.files[updatedProject.files.length - 1];
+    updatedProject.files.id(req.body.parentId).children.push(newFile.id);
+    updatedProject.save(innerErr => {
```
meiamsome: This can be promisified too.

server/controllers/embed.controller.js
```diff
@@ -8,35 +8,38 @@ import {
...
+      resolveScripts(sketchDoc, files);
+      resolveStyles(sketchDoc, files);
 
-        res.send(serializeDocument(sketchDoc));
-      });
+      res.send(serializeDocument(sketchDoc));
     });
+  } catch (err) {
+    get404Sketch(html => res.send(html));
+    return;
```
meiamsome: You can delete this return

siddhant1:
Thanks for the review , I will make the changes!

server/controllers/file.controller.js
siddhant1: My global prettier generally makes these changes , since the project does not has any local .prettierrc file , my global prettier kicks in and make the changes

siddhant1:
@meiamsome @catarak Should I make another pr when I am done refactoring other files?

server/controllers/file.controller.js
catarak: for this project, i squash commits for a PR, so i'm okay with you making commits to revert stuff. i also agree, these stylistic changes should not be included in this PR.

catarak:
> @meiamsome @catarak Should I make another pr when I am done refactoring other files?
honestly, feel free to do one PR per controller. i prefer smaller PRs as they are easier to test and merge.

server/controllers/embed.controller.js
```diff
@@ -8,35 +8,37 @@ import {
...
-        get404Sketch(html => res.send(html));
-        return;
-      }
-      // TODO this does not parse html
-      const { files } = project;
-      const htmlFile = files.find(file => file.name.match(/\.html$/i)).content;
-      const filesToInject = files.filter(file => file.name.match(/\.(js|css)$/i));
-      injectMediaUrls(filesToInject, files, req.params.project_id);
+export async function serveProject(req, res) {
+  try {
```
catarak: i'm not a fan of the `try` body being huge here. it's only here to catch errors from mongodb, right? so only line 13 should be in the body.

server/controllers/file.controller.js
```diff
@@ -7,64 +7,82 @@ import { deleteObjectsFromS3, getObjectKey } from './aws.controller';
...
-      }
-      const newFile = updatedProject.files[updatedProject.files.length - 1];
-      updatedProject.files.id(req.body.parentId).children.push(newFile.id);
-      updatedProject.save((innerErr) => {
-        if (innerErr) {
-          console.log(innerErr);
-          res.json({ success: false });
-          return;
+export async function createFile(req, res) {
+  try {
```
catarak: same issue here. i don't like the giant `try` body.

server/controllers/file.controller.js
```diff
@@ -7,64 +7,82 @@ import { deleteObjectsFromS3, getObjectKey } from './aws.controller';
...
+      res.status(403).send({
+        success: false,
+        message: 'Project does not exist, or user does not match owner.'
       });
+      return;
     }
-  );
+    const newFile = updatedProject.files[updatedProject.files.length - 1];
+    updatedProject.files.id(req.body.parentId).children.push(newFile.id);
+    try {
```
catarak: especially nested try/catch... i would try to avoid it if possible.

siddhant1: Is there any way to avoid them?

server/controllers/file.controller.js
siddhant1: Makes sense , I will refactor

server/controllers/embed.controller.js
meiamsome: Well, it also catches all errors that are unintended to make sure that the API user gets a response (500) when the server has a problem, instead of just having the connection hang. In non-async handlers this is handled by default by express because they try/catch all the handlers, but express doesn't support `async` natively. There is [express-async-handler](https://www.npmjs.com/package/express-async-handler) which would help us wrap these functions correctly to catch those errors, but you're just moving where the wrapping handler is.

catarak: wait, i'm confused. are you saying that express doesn't support `async/await` in the handler functions?

meiamsome: Yeah, that's right, it doesn't natively handle it, it expects the user to either throw errors synchronously, or pass them all to `next()` manually. Indeed if you look at [the source of that `async` wrapper](https://github.com/Abazhenov/express-async-handler/blob/master/index.js), it is essentially just doing this large `try`/`catch` there. It's a shame that express doesn't understand Promises, but the people behind express decided it'd be too dramatic a change for express to support them. (As an aside they actually decided to redesign express entirely and made [Koa](https://koajs.com/) which is great but differs in some ways.)

So, that leaves you with two options:
1. Don't use `async`/`await` and lose the great syntactical bonuses that gives you, keeping in mind that you still need to handle the asynchronous errors in your promise chains by eventually passing them to `next()`
2. Do use `async`/`await` and be stuck with wrapping each route's handler in some sort of wrapper, be that a library one or just a big `try`/`catch`. The standard `try`/`catch` is something akin to:
```javascript
async function (req, res, next) {
  try {
    // Really long function here
  } catch(e) {
    next(e);
  }
}
```

catarak: wow, okay. i had no idea! thanks for the detailed explanation. my instinct is to use the express-async-handler middleware, as wrapping everything in a try/catch feels like it gets redundant fast. just to make sure i totally understand this, say the middleware were included in the project, and i wanted to catch errors specific to the mongodb query/connection. that would look like 
```js
let project;
try {
  project = await Project.findById(req.params.project_id);
} catch(e) {
  // do something with the error, e.g. return 404 or 500
}
```

meiamsome: Yep, that'd work fine, and you can naturally optionally chose to rethrow the error in certain circumstances for it to be caught by express:
 ```javascript
async function (req, res, next) {
  // Calculate projectParams
  let project;
  try {
    project = await Project.create(projectParams);
  } catch(e) {
    if (!(e instanceof mongoose.Error.ValidationError)) {
      throw e; // rethrow unexpected exceptions so Express will create a 500
    }
    // Forward validation errors to clients
  }
}
```

Comment (siddhant1) [reply server/controllers/embed.controller.js]: I think that's a very good error handling approach , we can wrap out controllers inside a function which takes out route methods as callbacks and run in try / catch blocks and call next if error occurs , not we have to design custom error handlers to catch different errors (say 404/500)

Comment (catarak) [reply server/controllers/embed.controller.js]: okay great! to me, this feels like the cleanest approach.

Comment (siddhant1) [reply server/controllers/embed.controller.js]: @catarak I am somewhat confused which message are you referring to , can you give me points on what needs to be done next so I can make incremental prs to achieve that

Comment (catarak) [reply server/controllers/embed.controller.js]: @siddhant1 sure! i think that it would be best to add `express-async-handler` to this project, so that the entire body of the handler functions doesn't have to be a `try` body. otherwise you can keep doing what you were doing.

Comment (ganeshpatro321) [reply server/controllers/embed.controller.js]: @catarak There are some scenarios where express-async-handler doesn't work as expected.

@siddhant1 check @awaitjs/express before starting work.
Check this blog: http://thecodebarbarian.com/introducing-await-js-express-async-support-for-express-apps

Comment (catarak) [reply server/controllers/embed.controller.js]: @ganeshpatro321 thanks for the update! that's super frustrating. i'm starting to feel like nested callbacks is better than this mess...

Comment (siddhant1) [reply server/controllers/embed.controller.js]: The only problem with callbacks is not nesting or callbacks hell , I think if not now , we should definitely consider this as an important task in the future.

Comment (catarak) [reply server/controllers/embed.controller.js]: @siddhant1 i agree! i'm not a huge fan of the nested callbacks. however, it feels like the tools to move everything to async/await are in flux, and i think it's best to wait a little bit. 

one option could be to just move to promises. but, honestly, and maybe this is just me, but i'm not sure the promise syntax is too much cleaner/easier to read without async/await

Comment (meiamsome) [reply server/controllers/embed.controller.js]: I think it's important to note that one of the major benefits of using Promises/(async/await) is the ability to have errors bubble up naturally, rather than not being sure whether some errors may just be being silently ignored instead of raised correctly. As discussed before, Koa is a viable Express replacement that is functionally identical except for supporting promises natively. I would be happy to do an exploration PR into swapping out for that if you think it makes sense to have a look at it @catarak ?

Comment (catarak) [reply server/controllers/embed.controller.js]: that is a great point.

it seems like koa is a fairly radical departure from express, and would involve some major changes. for me, there are higher priority issues, so working on this change would have to be fairly self-directed. however, i am happy to test and give support when i can. if you think there's a possibility of significantly improved error-handling and major code clean up, it's definitely worth investigating!

File: server/controllers/file.controller.js (hunk 9 above)
Comment (anku255) [reply server/controllers/file.controller.js]: @catarak @siddhant1 I have a suggestion that can clear up this try catch litter. We can define a wrapper function to catch error and wrapper every async function with it. This way we can avoid try/catch with async await.


```js
/* Helper buddy for removing async/await try/catch litter.  
 We can make the name of this function a little smaller */
function catchError(promise) {
  return promise.then(data => {
    if (data instanceof Error) return [data]
    return [null, data]
  }).catch(err => [err])
}

// Example Usage
async function usageExample(params) {
  const [ err, data ] = await catchError(myPromise(params)) // We are using the wrapper here

  if (err) {
    // handle error
  }
  // Do stuff with data
}
```


What do you think about it?

catarak: hmmmm, i prefer `express-async-handler` as i like that it handles the error handling as middleware. to me, it feels like a cleaner solution, rather than adding a helper function to every single controller function.

siddhant1: Express-async-handler is good 
I will refactor it along with adding a better error handling approach , because for now we are just logging the errors

catarak: i recommend not just pushing ahead on something. i think there are a few different options here  and i'd like to see a comparison of these options, so we can make a decision about what the best path forward is.

catarak:
closing this PR due to inactivity, but making sure this conversation is linked to in #521.

Issue #963, opened 2019-03-21
Need to double-click data folder to add files


awarua:
#### Nature of issue?

- Found a bug

#### Details about the bug: 

- Web browser and version: Chrome

- Operating System: Windows

- Steps to reproduce this bug:
* Add a 'data' folder to the project folder
* Click on the down-arrow next to the folder to 'add file'
* Choose the file (e.g. an image) to upload.
* It will upload, but will not be inside the folder.

You need to double-click the folder in order to upload it.

catarak:
thanks for reporting! i can reproduce.

siddhant1:
I will fix this soon

PR #979, opened 2019-03-23
Fixed Folder bug (#963)

siddhant1:

### DIFF
--- a/client/modules/IDE/components/FileNode.jsx
+++ b/client/modules/IDE/components/FileNode.jsx
@@ -157,7 +157,7 @@ export class FileNode extends React.Component {
                   aria-label="view file options"
                   ref={(element) => { this[`fileOptions-${this.props.id}`] = element; }}
                   tabIndex="0"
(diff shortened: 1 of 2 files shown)


siddhant1:
Ref #963

catarak:
i don't think this is the right way to fix this issue. the add file/folder overlay needs a local state (as in, the state is in React and not Redux), which is the parent folder that the file/folder is being added to.

siddhant1:
Not sure if I understand that completely , I will try a different implementation (as you suggested) but I am not able to catch the problems with the current one.

catarak:
i'm not saying that this implementation doesn't work, i'm saying that i don't think it is the right approach.

siddhant1:
@catarak can you guide me in the right direction?

catarak:
@siddhant1 sure! rather than dispatching an action that changes the selected file, the add file/add folder modal should contain within its state the parent folder (by passing that id from the file that a user clicked the down arrow on to the modal), and then creating the file/folder within that file.

PR #980, opened 2019-03-24
Inc fix #933 - Rearranging IDE Panes for Mobile devices

### LINKED ISSUE #933: Responsive IDE: Rearranging IDE Panes for Mobile devices
#### Nature of issue?

Some panes of IDE maybe arranged column-wise instead of row-wise specifically for mobile devices for enhancing UI

In future, this feature could also provide a reponsive UI for PWA

DexterShepherd:

An incremental fix for #933. Reflows editor and preview panes to vertical orientation on screen widths less than 600px.

There is more work to do here, ideally, we would fix the top nav and icon responsiveness, but those were not explicitly mentioned in the issue so I left them alone for now. An option to full screen either the editor or preview would also be UX win, but I saw comps for that on zepplin so held off in case someone else is already working on that.

I'm not sure if this is better on the resize handler, or if we should just add a button to the UI somewhere to change layout orientation. The resize option avoided design changes so I went that route but this could easily be updated if the community prefers the other option.

I'm not a react expert so please let me know if I've done something weird or changed the wrong files. Happy to make changes and accept feedback.


DexterShepherd:
p5-editor-resize

iphone 7+
Image-1

siddhant1:
Looks like login and sign up are having some issues.

DexterShepherd:
Issues with functionality or layout? The whole top nav has layout problems on small widths, but that wasn’t the specific issue in the ticket so I didn’t tackle that here. I’d be happy to look into it though.

DexterShepherd:
@siddhant1 As far as I can tell the functionality is still working. Let me know if you think I should tackle the layout issues here, or keep this as an incremental fix and leave the more css heavy stuff to another PR/Issue.

siddhant1:
I think we can tackle this one In this , I think it's saying log in up , can you just make the 'up' stay with the sign ,I think that will be perfect.

DexterShepherd:
Awesome. I've updated the nav stylesheet to keep the log in and sign up buttons in single lines on smaller screens. I also added the $break-small var to `_variables.scss` to handle the media query. I couldn't see any other media queries in the codebase so let me know if this wasn't the desired approach. I also swapped out the`or` text for a horizontal line on mobile to avoid vertical overflows with 3 lines trying to fit in the top nav.

Screen Shot 2019-03-25 at 12 44 09 PM

Thank you for taking the time to review this. Let me know if these changes are too invasive or need to be refactored in some way.

DexterShepherd:
@catarak I think we are good to go here whenever you are. Let me know if you want me to update anything else.

catarak:
thanks for working on this! i'd prefer for the nav changes to be separate from the panel re-arrangement stuff—i think making these mobile changes is going to be easier if it's done on a per-component basis. 

i noticed that when the panes re-arrange to be above each other, you can't scroll at all. i think it would make sense for them to have a minimum height or something. in fact, the preview gets cut off because you can't scroll at all.

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -182,6 +193,27 @@ class IDEView extends React.Component {
     }
   }
 
+  updatePaneOrientation() {
+    const minHorizontalWidth = 600;
```
catarak: ideally, the panel moving would be done with CSS only. in this case, it might not be possible. what you could do, however, is change the orientation from horizontal to vertical (can you use the smallest Bootstrap breakpoint for this?) and then add a class, which would be used to change the CSS.

client/styles/abstracts/_variables.scss
```diff
@@ -1,5 +1,8 @@
 $base-font-size: 12;
 
+//break points
+$break-small: 600px;
```
catarak: could you use the Bootstrap breakpoint numbers? thanks!

client/styles/components/_nav.scss
```diff
@@ -103,6 +103,29 @@
   }
 }
 
+.nav__items-right--logged-out {
```
catarak: i think it would make sense to treat the nav changes in a separate PR—changing the panels to be vertical instead of horizontal feels like a no-brainer, whereas the nav changes on mobile i think need a larger re-design.

DexterShepherd:
@catarak Hmm. I hadn't noticed that but I can look into it further tomorrow. I had to add ```
  -webkit-overflow-scroll: touch;
  overflow-y: auto;
```  to `.preview-frame__content` to allow touch scrolling of the preview iframe so I suspect that caused the issue.

client/styles/components/_nav.scss
DexterShepherd: Makes sense to me. I can revert these.

client/modules/IDE/pages/IDEView.jsx
DexterShepherd: I agree, CSS was my first thought as well. I went down this route after not having much luck with CSS but I can investigate further.

client/styles/abstracts/_variables.scss
DexterShepherd: Sure thing. Sorry about that.

client/modules/IDE/pages/IDEView.jsx
catarak: for sure! i think you'll find it's easier if you add classes with JS

client/styles/abstracts/_variables.scss
catarak: no worries!

catarak:
@DexterShepherd okay! i think it might be two separate things too, (1) i think on mobile the code editor and preview frame need to have a minimum height, and (2) possibly something up with scrolling.

catarak:
I appreciate all of the work you did on this! I had a GSoC mentee implement this, and I appreciate so much the push to think and work on this.

PR #984, opened 2019-03-24
Fixed Issue #895

### LINKED ISSUE #895: Entire line selection on clicking the index number/
#### Nature of issue?

- New feature request

#### Details about the bug: 
Like other editors, there should be a feature to select the entire line on clicking the index number of that line.
- Web browser and version:  On Chrome
- Operating System: MacOSX

#### New feature details:
This can be done by setting an onclick method and indexing the lines to their respective numbers and highlighting it. What do you think @catarak ?

AnkurIngale:

* [x] Fixes #895 

The selection works once a click on the editor is done. (meaning it doesn't work on a newly made sketch until the editor is clicked once). @catarak  views?

### DIFF
--- a/client/modules/IDE/components/Editor.jsx
+++ b/client/modules/IDE/components/Editor.jsx
@@ -135,24 +135,10 @@ class Editor extends React.Component {
         this.tidyCode();
       }
     });
-    let clicks = 0;
+
+    this._cm.focus();
     this._cm.on('gutterClick', (instance, n, gutter, e) => {
-      if (clicks >= 2)clicks = 0;
-      clicks += 1;
-      if (clicks === 1) {
-        setTimeout(() => {
-          clicks -= 1;
-          if (clicks === 1) {
-            clicks = 0;
-            if (gutter.indexOf('CodeMirror-foldgutter') !== -1) {
-              this._cm.setSelection({ line: n, ch: 0 }, { line: n + 1, ch: 0 });
-            }
-            clicks = 0;
-            return null;
-          }
-          return null;
-        }, 400);
-      }
+      this._cm.doc.setCursor({ line: n, ch: this._cm.doc.getLine(n).length });
     });
     this._cm.getWrapperElement().style['font-size'] = `${this.props.fontSize}px`;
     this._cm.setOption('indentWithTabs', this.props.isTabIndent);
--- a/client/modules/IDE/components/Editor.jsx
+++ b/client/modules/IDE/components/Editor.jsx
@@ -135,7 +135,25 @@ class Editor extends React.Component {
         this.tidyCode();
       }
     });
-
+    let clicks = 0;
+    this._cm.on('gutterClick', (instance, n, gutter, e) => {
+      if (clicks >= 2)clicks = 0;
+      clicks += 1;
+      if (clicks === 1) {
+        setTimeout(() => {
+          clicks -= 1;
+          if (clicks === 1) {
+            clicks = 0;
+            if (gutter.indexOf('CodeMirror-foldgutter') !== -1) {
+              this._cm.setSelection({ line: n, ch: 0 }, { line: n + 1, ch: 0 });
+            }
+            clicks = 0;
+            return null;
+          }
+          return null;
+        }, 400);
+      }
+    });
     this._cm.getWrapperElement().style['font-size'] = `${this.props.fontSize}px`;
     this._cm.setOption('indentWithTabs', this.props.isTabIndent);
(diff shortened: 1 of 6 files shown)


client/modules/IDE/components/Editor.jsx
```diff
@@ -135,7 +135,25 @@ class Editor extends React.Component {
         this.tidyCode();
       }
     });
-
+    let clicks = 0;
+    this._cm.on('gutterClick', (instance, n, gutter, e) => {
+      if (clicks >= 2)clicks = 0;
+      clicks += 1;
+      if (clicks === 1) {
+        setTimeout(() => {
```
catarak: is there a way to do this without using a `setTimeout`? this clicks counting just seems... prone to bugs.

AnkurIngale: I can remove clicks counting but not the timeout. As timeout is needed to differentiate between a click and a double click. Still, I'll do some research to find an alternative.

AnkurIngale: I tried other ways to resolve this but ended up with some bugs. But the above one doesn't give me any error. I'll just shorten the timeout.

catarak: i think it's okay if a click or double click highlights the line. that's the behavior in VSC at least.

AnkurIngale: That's what this is doing. What do you mean exactly? Just highlight and not select?

catarak: i guess i'm wondering why you need to differentiate between a click and a double click?

AnkurIngale: That's because I wanted to differentiate between folding and highlighting. Single click on the line containing folding marker will highlight the folded code. But I wanted to highlight that specific line. That's it.

catarak:
i think it's awkward that when you click on a line with a foldgutter, you have to click it twice to highlight the line. i'm also not able to get the line highlighting to work consistently.

AnkurIngale:
@catarak Works completely fine for me. What errors are you getting? Also, on double clicking a line with a foldgutter, it opens the foldgutter if closed and highlights the line. Is it behaving in the same manner for you as well right?

catarak:
it's actually just, not working for me at all. Using Chrome/Mac OSX. If I click or double-click in the gutter, nothing happens, unless it's on a line with a foldgutter, and then the function will open or close.

AnkurIngale:
You must be trying on a newly made sketch. I forgot to focus the editor before it. Just click once on any line inside the editor and try again. Sorry for the inconvenience.

catarak:
ah okay! thanks for clarifying.

catarak:
overall this feels awkward—i like the UX in visual studio code, in which the cursor changes, and you only have to click once. i understand that it's hard to differentiate it from the foldgutter click, so i'm not sure how to solve this.

AnkurIngale:
If there's any way to distinguish between foldmarker and a normal line, it would make things easy. Like this editor has https://ide.geeksforgeeks.org/ . Cursor can be changed that's not an issue. Problem is that entire number area is a foldgutter so I cannot distinguish it. Also, if you keep single click, folding doesn't work. What we can do for now is on clicking a line, make it a active line. That sounds good at this point in time. @catarak what do you say?

catarak:
i would try to imitate the behavior on the IDE you've linked to, if possible. i don't think clicking on a line should highlight the whole thing—for me that would be super annoying, as sometimes you're just trying to put your cursor somewhere.

AnkurIngale:
@catarak  Yeah, you are right. Can you please check the working of my latest commit? On line click it brings the cursor to the end of that line. I'll close this PR once you review this.

catarak:
i think this is working better. i'm not sure how to handle differentiating clicking on the gutter vs on the arrow unfortunately!

AnkurIngale:
@catarak I agree. I guess I'll let the issue open and close this request.

catarak:
you don't need to close this! i'm just saying i don't know how to differentiate those events, but it may be possible.

AnkurIngale:
@catarak No, actually I see a series of problems for now. Differentiating between fold markers and rest, for example, can be done using double click but again that is annoying. Rather, I feel we need to discuss this issue properly once.

catarak:
okay, i hear that. can you update the issue with the code solution outlined here, so it doesn't get lost?

AnkurIngale:
@catarak  Ok I'll do that.

Issue #987, opened 2019-03-25
Multiple File Options


nimishagarwal76:
#### Nature of issue?

- Found a bug

#### Details about the bug: 

Latest Merge ```Fixed Issue #659 (#964)``` causes multiple menus to remain open.
![Screenshot (1)](https://user-images.githubusercontent.com/37630020/54921104-ef171980-4f2a-11e9-8650-6df203cfad4f.png)

#876 removed the issue but it comes again now.

Marton6:
Hello! I would like to work on this issue.

PR #988, opened 2019-03-25
Fix multiple file options showing up (#987)

Marton6:
This change ensures that the file options composite widget is
hidden if and only if none of its components is in focus. This
is achieved by having a variable keep track of the state of
the composite widget (in focus / not in focus).

### DIFF
--- a/client/modules/IDE/components/FileNode.jsx
+++ b/client/modules/IDE/components/FileNode.jsx
@@ -105,6 +105,9 @@ export class FileNode extends React.Component {
       'sidebar__file-item--editing': this.state.isEditingName,
       'sidebar__file-item--closed': this.props.isFolderClosed
     });
+
+    let isFocused = false;
+
     return (
       <div className={itemClass}>
         {(() => { // eslint-disable-line
@@ -156,6 +159,15 @@ export class FileNode extends React.Component {
                   ref={(element) => { this[`fileOptions-${this.props.id}`] = element; }}
                   tabIndex="0"
                   onClick={this.toggleFileOptions}
+                  onBlur={() => {
+                    isFocused = false;
+                    setTimeout(() => {
+                      if (!isFocused) {
+                        this.hideFileOptions();
+                      }
+                    }, 200);
+                  }}
+                  onFocus={() => { isFocused = true; }}
                 >
                   <InlineSVG src={downArrowUrl} />
                 </button>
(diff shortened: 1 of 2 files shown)


client/modules/IDE/components/FileNode.jsx
```diff
@@ -105,6 +105,9 @@ export class FileNode extends React.Component {
       'sidebar__file-item--editing': this.state.isEditingName,
       'sidebar__file-item--closed': this.props.isFolderClosed
     });
+
+    let isFocused = false;
```
catarak: rather than this being a local variable with `render`, this should be stored within the component state. the functions that are called from `onBlur` could be moved outside of render too, methods of `FileNode`.

client/modules/IDE/components/Sidebar.jsx
```diff
@@ -41,6 +41,7 @@ class Sidebar extends React.Component {
   }
 
   render() {
+    let isFocused = false;
```
catarak: The same thing that I said about `FileNode applies here.`

client/modules/IDE/components/FileNode.jsx
Marton6: Thank you for the feedback! I'll fix this soon.

Marton6: Fixed it.

client/modules/IDE/components/Sidebar.jsx
Marton6: I made the required changes.

client/modules/IDE/components/FileNode.jsx
```diff
@@ -156,6 +168,8 @@ export class FileNode extends React.Component {
                   ref={(element) => { this[`fileOptions-${this.props.id}`] = element; }}
                   tabIndex="0"
                   onClick={this.toggleFileOptions}
+                  onBlur={this.blurComponent}
+                  onFocus={() => { this.setState({ isFocused: true }); }}
```
catarak: could you move this into a function called `onfocusComponent`?

client/modules/IDE/components/FileNode.jsx
```diff
@@ -156,6 +168,8 @@ export class FileNode extends React.Component {
                   ref={(element) => { this[`fileOptions-${this.props.id}`] = element; }}
                   tabIndex="0"
                   onClick={this.toggleFileOptions}
+                  onBlur={this.blurComponent}
```
catarak: could you call this function `onBlurComponent` rather than `blurComponent`?

catarak:
thanks for making the changes! i've asked for a couple more organizational things but this is looking great.

client/modules/IDE/components/FileNode.jsx
Marton6: Definitely! I will fix this now.

client/modules/IDE/components/FileNode.jsx
Marton6: Yes! Thank you for the feedback.

catarak:
this is working great for me. merging!

Issue #917, opened 2019-03-17
Navbar not responsive on mobile devices


Shubhanshu88:
#### Nature of issue?

- Found a bug

#### Details about the bug: 
On mobile devices , some navbar elements tend to overlap due to which the ui is not displayed as predicted

- Steps to reproduce this bug: Either use mobile view in chrome devtools or open the editor on a mobile device

![IMG_20190317_132647](https://user-images.githubusercontent.com/42772987/54487760-8f33c980-48bf-11e9-824d-367693ca957d.jpg)

GaurangTandon:
Hey, I tried fixing this but soon enough realized that this cannot be fixed reasonably well enough unless we use a collapsing/hamburger navbar at lower widths. 

One simple fix I tried was to remove all padding between the list items, even then the navbar breaks at near ~400px viewport width. Reducing the font size beyond the current set would make the navbar almost invisible though, so there does not seem to be any other option other using a hamburger menu.

Thoughts?

siddhant1:
I think that the whole ui is still not mobile responsive and needs to be done , maybe @catarak can shed some light on this.

Shubhanshu88:
There might be 2 ways to fix this :
1. The current base-font-size maybe decreased accordingly with screen size

2. A certain constant width of Navbar (and hence the whole Editor) has to be fixed below which horizontal overflow is possible and the user can scroll horizontally   (this width might be around 363px for current base font size)

OR Possibly a combination of both OR other ideas can be implemented

GaurangTandon:
Hey @Shubhanshu88, I have questions regarding your fixes:

1. We cannot decrease the font-size beyond a certain limit of readability, so the display will still not be proper on smaller phones (5.5" and below I believe)

2. I am sorry but I have never seen horizontally scrolling navbars in popular websites, and thus they might be very counter-intuitive to new users. Are they common? Can you link to a nice popular implementation on a production website?

Thanks!

Shubhanshu88:
I agree that the font size can be decreased only to certain extent, yet unlike other sites a dropdown for nav items would not be user-friendly.

A Better implementation could be:  
To keep the p5 logo and Log in nav-item on one nav and just below that the other nav-items(file, edit .. ) on another nav. 
This would be much like what Github itself implements for mobile devices.This might work out.

Till then it would be better to wait for further instructions

catarak:
[...]

PR #1000, opened 2019-03-28
Responsiveness added to log in screen, sign up screen

### LINKED ISSUE #972: Responsiveness for Log in screen: Currently, elements not working as intended
#### Nature of issue?

- Found a bug

#### Details about the bug: 

    - Open Log in screen on any mobile device (either portrait or landscape view or both)

         - The buttons do not behave as intended (decreased height, even shorter than logo),
         - Also below certain width the "Log In" form heading is not properly aligned vertically (in few cases, not even visible on screen)
         - In portrait view, the username and password input boxes have 100% width, leaving no padding

issue_login_p5

Shubhanshu88:
I have verified that this pull request: Fixes #972 & #917

### DIFF
--- a/client/styles/abstracts/_placeholders.scss
+++ b/client/styles/abstracts/_placeholders.scss
@@ -110,7 +110,7 @@
 	border-radius: 2px;
 	padding: #{8 / $base-font-size}rem #{25 / $base-font-size}rem;
 	line-height: 1;
-  margin: #{39 / $base-font-size}rem 0 #{24 / $base-font-size}rem 0;
+	margin: #{24 / $base-font-size}rem 0 #{12 / $base-font-size}rem 0;
 	&:enabled:hover {
 		border-color: $form-button-background-hover-color;
 		background-color: $form-button-background-hover-color;
--- a/client/styles/abstracts/_variables.scss
+++ b/client/styles/abstracts/_variables.scss
@@ -1,4 +1,8 @@
 $base-font-size: 12;
+$sm: 576; // Small devices (landscape phones, 576px and up)
+$md: 768px; // Medium devices (tablets, 768px and up)
+$lg: 992px; // Large devices (desktops, 992px and up)
+$xl: 1200px; // Extra large devices (large desktops, 1200px and up)
 
 //colors
 $p5js-pink: #ed225d;
--- a/client/styles/components/_form-container.scss
+++ b/client/styles/components/_form-container.scss
@@ -1,6 +1,6 @@
 .form-container {
 	text-align: center;
-	height: 100%;
+	min-height: #{556 / $base-font-size}rem;
 	display: flex;
 	flex-direction: column;
   align-items: center;
@@ -19,6 +19,10 @@
 	flex-direction: column;
 	justify-content: center;
 	align-items: center;
+	width:95%;
+	@include media-breakpoint-up($sm) {
+		width:auto;
+  }
 }
 
 .form-container__title {
--- a/client/styles/components/_forms.scss
+++ b/client/styles/components/_forms.scss
@@ -1,3 +1,10 @@
+.form {
+  width:90%;
+  @include media-breakpoint-up($sm) {
+    width:auto;
+  }
+}
+
 .form-error {
   display: block;
   padding-top: #{4 / $base-font-size}rem;
@@ -27,7 +34,12 @@
 }
 
 .form__input {
-  width: #{360 / $base-font-size}rem;
+  // position: relative;
+  width:100%;
+  max-width: #{360 / $base-font-size}rem;
+  @include media-breakpoint-up($sm) {
+    width: #{360 / $base-font-size}rem;
+  }
   height: #{40 / $base-font-size}rem;
   color: $icon-hover-color;
   border-color: $secondary-form-title-color;
@@ -44,6 +56,9 @@
 
 .form input[type="submit"] {
   @extend %forms-button;
+  @include media-breakpoint-up($sm) {
+    margin: #{39 / $base-font-size}rem 0 #{24 / $base-font-size}rem 0;
+  }
 }
 
 .form input[type="submit"]:disabled {
--- a/client/styles/components/_github-button.scss
+++ b/client/styles/components/_github-button.scss
@@ -12,7 +12,11 @@
       border-color: getThemifyVariable('secondary-text-color');
     }
   }
-  width: #{300 / $base-font-size}rem;
+  width:95%;
+  max-width: #{300 / $base-font-size}rem;
+  @include media-breakpoint-up($sm) {
+    width: #{300 / $base-font-size}rem;
+  }
   display: flex;
   justify-content: center;
   align-items: center;
@@ -20,6 +24,7 @@
 
 .github-icon {
   margin-right: #{10 / $base-font-size}rem;
+  display: flex;
 }
 
 
@@ -42,7 +47,11 @@
   flex-direction: row;
   justify-content: center;
   align-items: center;
-  width: #{300 / $base-font-size}rem;
+  width:95%;
+  max-width: #{300 / $base-font-size}rem;
+  @include media-breakpoint-up($sm) {
+    width: #{300 / $base-font-size}rem;
+  }
 }
 
 .google-icon {
--- a/client/styles/abstracts/_mixins.scss
+++ b/client/styles/abstracts/_mixins.scss
@@ -37,3 +37,15 @@
   cursor: pointer;
   padding: 0;
 }
+
+@mixin media-breakpoint-up($screen) {
+  @media only screen and (min-width: $screen + 'px') {
+    @content;
+  }
+}
+
+@mixin media-breakpoint-down($screen) {
+  @media only screen and (max-width: $screen + 'px') {
+    @content;
+  }
+}


Shubhanshu88:
This PR adds the following Enhancements:
  - Responsiveness added to toolbar
  - Fixes navbar nav-element overlaps and responsiveness (till 380px)
  - Fixes Log In screen and Sign Up screen input field widths for mobile devices
  - Fixes Log In screen form and button behavior

client/styles/abstracts/_variables.scss
```diff
@@ -1,4 +1,8 @@
 $base-font-size: 12;
+$sm: 575.98; // Small devices (landscape phones, 576px and up)
```
catarak: why is this not 576?

client/styles/components/_forms.scss
```diff
@@ -1,3 +1,9 @@
+.form {
+  @include media-breakpoint-down(390) {
```
catarak: can you not use an inline number here, but assign it to a variable?

client/styles/components/_form-container.scss
```diff
@@ -1,6 +1,7 @@
 .form-container {
 	text-align: center;
 	height: 100%;
+	min-height: 556px;
```
catarak: can you use a `rem` measurement here?

client/styles/components/_nav.scss
```diff
@@ -38,17 +38,33 @@
   justify-content: center;
   height: 100%;
 
+  // @include media-breakpoint-down($sm) {
```
catarak: can you remove this commented out code?

Shubhanshu88:
@catarak I've made some changes, are there any further changes required?

client/styles/abstracts/_variables.scss
```diff
@@ -1,4 +1,10 @@
 $base-font-size: 12;
+$sm: 576; // Small devices (landscape phones, 576px and up)
+$md: 768; // Medium devices (tablets, 768px and up)
+$lg: 992; // Large devices (desktops, 992px and up)
+$xl: 1200; // Extra large devices (large desktops, 1200px and up)
+$form-breakpoint: 390;
```
catarak: rather than this being $form-breakpoint, i think it makes sense to have a number around this small that's the small device portrait number

client/styles/abstracts/_variables.scss
```diff
@@ -1,4 +1,10 @@
...
+$github-button-breakpoint: 320;
```
catarak: i don't like adding arbitrary breakpoints to get something to look right

Shubhanshu88:
Sorry for the arbitrary breakpoints, I should have disscussed beforehand about including them,I had taken the sm, md and xl variables same as those used by bootstrap for responsive mixins. For now I think we can keep only sm and form-breakpoint and instead the form-breakpoint can be changed to xsm and I'll remove github-button breakpoint and use the xsm breakpoint instead.
@catarak would these solve the problem?

And the margins changed in the placeholder were because I changed all max-width media queries to min-width and so the I used the original margins in a media query for larger screens and I thought it might be better to keep the margins for mobile views lesser so that the vertical spacing between elements does not seem awkward when viewed from smaller screens

catarak:
is it possible to use 576 as the smallest breakpoint?

Shubhanshu88:
Actually I had shifted the original margins(without changes) for width above 576 without changing them using min width media queries
https://github.com/processing/p5.js-web-editor/blob/d8b9f0c8d4caa8d1770aceb10c742155f9956073/client/styles/components/_forms.scss#L58-L60

While I have only changed the margins in placeholder as they were for width less than 576 
https://github.com/processing/p5.js-web-editor/blob/d8b9f0c8d4caa8d1770aceb10c742155f9956073/client/styles/abstracts/_placeholders.scss#L113
and had no effects on margins used for desktop and above 576(original margins) and only changing for below 576 width. I did this as I tried to use mobile first css and some min-width media queries
Sorry again for not mentioning this beforehand

For the variables, I can use 576 as smallest breakpoint using some max-width properties

client/styles/abstracts/_variables.scss
```diff
@@ -1,4 +1,5 @@
 $base-font-size: 12;
+$sm: 576; // Small devices (landscape phones, 576px and up)
```
catarak: why did you take out the other breakpoints? it's nice to aspire 🌈

client/styles/components/_form-container.scss
```diff
@@ -1,6 +1,7 @@
 .form-container {
 	text-align: center;
 	height: 100%;
+	min-height: #{556 / $base-font-size}rem;
```
catarak: i imagine this min-height should be removed for desktop

client/styles/components/_nav.scss
```diff
@@ -39,16 +39,27 @@
   height: 100%;
 
   & button {
-    padding: #{12 / $base-font-size}rem #{10 / $base-font-size}rem;
+    padding: 0 #{5 / $base-font-size}rem;
```
catarak: i don't think these changes to the Nav are really helping that much. i'd prefer to have a drill down situation here. and maybe add another toolbar or something that's stuff that folks would want to do a lot on mobile, i.e. examples, open.

client/styles/components/_toolbar.scss
```diff
@@ -63,7 +63,10 @@
 }
 
 .toolbar {
-	padding: #{20 / $base-font-size}rem #{34 / $base-font-size}rem;
+	padding: #{20 / $base-font-size}rem #{15 / $base-font-size}rem;
```
catarak: also the changes here, i feel like they're helping a little but overall the toolbar still doesn't look great. i'd prefer to hide some of these elements and be able to drill down into them.

client/styles/abstracts/_variables.scss
Shubhanshu88: I will revert the changes and include only the standard breakpoints

client/styles/components/_form-container.scss
Shubhanshu88: The min-width has no effect on desktop view (556 is default rendering height of the form), although I can add a media query , Should I?

client/styles/components/_nav.scss
Shubhanshu88: The current changes were to prevent overlap of nav-items and logo below 576 till 390, if a new nav is required for below 576, I can revert the changes

Idea for the nav:
 - The logo, login button/user account can remain in the first nav, 
 - A second nav containing file, edit, sketch, help and examples nav items (or their respective icons) can be implemented while the toolbar can be put beneath them. 
 - The toolbar can have current sketch information, name and alongside can be a dropdown containing options for full preview, full code editor view or combined view.

A second implementation may include using a hamburger menu, but the content to be used in it needs to be discussed as it might hinder the discoverability and usage for users but it may also be a good alternative for responsive navbar.
@catarak could you suggest something?

client/styles/components/_toolbar.scss
Shubhanshu88: I think I should revert these changes for now and devote this pr to the form responsive commits only. After a solution to navbar and toolbar appearance, I can create a seperate pr.

client/styles/components/_form-container.scss
catarak: ah okay!  is this necessary because the parent is `height: 100%`? maybe there's a way to take that out?

client/styles/components/_nav.scss
catarak: hmmm. i think i need to set aside some time to really think about what is necessary for mobile going forward—the design just isn't there yet and it's hard to say yes or no to this. i thought it made sense to think about just rendering sketches on mobile at first, but it seems like there's a big demand for writing code on mobile as well.

client/styles/components/_form-container.scss
Shubhanshu88: Maybe I could set the height property to 556 or remove the property and just leave min-width, would any of these be fine?

Shubhanshu88:
@catarak, could you please review again?

client/styles/components/_form-container.scss
catarak: i think it would be okay to just leave min-width!

Shubhanshu88:
form_responsive


@catarak , I think the login screen is almost responsive, are there any other changes I missed out?

client/styles/abstracts/_placeholders.scss
```diff
@@ -110,7 +110,7 @@
 	border-radius: 2px;
 	padding: #{8 / $base-font-size}rem #{25 / $base-font-size}rem;
 	line-height: 1;
-  margin: #{39 / $base-font-size}rem 0 #{24 / $base-font-size}rem 0;
+	margin: #{24 / $base-font-size}rem 0 #{12 / $base-font-size}rem 0;
```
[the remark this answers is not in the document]
catarak: rather than making this change for all sizes, could this be for only small screen sizes?

client/styles/components/_form-container.scss
```diff
@@ -19,6 +19,10 @@
 	flex-direction: column;
 	justify-content: center;
 	align-items: center;
+	width:95%;
```
catarak: rather than doing a percent width here, i think it makes sense to add a fixed padding or margin.

client/styles/components/_forms.scss
```diff
@@ -1,3 +1,10 @@
+.form {
+  width:90%;
```
catarak: rather than doing a percent width here, i think it makes sense to add a fixed padding or margin.

client/styles/components/_github-button.scss
```diff
@@ -12,14 +12,19 @@
       border-color: getThemifyVariable('secondary-text-color');
     }
   }
-  width: #{300 / $base-font-size}rem;
+  width:95%;
```
catarak: rather than doing a percent width here, i think it makes sense to add a fixed padding or margin.

client/styles/components/_github-button.scss
```diff
@@ -42,7 +47,11 @@
   flex-direction: row;
   justify-content: center;
   align-items: center;
-  width: #{300 / $base-font-size}rem;
+  width:95%;
```
catarak: rather than doing a percent width here, i think it makes sense to add a fixed padding or margin.

catarak:
thanks for your patience working on this! i have a few more suggestions.

Issue #985, opened 2019-03-24
Cannot exit log in and sign up pages on opening sign up page from log in page.


AnkurIngale:
#### Nature of issue?

- Found a bug

#### Details about the bug: 
Opened log in page, from there I opened sign up page. Now when I click on 'X' (exit), it brings me back to log in. Again clicking on exit brings me back to Sign up. And this continues.
- Web browser and version:  On Chrome
- Operating System: MacOSX
- Steps to reproduce this bug:
Do as above.

@catarak views? Obviously, I won't be playing around here. Normally, I should fill up and submit one of the two. But this is annoying.

siddhant1:
Hm , looks like some kind of bug , will try to reproduce

catarak:
thanks for reporting! this is a bug and i can reproduce.

Marton6:
I would like to fix this.

AnkurIngale:
@Marton6 Make sure you get the permission before doing it. Thanks btw.

Marton6:
@AnkurIngale I'm sorry, will pay attention to this in the future. Thanks for pointing it out!

catarak:
@Marton6 / @AnkurIngale unless an issue has been claimed, you can assume it's open and available to work on! no need to as permission.

catarak:
Fixed by the merge of #1340

PR #1002, opened 2019-03-28
Fixes "Cannot exit log in and sign up pages" #985

Marton6:


catarak:
this fixes the bug of not being able to exit out of sign up or login; however, i don't like what this does to the browser history! ideally, the last page that was open before the user opened login or signup would be pushed onto the browser history.

Marton6:
Okay. I will look into this.

siddhant1:
In my opinion when the user clicks on the cross (x) button , he/she should land at the home page

siddhant1:
I was looking at how jsfiddle handles it , and it looks like they do not even provide a (x) button , they provide a logo which routes to home page.

siddhant1:
Also looking at trello.com , they also handle similar to jsfiddle (no cross button , just a logo to go back at the home page)

Issue #989, opened 2019-03-25
Folder can be renamed to "" empty string


Marton6:
#### Nature of issue?

- Found a bug

#### Details about the bug: 
It is possible to clear the name of a folder by renaming it to "" and files can be renamed to only contain a valid extension (".js", ".css" etc.).
Moreover two or more folders with an empty name ("") can be contained in the same folder which will probably cause issues.

- Web browser and version: Chrome 71.0.3578.98
- Operating System: Windows and Linux
- Steps to reproduce this bug: 
    1. Create a folder with a proper name.
    2. Click rename in the file navigator.
    3. Clear the name of the folder and click outside of it. 

![Screenshot_20190325_220842](https://user-images.githubusercontent.com/13394704/54954293-9788a600-4f4a-11e9-984b-b190b035e34b.png)

siddhant1:
Looks like a bug to me , will fix this soon!

catarak:
thanks for reporting! this is a bug and i can reproduce.

siddhant1:
@catarak I think just adding a length cap might work?

siddhant1:
I think as discussed on https://github.com/processing/p5.js-web-editor/pull/962#issuecomment-475330642 we should also add some validations at the backend and s3

catarak:
i think there should be frontend and backend validations.

olalekanteeblaze:
I'm a first time contributor, is this still open for me to work on?

siddhant1:
Hey sure , go for it ! What you need to do an to counter this one.
1. Add backend model level validations for folder name
2. Add a front end validation or some sort of min length validation
3.  I think we have some prs already doing the 1 , you can pick up (2).
 
If you have any doubt , most people here will be happy to help(or just tag me).
Congratulations of your first oss project.
Cheers!

PR #1025, opened 2019-04-09
Fix: Add backend validation for sketch and file names

### LINKED ISSUE #568: Add character limit to sketch name - Back end validation

shakeabi:
- Truncate sketch names to max length
- Add validations for sketch names
- Add validation for file names

Fixes #568 and #989 

@catarak what do you think?

### DIFF
--- a/client/modules/IDE/components/FileNode.jsx
+++ b/client/modules/IDE/components/FileNode.jsx
@@ -67,6 +67,7 @@ export class FileNode extends React.Component {
   validateFileName() {
     const oldFileExtension = this.originalFileName.match(/\.[0-9a-z]+$/i);
     const newFileExtension = this.props.name.match(/\.[0-9a-z]+$/i);
+    const fileName = this.props.name.replace(/\.[^/.]+$/, '');
     if (oldFileExtension && !newFileExtension) {
       this.props.updateFileName(this.props.id, this.originalFileName);
     }
@@ -75,6 +76,10 @@ export class FileNode extends React.Component {
       newFileExtension &&
       oldFileExtension[0].toLowerCase() !== newFileExtension[0].toLowerCase()
     ) {
+      this.props.updateFileName(this.props.id, this.originalFileName);
+    }
+    //  Handles empty file names
+    if (!fileName.length) {
       this.props.updateFileName(this.props.id, this.originalFileName);
     }
   }
--- a/client/modules/IDE/components/FileNode.jsx
+++ b/client/modules/IDE/components/FileNode.jsx
@@ -67,7 +67,7 @@ export class FileNode extends React.Component {
   validateFileName() {
     const oldFileExtension = this.originalFileName.match(/\.[0-9a-z]+$/i);
     const newFileExtension = this.props.name.match(/\.[0-9a-z]+$/i);
-    const fileName = this.props.name.replace(/\.[^/.]+$/, '');
+    const fileNameWithoutExtension = this.props.name.replace(/\.[^/.]+$/, '');
     if (oldFileExtension && !newFileExtension) {
       this.props.updateFileName(this.props.id, this.originalFileName);
     }
@@ -79,7 +79,7 @@ export class FileNode extends React.Component {
       this.props.updateFileName(this.props.id, this.originalFileName);
     }
     //  Handles empty file names
-    if (!fileName.length) {
+    if (!fileNameWithoutExtension.length) {
       this.props.updateFileName(this.props.id, this.originalFileName);
     }
   }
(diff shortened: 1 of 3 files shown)


shakeabi:
@catarak This adds validation for the file and sketch names. However, it does not notify the user if the save fails. I think the user must be notified with some kinda error message. Currently, the error message is not being passed to the frontend from the server. The server just sends a {success: false} JSON response. So, should I add a custom error notification to be displayed to the user?

catarak:
i think that the file/sketch names should be automatically truncated!

shakeabi:
@catarak Currently, the sketch name is already being truncated. I'll add truncation to file names in the next commit. I'm talking about the error notification that should be sent if the name is of 0 in length...
Currently, sketch auto restores to their previous value if one attempts to give it a name 0 in length. 
Should we do something similar to file names too? However, the sketch name can still be set to null by manipulating the dom.

catarak:
ahhhhh didn't understand. yes, i think that if the user tries to enter an empty sketch or file name, it should be reset to the old one.

shakeabi:
@catarak I've added reset feature. Please review it..

client/modules/IDE/components/FileNode.jsx
```diff
@@ -67,6 +67,7 @@ export class FileNode extends React.Component {
   validateFileName() {
     const oldFileExtension = this.originalFileName.match(/\.[0-9a-z]+$/i);
     const newFileExtension = this.props.name.match(/\.[0-9a-z]+$/i);
+    const fileName = this.props.name.replace(/\.[^/.]+$/, '');
```
catarak: what does this regular expression do?

shakeabi: @catarak It gets the filename **without** the extension

catarak: got it! could you name this variable "fileNameWithoutExtension" so it's clear?

shakeabi: @catarak Sure. I've fixed it.

server/models/project.js
```diff
@@ -47,6 +65,35 @@ projectSchema.set('toJSON', {
 projectSchema.pre('save', function generateSlug(next) {
   const project = this;
   project.slug = slugify(project.name, '_');
+  project.name = project.name.substring(0, MAX_PROJECT_NAME_LENGTH); // Truncates project name
```
catarak: you should truncate the name BEFORE it's slugified.

catarak:
this seems to be working for me!

it would be great if you also added the front end file name length validation (you can do it in the same way as the project name, in which the `<input>` has a max length.

in order to test this in a way that wouldn't drive me insane, i changed `MAX_PROJECT_NAME_LENGTH`/`MAX_FILE_NAME_LENGTH` to 20, temporarily, which worked great. however, i noticed that, because on the FE the length limit was longer, a weird thing would happen where the project would save, and the name would be not-truncated until you refreshed the whole page. i'm not sure how to handle this, and i think it's worth thinking about, considering it's super easy to manually change the FE validation by editing the DOM in the developer tools.

shakeabi:
@catarak I've added the frontend validation. About the latter issue in your comment, yeah I realised it when fixing it. It is indeed very easy to manipulate frontend validation, however, this would be validated in the backend. But, as you said, that weird behaviour happens. One possible solution would be to refresh on every save. But, again that would be highly **not** user friendly.

shakeabi:
@catarak I believe this works fine. Please review it when you're free.

server/models/project.js
```diff
@@ -4,9 +4,25 @@ import slugify from 'slugify';
...
+    return false;
+  }
+
+  return true;
+};
+
 const fileSchema = new Schema(
   {
-    name: { type: String, default: 'sketch.js' },
+    name: {
```
catarak: if the pre 'save' function is running and truncating the file name, then why is a fileNameValidator also necessary?

shakeabi: @catarak That's a precaution. Let's say we change the update function later, this should not break right.

catarak: my concern is that this creates an error that is not being handled, so if you're going to keep it, i'd recommend also checking for this error (maybe something like https://mongoosejs.com/docs/validation.html#custom-validators). otherwise, i think it makes sense to take it out. also, if you're going to add this validator for the file name length, why not for a sketch name length as well?

shakeabi: @catarak well, this custom validator exists just for filename because there exist inbuilt min and max length validator for sketch name ( which I thought I added it but realised just now that it isn't there ). Why a custom validator for filename is that the length validation must be done on it **without considering the extension**.<br>
I don't clearly get the error that's not being handled( The validator returns an error telling that "filename validation failed" in the case of failure ). <br>
So either of the following can be done:
. remove the validation as you said earlier.
. add min and max validations for sketchname and min validation for filename

catarak: the validator does return an error; however, the error is not actually caught if it is thrown. instead, saving the file would just fail.

PR #1031, opened 2019-04-13
Fix race condition that's causing preview/console and code to be out of sync (#990)

### LINKED ISSUE #990: Inconsistency between the results and the code
#### Nature of issue?

- Found a bug

#### Details about the bug: 

- Web browser and version: Google Chrome 72.0.3626.121 (Official Build) (64-bit)
- Operating System: Mac OS X 10.13

1) Copy the following into the editor:
```
function setup() {
}
 
function draw() {
	console.log(1);
}
```
2) Run the code, verify logging works;
3) Do this as fast as possible: comment the line `console.log(1);` and run the code again;
BUG: 1s still get logged!
4) wait 1 second;
5) Do this as fast as possible: uncomment the same line and run the code again;
BUG: 1s no longer get logged although `console.log(1);` is uncommented!

My guess is that somehow the code editing/interpretation happens asynchronously and won't take effect until after a split second, thus when events happen at roughly the same time the orders got reversed.

Not sure if this is considered a bug but if so, I would be happy to help put up a fix, although cannot promise when that would happen.

ZiyaoWei:
This change attempts to fix #990. The cause seems to be a race condition - similar to #790 - where the state stored in redux is slightly behind the editor. This PR pulls the current file from the editor and use its content instead of the state in redux to compute the preview and console output.


ZiyaoWei:
Not sure if this is an okay fix, and any feedback are welcomed!

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -146,7 +146,10 @@ class PreviewFrame extends React.Component {
   injectLocalFiles() {
     const htmlFile = this.props.htmlFile.content;
     let scriptOffs = [];
-    const resolvedFiles = this.resolveJSAndCSSLinks(this.props.files);
+    const files = this.props.files.slice();
```
catarak: could you move this into a little helper function? i know this function is already long and messy, so i'm hoping to not make it worse 😈

catarak:
thanks for this! i think this is the right approach.

client/modules/IDE/components/PreviewFrame.jsx
ZiyaoWei: Updated! Sorry for the turnaround time. LMK if this isn't the right approach.

catarak:
i'm setting an error in the console:
```
webpack-internal:///99:20 Warning: Failed prop type: The prop `cmController` is marked as required in `PreviewFrame`, but its value is `undefined`.
    in PreviewFrame (created by IDEView)
    in IDEView (created by Connect(IDEView))
    in Connect(IDEView) (created by withRouter(Connect(IDEView)))
    in withRouter(Connect(IDEView)) (created by RouterContext)
    in div (created by App)
    in App (created by Connect(App))
    in Connect(App) (created by RouterContext)
    in RouterContext (created by Router)
    in Router (created by App)
    in Provider (created by App)
    in App (created by HotExportedApp)
    in AppContainer (created by HotExportedApp)
    in HotExportedApp
```
I think this is happening because there's a moment when the CodeMirror editor hasn't been instantiated yet. What I did in `Nav.jsx` (which also has a `cmController` prop) was make a default value for `cmController` that's an empty object.

ZiyaoWei:
Good catch, thanks! Hopefully this latest commit fixes that.

catarak:
cool, this is working great for me. merging!

PR #1044, opened 2019-04-18
Focus on find input

### LINKED ISSUE #1034: Focus on choosing option in find menu
#### Nature of issue?

- Found a bug
#### Details about the bug: 
Clicking on search specifiers causes the editor lose focus from the search box. 

Most of the editors retain focus in the search box on clicking these specifiers (eg. VS Code). We can have that too. @catarak  what are your views on it.

Screenshot (5)

nimishagarwal76:

*  has no linting errors (`npm run lint`)
* is from a uniquely-named feature branch and has been rebased on top of the latest master. (If I was asked to make more changes, I have made sure to rebase onto master then too)
* is descriptively named and links to an issue number, i.e. `Fixes #123`

Fixes #1034

### DIFF
--- a/client/utils/codemirror-search.js
+++ b/client/utils/codemirror-search.js
@@ -44,6 +44,11 @@ export default function(CodeMirror) {
     return cm.getSearchCursor(query, pos, getSearchState(cm).caseInsensitive);
   }
 
+  function isMouseClick(event) {
+    if(event.detail > 0) return true;
+    else return false;
+  }
+
   function persistentDialog(cm, text, deflt, onEnter, onKeyDown) {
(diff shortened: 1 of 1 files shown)


catarak:
rather than doing
```js
if (event.detail) {
```
could you instead do 
```js
if (event.detail > 0) {
```
or
```js
if (event.detail !== 0)
```
as i feel like it's clearer what's going on, with respect to https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail. or, make it into a function with a descriptive name. just something so it doesn't feel like magic.

nimishagarwal76:
Done !!

PR #1045, opened 2019-04-18
Fix for Save Indicator

### LINKED ISSUE #956: Unsaved file circle disappears on opening assets but file remains unsaved and changes still present.
#### Nature of issue?

- Found a bug

#### Details about the bug: 
The circle which displays if the file is unsaved disappears on opening My assets( or anything which prompts that you have unsaved files). According to me, the unsaved changes should either disappear or the circle should still be present because it's still not saved. Look into gif for better understanding.
ezgif-3-41c5130a6add

- Web browser and version:  On Chrome
- Operating System: MacOSX

If this is an issue, I'll resolve it.
Views @siddhant1 @catarak @GaurangTandon @plxity ?

nimishagarwal76:

*  has no linting errors (`npm run lint`)
*  is from a uniquely-named feature branch and has been rebased on top of the latest master. (If I was asked to make more changes, I have made sure to rebase onto master then too)
*  is descriptively named and links to an issue number, i.e. `Fixes #123`

Fixes #956 

@catarak  if this approach is fine I would add the changes to assets tab also.

### DIFF
--- a/client/modules/IDE/pages/IDEView.jsx
+++ b/client/modules/IDE/pages/IDEView.jsx
@@ -177,6 +177,10 @@ class IDEView extends React.Component {
       // don't warn
       this.props.persistState();
       window.onbeforeunload = null;
+    } else if (route && (/\/assets\/?$/.test(route.pathname))) {
+      // don't warn
+      this.props.persistState();
+      window.onbeforeunload = null;
     } else if (route && (route.pathname === this.props.ide.previousPath)) {
       // don't warn
       this.props.persistState();
--- a/client/modules/IDE/pages/IDEView.jsx
+++ b/client/modules/IDE/pages/IDEView.jsx
@@ -173,6 +173,14 @@ class IDEView extends React.Component {
       // don't warn
       this.props.persistState();
       window.onbeforeunload = null;
+    } else if (route && (/[\S]*\/sketches\/?$/.test(route.pathname))) {
+      // don't warn
+      this.props.persistState();
+      window.onbeforeunload = null;
+    } else if (route && (route.pathname === this.props.ide.previousPath)) {
+      // don't warn
+      this.props.persistState();
+      window.onbeforeunload = null;
     } else if (this.props.ide.unsavedChanges) {
       if (!window.confirm('Are you sure you want to leave this page? You have unsaved changes.')) {
         return false;


client/modules/IDE/pages/IDEView.jsx
```diff
@@ -173,6 +173,14 @@ class IDEView extends React.Component {
       // don't warn
       this.props.persistState();
       window.onbeforeunload = null;
+    } else if (route && (/[\s\S]*\/sketches\/?$/.test(route.pathname))) {
```
catarak: why `[\s\S]`? since it's a url, wouldn't there never be whitespace characters?

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -173,6 +173,14 @@ class IDEView extends React.Component {
...
+      // don't warn
+      this.props.persistState();
+      window.onbeforeunload = null;
+    } else if (route && (route.pathname === this.props.ide.previousPath)) {
```
catarak: what's this if statement checking?

catarak:
this seems to be working for me, though i'm a little hesitant as if there's a bug, it could potentially cause some folks to lose some unsaved work. how have you tested this? 

i guess if a user opens the "my assets" modal, it should also not warn about unsaved changes (which currently, this PR doesn't fix).

client/modules/IDE/pages/IDEView.jsx
nimishagarwal76: Now suppose you click on my sketches and you have ```unsavedChanges``` to be true so when you click to open another sketch name it will display a warning but if you click the cross button to close the modal no warning should be displayed. if condition is checking if you are going to the same page which you came from.

nimishagarwal76:
> this seems to be working for me, though i'm a little hesitant as if there's a bug, it could potentially cause some folks to lose some unsaved work. how have you tested this?
> 
> i guess if a user opens the "my assets" modal, it should also not warn about unsaved changes (which currently, this PR doesn't fix).

Yeah I have tested it, I think it is working fine. I tried all the possible combinations.

client/modules/IDE/pages/IDEView.jsx
nimishagarwal76: ok i will remove \s
convert it to
```
/[\S]*\/sketches\/?$/
```

client/modules/IDE/pages/IDEView.jsx
catarak: got it! that makes sense.

catarak:
okay, i think the only thing to add is also checking for the route `/<username>/assets`, since it's an overlay like the sketch list.

nimishagarwal76:
> okay, i think the only thing to add is also checking for the route `/<username>/assets`, since it's an overlay like the sketch list.

Done, the thing is assets takes you to route ``/assets`` rather than ``/<username>/assets`` so I made changes accordingly. 

But,
Screenshot (6)
cause of this bug when you click cross button no warning is displayed, but if you click the same file name that is opened in editor it will show warning as username gets changed to undefined

catarak:
hmm, I am able to reproduce the issue you're having. i'm not sure what the best solution is—what do you think? maybe it's okay to leave for now, since this fixes the most common times it comes up. is there a bigger issue here, which is that the modals shouldn't change the route? should this be handled with hashes?

nimishagarwal76:
I didn't get it, are you suggesting that we change route some thing like ```/user/sketches/sketchId#assets``` just like github does when you select a line. We can also show modals without changing any route also (Something like bootstrap modal does).

catarak:
> I didn't get it, are you suggesting that we change route some thing like `/user/sketches/sketchId#assets` just like github does when you select a line. We can also show modals without changing any route also (Something like bootstrap modal does).

yeah, something like that! i don't what to commit to that _exact_ route, but i wonder if using routes for modals is the right approach, and as you suggest, the route shouldn't change at all. though, we'd have to be mindful of not breaking existing urls. i'm open to suggestions.

PR #1046, opened 2019-04-18
Fixes #760

### LINKED ISSUE #760: userid is email when signed up with google
This is likely a duplicate of #655 but I thought I would file separately since there are other bugs discussed in that issue. At creative coding fest, we noticed that if a user signs up with their google account, their username is their e-mail. This is fine for logging in, but it ends up in the path of their sketch which is a privacy concern.

```
https://editor.p5js.org/emailaddress@gmail.com/sketches/#######
```

Perhaps there is a way we can allow a user to set a custom "user id" when logging in with google. (Note when logging in with GitHub it defaults to the GitHub username.)

catarak:

### DIFF
--- a/server/config/passport.js
+++ b/server/config/passport.js
@@ -1,6 +1,6 @@
-import User from '../models/user';
 import slugify from 'slugify';
 import friendlyWords from 'friendly-words';
+import User from '../models/user';
 
 const lodash = require('lodash');
 const passport = require('passport');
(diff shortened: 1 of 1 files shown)


siddhant1:
Hey , this code is still using callbacks.
Shouldn't we merge promise based code as discussed in a previous pr?

catarak:
i'm going to merge this and see what the response is—if it's too confusing i'm happy to revert it!

PR #1050, opened 2019-04-21
Responsiveness for settings overlay

### LINKED ISSUE #1036: Responsiveness for settings overlay
#### Nature of issue?

- Found a bug

#### Details about the bug: 

The settings overlay is not responsive for both landscape and portrait view, 
 - The solution for portrait view might be to increase width of the overlay for mobile devices or to keep the general settings and accessibility tab as column display for small screens
- The solution for landscape view would be to set a minimum or fixed height
overlay_1
overlay_2

Shubhanshu88:
I have verified that this pull request: Fixes #1036 

### DIFF
--- a/client/styles/components/_overlay.scss
+++ b/client/styles/components/_overlay.scss
@@ -22,8 +22,14 @@
   display: flex;
   flex-wrap: wrap;
   flex-flow: column;
-  max-height: 80%;
-  max-width: 65%;
+  max-width: 100%;
+  max-height: 100%;
+  overflow-y: auto;
+  @include media-breakpoint-up($sm) {
+    overflow-y: unset;
+    max-width: 65%;
+    max-height: 80%;
+  }
 }
 
 .overlay__header {


Shubhanshu88:
overlay_responsive
 - If the height is set to 100% always, it may leave too much space between the elements and bottom of screen, although it can be easily set.
 - Also, for landscape view, a media query for height can be added where if height is less than smallest media breakpoint, the overflow would be set to scroll again
- I have unset the overflow property for larger widths as I observed a blur effect on the overlay for larger screens

shakeabi:
@Shubhanshu88 Aah I meant 100% width earlier. I believe the current version look better than the overlay spanning 100%. But I can't  identify what causes that blur effect...

Shubhanshu88:
@shakeabi I think it works fine now, also earlier some elements were overflowing the overlay especially in applications tab.

Most probably the blur is a chrome specific bug, I did not notice it in firefox

catarak:
this PR is dependent on the breakpoints being added—maybe it makes sense to create a PR that simply adds the mixins and variables for the responsive styles?

plxity:
@catarak Hey, I think Let's close this PR as there is no update and If possible I will work on this. :)

PR #1059, opened 2019-05-02
Fixed folder can be renamed to empty string "" #989

### LINKED ISSUE #989: Folder can be renamed to "" empty string

olalekanteeblaze:

* [x] has no linting errors (`npm run   lint`)
* [x] is descriptively named and links to an issue number, i.e. `Fixes #989`

### DIFF
--- a/client/modules/IDE/actions/files.js
+++ b/client/modules/IDE/actions/files.js
@@ -5,7 +5,6 @@ import { reset } from 'redux-form';
 import * as ActionTypes from '../../../constants';
 import { setUnsavedChanges } from './ide';
 import { setProjectSavedTime } from './project';
-import { showToast, setToastText } from './toast';
 
 const __process = (typeof global !== 'undefined' ? global : window).process;
 const ROOT_URL = __process.env.API_URL;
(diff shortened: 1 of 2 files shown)


catarak:
thanks for your work on this! i don't think this is the right solution. rather than showing a toast, i think this should be a form error, like when you're signing up for a website and you put in a username that already exists.

olalekanteeblaze:
Thanks. I think i know what to do now.

olalekanteeblaze:
Hello @catarak , do you think doing a sort of validation in the onChange handler fix this.

PR #1063, opened 2019-05-04
Fixed folder can be renamed to empty string "" #989

### LINKED ISSUE #989: Folder can be renamed to "" empty string

olalekanteeblaze:

* [x] is descriptively named and links to an issue number, i.e. `Fixes #989 `

### DIFF
--- a/client/modules/IDE/components/FileNode.jsx
+++ b/client/modules/IDE/components/FileNode.jsx
@@ -55,7 +55,9 @@ export class FileNode extends React.Component {
   }
 
   handleFileNameChange(event) {
-    this.props.updateFileName(this.props.id, event.target.value);
+    if (event.target.value.length) {
+      this.props.updateFileName(this.props.id, event.target.value);
+    }
   }
 
   handleKeyPress(event) {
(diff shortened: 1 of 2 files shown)


catarak:
there's a function in `FileNode.jsx` called `validateFileName`, which should be updated, rather than `handleFileNameChange`

olalekanteeblaze:
I've updated it. I think adding a length cap should fix it

client/modules/IDE/components/FileNode.jsx
```diff
@@ -77,6 +77,9 @@ export class FileNode extends React.Component {
     ) {
       this.props.updateFileName(this.props.id, this.originalFileName);
     }
+    if (this.props.name.length) {
```
catarak: i think this needs to come first in the `if` statements. also i think the logic is messed up here—it should be `if` and `if else` statements here so the file name doesn't get updated multiple times.

catarak:
i just checked out this branch and tested it, and i'm still able to rename a folder to a empty string.

Issue #392, opened 2017-07-11
Update Account Settings View


catarak:
<img width="515" alt="screen shot 2017-07-11 at 2 08 26 pm" src="https://user-images.githubusercontent.com/6063380/28083149-72527052-6642-11e7-9acf-1b7125d13e7c.png">

View more on [Zeplin](https://zpl.io/Z2cihf1).

himanshuc3:
Is there any route currently that leads to My Account settings?

catarak:
yes, `/<username>/settings`.

himanshuc3:
do you mean revamping `/:username/account`. 
There is no current path `/:username/settings` defined.

catarak:
oops, yes!

LakshSingla:
Since this has been dead for a long time, may I take up this issue?

Just to be clear, the objective is to redesign the following view (the input text fields have black background due to OS specific settings)...
![Screenshot from 2019-03-27 00-49-01](https://user-images.githubusercontent.com/30999375/55027003-3506d700-502a-11e9-98ed-80cb3f592e07.png)

catarak:
@LakshSingla please do!

PR #1075, opened 2019-05-14
Adds Personal Access Tokens for API access (fixes #541)

### LINKED ISSUE #541: API authentication with Token (or oauth?)
#### Nature of issue?

- [X] New feature request

#### New feature details:

if users can authentication via a token or oauth, then they can upload sketches programmatically. and maybe other fun stuff i can't think of at the moment.

andrewn:
Implements Personal Access Tokens for API access. This builds on top of the excellent work in #731 to allow access tokens to be created by users and used to authenticate instead of a password.

I plan on making the following changes from #731:

 - [x] Move token generation to the server-side from the client. This seems to be how other providers do this (Digital Ocean, Github, Harvest).
- [x] Displays tokens as part of Account page (moving from the Advanced page)
- [x] Write UserController tests for creating/deleting keys
- [x] Improve display of tokens in UI
- [ ] Write UI tests **<- I'll do these in a future PR**

## UI

I've added tabs to the Account screen. The _Access Tokens_ tab allow tokens to be created and deleted.

User interface for managing Access Tokens

## Automated testing

I've written tests for the controller endpoints but created Mongoose model mocks by hand. This isn't ideal so I'd like to re-visit these in a future PR once we've figured out a good way forward for testing.

## Manual test plan

1. Create a new access token at /account
2. Copy access token and use `curl` to access the test endpoint:
    `curl -X GET --user <username>:<access-token> -i localhost:8000/api/auth/access-check`
    If successful, the user's session data should be returned and _Last used_ time in UI should update.
3. Change username, access should be forbidden
4. Change token, access should be forbidden
5. Delete token via UI
6. API access should be forbidden


server/models/user.js
```diff
@@ -22,6 +40,7 @@ const userSchema = new Schema({
   github: { type: String },
   email: { type: String, unique: true },
   tokens: Array,
```
andrewn: @catarak I was thinking about renaming `apiKeys` on the line below to `accessTokens` but then saw this `tokens` array. It's where we store the [accessTokens from Github/Google](https://github.com/processing/p5.js-web-editor/blob/master/server/config/passport.js#L89) when a user does a social login. 

Do you know what we use these for? I guess it's a good idea to keep the naming distinct to avoid confusion? I briefly thought that we might put the Personal Access Tokens in this array too, but maybe that's a bad idea.

catarak: I went back to the code I used as reference for setting up passport (https://github.com/sahat/hackathon-starter), and I think the tokens are supposed to be used for linking/unlinking social accounts. So, maybe, this access token should be put into that array? I don't know. I think the linking/unlinking social accounts isn't fully fleshed out anyway.

andrewn: 🤔 

So the `tokens` are keys the app has been given by external services. And `apiKeys` will be keys that the app has given to others. In which case, perhaps they should be separate.

catarak: that makes sense to me!

client/styles/components/_api-key.scss
```diff
@@ -0,0 +1,102 @@
...
+  th {
+    padding: #{5 / $base-font-size}rem;
+  }
+
+  td {
+    padding: #{15 / $base-font-size}rem #{5 / $base-font-size}rem;
+  }
+
+  tbody tr:nth-child(odd) {
+    background-color: #f2f2f2;
```
catarak: could this use a color in the theme map rather than a raw color value?

catarak:
Some other stuff I noticed:
- I created a new account, and then immediately opened "Settings -> API Tokens" and the application crashed. It worked totally fine after I refreshed the page.
- Extremely small UI fix, could use some space between these two elements (same as the form labels/inputs):
Screen Shot 2019-05-21 at 4 47 48 PM
- Just... thinking about the overall design of this page. The border for the tabs feels a little funky, though I'm not sure how to change it.

catarak:
but i ran through your test plan and it's working for me!

client/styles/components/_api-key.scss
andrewn: The account page doesn't respect the theme, but it probably should?

andrewn: Ok, I've started work on this in #1080.

andrewn:
> * Just... thinking about the overall design of this page. The border for the tabs feels a little funky, though I'm not sure how to change it.

I know, I'm not very happy with the layout of the page. There's a design for it in #392 but I'm not convinced that it improves things much as none of the fields look editable.

The tabs here don't really work in the full-width page, but we could adopt a side-navigation like GitHub does?

Screenshot 2019-05-22 11 22 45

andrewn:
@catarak, I made those fixes so this is ready for a re-review.

> * I created a new account, and then immediately opened "Settings -> API Tokens" and the application crashed. It worked totally fine after I refreshed the page.

I fixed this in 5777be3 by ensuring the API consistently returns `apiKeys` in the user-shaped response.

> * Extremely small UI fix, could use some space between these two elements (same as the form labels/inputs)

Added some padding around the button.

catarak:
I was just thinking about merging this to master and then this being available in production—maybe it makes sense to merge it to another branch until this is built out more? Maybe it could be visible by a secret url or something? It's hard since this project doesn't have a staging deployment, though it could!

catarak:
one option for the design, that would keep the tabs at the top, would be to wrap the settings container in a box (e.g. like the [login design here](https://app.zeplin.io/project/55f746c54a02e1e50e0632c3/screen/59413d7cfdc65d2dbd1c3218)) so that the tab strip wouldn't be floating.

andrewn:
> I was just thinking about merging this to master and then this being available in production—maybe it makes sense to merge it to another branch until this is built out more? Maybe it could be visible by a secret url or something? It's hard since this project doesn't have a staging deployment, though it could!

That's a very good point. I've created a new branch `feature/public-api` for this purpose and changed the base branch of this PR.

I'll create a new PR for merging `feature/public-api` into `master` and use that to track the status of the API work.

It'd be great to have a deployment of this branch for testing. How much effort is that do you think? 

Alternatively, we could deploy to production and hide behind [feature flags](https://www.martinfowler.com/articles/feature-toggles.html).

catarak:
> It'd be great to have a deployment of this branch for testing. How much effort is that do you think?

I think, with the power of CI/CD and Kubernetes, it's not too hard. What I'd need to do:
- [ ] add `staging_deploy` script/task to Travis, which will be extremely similar to `deploy.sh`
- [ ] update Kubernetes cluster config with new deployment/namespace for staging
- [ ] add new repository on Dockerhub for staging image
- [ ] add these variables to Travis

I might wait to start on this until we merge in your `.travis.yml` changes, so we don't have to deal with conflicts there.

andrewn:
Now that  #1081 is merged, this is ready too. 👍

catarak:
> Now that #1081 is merged, this is ready too. 👍

great, i can work on setting up the staging deployment this week!

trych:
Whatever happened to the API access features? Is there any plan to implement them?
The `public-api` branch doesn't seem to exist anymore?

catarak:
@trych the API has been implemented but it hasn't been released yet, but there are plans for this soon. Stay tuned :)

trych:
@catarak 

One more question regarding the API:
I want to include some sketches in a blog-like article via the embed feature, so I am dealing with iframes. I am having trouble to automatically get the height of my sketch, as this is of course a cross-origin scenario, so I cannot read the canvas' height via JS.

Is this something that could/would be solved by the API? As in: I would be able to access the sketch height via the API? Or are there even any other solutions for that, that I am missing?

(I could also open a separate issue for that, if that helps)

catarak:
@trych great question! I think this is a separate issue, so I'm going to open one and answer your question there 😄

Issue #1073, opened 2019-05-13
Specifying nested file trees without IDs in the API


andrewn:
#### Nature of issue?

- [x] Existing feature enhancement

#### Feature enhancement details:

As part of the _Sketch upload_ API work, we want a nicer way of specifying file trees.

Below, "file" means file or directory, as they're mostly the same for this discussion.

#### Current approach

The current API shape closely mirrors the shape of the data stored in mongodb:
- requires BSON-object ids to be generated by clients
- files are specified as a flat array of all files
- hierarchy/nesting is defined by the file IDs in a `children` array on the relevant object

<details>
 <summary>Example</summary>

```json
{
  "name": "Summer accordion",
  "updatedAt": "",
  "isSaving": false,
  "files": [
    {
      "name": "root",
      "id": "5cd98108d498ade50d793853",
      "_id": "5cd98108d498ade50d793853",
      "children": [
        "5cd98108d498ade50d793850",
        "5cd98108d498ade50d793851",
        "5cd98108d498ade50d793852",
        "5cd9810dd498ade50d793854"
      ],
      "fileType": "folder",
      "content": "",
      "isSelectedFile": false
    },
    {
      "name": "sketch.js",
      "content": "function setup() {\n  createCanvas(400, 400);\n}\n\nfunction draw() {\n  background(220);\n}",
      "id": "5cd98108d498ade50d793850",
      "_id": "5cd98108d498ade50d793850",
      "isSelectedFile": false,
      "fileType": "file",
      "children": []
    },
    {
      "name": "index.html",
      "content": "<!DOCTYPE html>\n<html>\n  <head>\n    <script src=\"https://cdnjs.cloudflare.com/ajax/libs/p5.js/0.7.3/p5.min.js\"></script>\n    <script src=\"https://cdnjs.cloudflare.com/ajax/libs/p5.js/0.7.3/addons/p5.dom.min.js\"></script>\n    <script src=\"https://cdnjs.cloudflare.com/ajax/libs/p5.js/0.7.3/addons/p5.sound.min.js\"></script>\n    <link rel=\"stylesheet\" type=\"text/css\" href=\"style.css\">\n    <meta charset=\"utf-8\" />\n\n  </head>\n  <body>\n    <script src=\"sketch.js\"></script>\n  </body>\n</html>\n",
      "id": "5cd98108d498ade50d793851",
      "_id": "5cd98108d498ade50d793851",
      "fileType": "file",
      "children": [],
[...]

Issue #542, opened 2018-02-12
Define API/Spec for batch uploading sketches


catarak:
#### Nature of issue?

- [X] New feature request

#### New feature details:

Define endpoints so that users can batch upload sketches. This also includes the data structure of the sketches themselves. It's a little complicated right now given recursive file structures of folders and this involves generating ID's for file objects. It would be cool if users didn't have to do that and it happened magically on the backend.

Vertmo:
Is the API being worked on right now ? That's seems like a bit of work, so it should probably be split in several parts; I'll start looking into it anyway ^^

catarak:
You're right, this is a huge issue! I think a good first step would be dividing this into smaller issues, figuring out all of the steps that would need to go into this. We can discuss on this issue!

Vertmo:
Ok, so the first thing would be to allow registered users to add API keys via their user settings, and of course store them in the database.
Then we will need to define secure authentification for the API (maybe we should use an already existing library).
We could then define two endpoints :
* `/<username>/upload` to upload a single sketch
* `/<username>/batch-upload` to upload several
Of course, only <username> would be able to successfully upload a sketch to their account.
OK, that's only my first thoughts. Then we will need to work out the format of the upload request (especially how to upload a whole folder). But I'll need to study the existing code more thoroughly before that !
In the meantime, can you tell me what you think of these first ideas ?

catarak:
The first part is covered in #541! 

My starting point for figuring out what API endpoints should be is, surprisingly, [a fairly old Rails guide](https://guides.rubyonrails.org/routing.html). Therefore, I think the endpoint for uploading a new sketch should be POST `/<username/sketches`. Which is the same as what the editor currently uses. Maybe for multiple sketches, it could be the same endpoint, but the API could check if the data is an array vs a singular project.

Also, to list some other stuff I think needs to happen:
- right now, the Node server that serves the client application, and the API are smushed together. I think that part of this work would be to separate them into two separate Node applications. We could even think about creating two different Docker images, and change the Kubernetes deployment to handle this. 
[...]

PR #1076, opened 2019-05-16
Public API for Sketch management

andrewn:
This is a proposal for a public-facing API that would allow users to automate the upload of sketches  into the editor, replacing the scripts that talk to mongodb directly. (See #542)

## Supported actions

The key actions for this API are:
- creating a sketch as a user
- adding files to the sketch

The scripts deleting existing sketches, but I've also proposed a way to replace the contents of a sketch too.

I've added the [proposed API documentation](https://github.com/processing/p5.js-web-editor/pull/1076/files) as a markdown file so that we can comment on the file directly.

## Files instead of IDs

The key change (as discussed in #1073) is representing trees hierarchically and referencing files/directories via their path. This seems more intuitive for API users, but some investigation is needed figure out if we still need to reference files by ID. And if so, perhaps the filename approach is only used as a convenience for sketch creation.

## Further work

- [x] **File storage:** I need to dig in and understand a little better how file storage currently works as the `files` returned by the API might need to be redirects to S3 or other hosting. And this would impact how the files are specified also.
- [ ] **Conflicts and timestamps:** The editor uses a `updatedAt` timestamp to protect against overwriting stale projects. These should probably be added everywhere.

### DIFF
--- a/developer_docs/public_api.md
+++ b/developer_docs/public_api.md
@@ -16,6 +16,16 @@
 - Operations are transactional, e.g. if a file is somehow invalid, the sketch won’t be left partially uploaded
 - All requests send and receive `Content-Type: application/json` unless otherwise stated
 
+# Versioning
+
+The API is versioned and this version is indicated in the root URL path e.g. version 1 of the API can be found at `http://editor.p5js.org/api/v1`.
+
+You must provide the version number when accessing the API.
+
+| Version | Release date |
+| ------- | ------------ |
+| v1      | Unreleased   |
+
 # Models
 
 The API accepts and returns the following model objects, as JSON.
--- a/developer_docs/public_api.md
+++ b/developer_docs/public_api.md
@@ -36,8 +36,7 @@ The API accepts and returns the following model objects, as JSON.
 
 ### Validations
 
-- `files` must have exactly one top-level file with the `.html` extension
-    - We could create a default `index.html` if not provided?
+- `files` must have exactly one top-level file with the `.html` extension. If none is provided, then a default `index.html` and associated `style.css` will be automatically created.
 
 ## DirectoryContents
 
@@ -107,6 +106,8 @@ No body.
 ## `POST /:user/sketches`
 
 Create a new sketch.
+
+A sketch must contain at least one file with the `.html` extension. If none if provided in the payload, a default `index.html` and linked `style.css` file will be created automatically.
 
 ### Request format
 See `Sketch` in Models above.
--- a/developer_docs/public_api.md
+++ b/developer_docs/public_api.md
@@ -22,21 +22,25 @@ The API accepts and returns the following model objects, as JSON.
 
 ## Sketch
 
-| Name  | Type              | Description                                                                          |
+| Name  |       Type        |                                     Description                                      |
 | ----- | ----------------- | ------------------------------------------------------------------------------------ |
 | name  | String            | The sketch’s title                                                                   |
 | files | DirectoryContents | The files and directories in this sketch. See `DirectoryContents` for the structure. |
+| slug  | String            | A path that can be used to access the sketch                                         |
 
 
     {
       "id": String, // opaque ID
       "name: String,
-      "files": DirectoryContents
+      "files": DirectoryContents,
+      "slug": String // optional
     }
 
 ### Validations
 
 - `files` must have exactly one top-level file with the `.html` extension. If none is provided, then a default `index.html` and associated `style.css` will be automatically created.
+- `slug` must be an URL-safe string
+- `slug` must be unique across all user's sketches
 
 ## DirectoryContents
 
@@ -151,10 +155,10 @@ See `Sketch` in Models above.
 
 ### Responses
 
-| HTTP code                | Body                                         |
-| ------------------------ | -------------------------------------------- |
-| 201 Created              | id of sketch                                 |
-| 422 Unprocessable Entity | file validation failed, unsupported filetype |
+|        HTTP code         |                               Body                                |
+| ------------------------ | ----------------------------------------------------------------- |
+| 201 Created              | id of sketch                                                      |
+| 422 Unprocessable Entity | file validation failed, unsupported filetype, slug already exists |
 
 
 ### Examples
@@ -182,6 +186,7 @@ Returns `Sketch`.
     
     {
       "name": "Another title",
+      "slug": "example-1",
       "files": {
         "index.html": { "<DOCTYPE html!><body>Hallo!</body></html>" },
         "something.js": { "var uselessness = 12;" }
--- a/developer_docs/public_api.md
+++ b/developer_docs/public_api.md
@@ -1,19 +1,18 @@
 # Public API
 
-⚠️: This is currently a proposal, not an API that exists.
+This API provides a way to programmatically import data into the p5.js Web Editor.
 
 # Authentication
 
-- Access to the API is available via a Personal Access Token, linked to an existing editor user account
-- Created via a logged-in user’s Settings page
-- Tokens can be deleted via the Settings page
-- Tokens are presented to the API using basic auth
-- Tokens can also be sent via the `Authorization: Bearer {your_access_token}` HTTP header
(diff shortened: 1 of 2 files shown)


developer_docs/public_api.md
```diff
@@ -0,0 +1,408 @@
...
+    {
+      "id": String, // opaque ID
+      "name: String,
+      "files": DirectoryContents
+    }
+
+### Validations
+
+- `files` must have exactly one top-level file with the `.html` extension
+    - We could create a default `index.html` if not provided?
```
catarak: i like this idea! maybe it also makes sense to add the default css as well?

catarak:
Overall I like this! Some thoughts/questions/etc..:
- Should there be a route to add a new file to a sketch? This could possibly also function as a route to upload media files. 
- What's your thinking behind having the PUT and PATCH routes for completely replacing a sketch vs updating it? What's a scenario in which you think this would be useful? Is it so a user can update a sketch without changing its url? There's a secret feature of the editor which doesn't have a frontend UI yet, in which you can access a sketch via a slug url. I added this for the release of the [Generative Design](http://www.generative-gestaltung.de/2/) book, (e.g. https://editor.p5js.org/generative-design/sketches/P_1_0_01 works!). I don't know if this would add to the API at all, but just throwing that information out there.
- With respect to media file uploads, my suggestion is:
`POST /:user/files/upload`, which accepts
```JS
{
  name: "dog.jpg",
  content: <file_content>
}
```
and returns
```JS
{
  name: "dog.jpg",
  url: <new_s3_url>
}
```
This could also accept a query parameter that's `uploadType=media|multipart` for multipart uploads. The URL could then be used as an attribute on a file. This gets a little hairy if a user then decides to use the URL for multiple files/sketches—right now, if a user deletes a sketch, then the file gets deleted from S3. This makes sense, since in the web editor, you can't manually match a file to a url, it's all done behind the scenes for a user. 

Another solution would be to only allow media uploads after a sketch has been created, and then use a new file route to upload the media file to the sketch itself.

- Now, timestamps: it feels like asking users to keep track of `updatedAt` dates is a lot. I looked at how the [Google Docs API](https://developers.google.com/docs/api/how-tos/move-text) handles this, and it uses transactional add text/delete text commands. Maybe, at a certain point, we can't protect users from themselves. I don't know. I wonder how other APIs handle this.

developer_docs/public_api.md
andrewn: Added to the document.

andrewn:
> - Should there be a route to add a new file to a sketch? This could possibly also function as a route to upload media files.

Yeah, there should be. Something like `PUT /:user/sketches/:id/files/:path` would create a new file/directory if it didn't exist. Or replace it if it did. I wonder whether [this route](https://github.com/processing/p5.js-web-editor/pull/1076/files#diff-1117482daa790df7625a9a5b644034c6R326) should just be replaced with `PUT`?

andrewn:
> - What's your thinking behind having the PUT and PATCH routes for completely replacing a sketch vs updating it? What's a scenario in which you think this would be useful? Is it so a user can update a sketch without changing its url? 

Yep, that's part of it. Both `PATCH` and `PUT` would maintain the URL of the sketch being changed. The difference is `PUT` would remove everything first. And `PATCH` would add new things, keeping what's there, possibly changing the file contents. I do wonder if supporting these two methods for modifying the file tree is a bit confusing though and would be quite challenging to implement.

The way the current import scripts work, they delete all sketches and then re-upload them, creating new sketch IDs. This means all existing sketch links are broken. Maybe that's desired, but it might be nice for the author to choose which behaviour they want.

> There's a secret feature of the editor which doesn't have a frontend UI yet, in which you can access a sketch via a slug url. I added this for the release of the Generative Design book, (e.g. https://editor.p5js.org/generative-design/sketches/P_1_0_01 works!). I don't know if this would add to the API at all, but just throwing that information out there.

I think the slug URL is a great feature and would remove the need for supporting sketch ID stability above. For v1 of the public API, shall we just support specifying a slug, but and not support `PUT`/`PATCH` for sketches? This would be much simpler.

I think `PATCH` might be necessary if we want to eventually support clients like the Editor. Although the lower-level [Files](https://github.com/processing/p5.js-web-editor/pull/1076/files#diff-1117482daa790df7625a9a5b644034c6R296) endpoints would allow better control over file operations anyway.

andrewn:
For file uploads, I think having files' _belong to_ sketches makes things easier as it means we can be certain when a sketch is deleted then all the associated files can be safely removed from S3. And also, when a file is deleted from a sketch, it won't be in use by other sketches so can again be deleted from S3.

I'll prototype this out in code so we can see how it works with the upload scripts.

andrewn:
> Now, timestamps: it feels like asking users to keep track of updatedAt dates is a lot. I looked at how the Google Docs API handles this, and it uses transactional add text/delete text commands. Maybe, at a certain point, we can't protect users from themselves. I don't know. I wonder how other APIs handle this.

I agree that a user managing timestamps seems a bit much. I'll do some more research in how different APIs handle this. For the bulk-import use case, we can assume that a user creating or modifying a resource within the sketch—either files or the sketch itself—updates an `updatedAt` timestamp returned by the API.

andrewn:
Thanks for the comments!

We can keep this open for more discussion and I'll start writing some code to implement a basic version of these endpoints.

catarak:
> I think the slug URL is a great feature and would remove the need for supporting sketch ID stability above. For v1 of the public API, shall we just support specifying a slug, and not support `PUT`/`PATCH` for sketches? This would be much simpler.

great! i'll make sure there's an index in the db for the slug key. 

> I think `PATCH` might be necessary if we want to eventually support clients like the Editor. Although the lower-level [Files](https://github.com/processing/p5.js-web-editor/pull/1076/files#diff-1117482daa790df7625a9a5b644034c6R296) endpoints would allow better control over file operations anyway.

do you mean if folks want to use the API in the context of another web application/GUI? like maybe we don't even want to suggest that could be possible right now, as it adds a huge layer of complication. i think we want to focus on moving sketches to/from the web editor, rather than adding a new UI for editing sketches.

andrewn:
> > I think `PATCH` might be necessary if we want to eventually support clients like the Editor. Although the lower-level [Files](https://github.com/processing/p5.js-web-editor/pull/1076/files#diff-1117482daa790df7625a9a5b644034c6R296) endpoints would allow better control over file operations anyway.
> 
> do you mean if folks want to use the API in the context of another web application/GUI? like maybe we don't even want to suggest that could be possible right now, as it adds a huge layer of complication. i think we want to focus on moving sketches to/from the web editor, rather than adding a new UI for editing sketches.

I was thinking that ultimately we might just have one API for the editor and public uses. But you're right, that overcomplicating things for now.

andrewn:
I've realised that the API needs to support referencing files hosted elsewhere on the internet. For example, the ml5 examples reference models hosted in GitHub via jsdelivr.

I've [split the file into two to support this](https://github.com/processing/p5.js-web-editor/pull/1076/commits/2eeab30489d98ae8212e645afbd8593c8f059708). What do you think @catarak?

developer_docs/public_api.md
```diff
@@ -0,0 +1,433 @@
...
+# Models
+
+The API accepts and returns the following model objects, as JSON.
+
+## Sketch
+
+| Name  |       Type        |                                     Description                                      |
+| ----- | ----------------- | ------------------------------------------------------------------------------------ |
+| name  | String            | The sketch’s title                                                                   |
+| files | DirectoryContents | The files and directories in this sketch. See `DirectoryContents` for the structure. |
```
andrewn: The more I think about it, I'm not sure this should be called `files` since it can contain both files and directories. I can't think of a better name though.

- `contents` but then it's easily typo'd with `File.content`
- `folder`
- `items`

Any ideas @catarak?

catarak: hmmmm! i think my reasoning behind `files` is the unix reasoning behind files, where both directories and text/media files are files. out of the ones you've listed, `files` to me is still the one i like the most.

andrewn:
I've just split this into two documents, `public_api.md` and `public_api_proposed.md` to make it clear what is being implemented in v1 and what is just ideas.

Will merge this into the `feature/public-api` branch now. We can have further discussions in new issues or PRs.

PR #1080, opened 2019-05-22
Theme aware user pages

andrewn:
~⚠️  **Requires #1075 to be merged first before review**~

Currently, "user" pages like Log In, Sign Up, Account Settings and re-usable like `form` and `form-container` are not theme-aware.

This makes it difficult to reuse CSS across the app. For example, the striped tables in #1075 are theme-aware, but the colours are wrong when embedded in the Account page.

This PR:
* [x] Convert pages to be theme-aware
  - Account
  - Email verification
  - Log in
  - Sign up
  - Password reset
  - New password
* [x] Make forms theme-aware
* [x] Use theme variables for table stripes across app
* [x] Implement [boxed design](https://app.zeplin.io/project/55f746c54a02e1e50e0632c3/screen/59413d7cfdc65d2dbd1c3218)
* [ ] Reduce CSS variables where possible

### DIFF
--- a/client/components/NavBasic.jsx
+++ b/client/components/NavBasic.jsx
@@ -1,10 +1,15 @@
+import PropTypes from 'prop-types';
 import React from 'react';
 import InlineSVG from 'react-inlinesvg';
 
 const logoUrl = require('../images/p5js-logo-small.svg');
+const arrowUrl = require('../images/triangle-arrow-left.svg');
 
 class NavBasic extends React.PureComponent {
-  
+  static defaultProps = {
+    onBack: null
+  }
+
   render() {
     return (
       <nav className="nav" title="main-navigation" ref={(node) => { this.node = node; }}>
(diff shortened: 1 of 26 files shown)


andrewn:
[@catarak suggested](https://github.com/processing/p5.js-web-editor/pull/1075#issuecomment-494925133):

> one option for the design, that would keep the tabs at the top, would be to wrap the settings container in a box (e.g. like the [login design here](https://app.zeplin.io/project/55f746c54a02e1e50e0632c3/screen/59413d7cfdc65d2dbd1c3218)) so that the tab strip wouldn't be floating.

andrewn:
@catarak I think this is looking much better. Do you think we need a background image as in the design? I can't seem to extract that one from the Sketch file.

Screenshot 2019-05-24 11 54 37

catarak:
> @catarak I think this is looking much better. Do you think we need a background image as in the design? I can't seem to extract that one from the Sketch file.

I think the idea behind that design is that there would be a p5.js sketch running in the background, like the p5.js website. I think it makes sense not to include that for the settings page!

andrewn:
Ah, ok. Yeah, I think it would be distracting to have all sounds and movement on these pages. 

I'll leave it blank for now as I don't want to spend too much more time on this.

catarak:
This is looking good! A few things:
- I think the "X" to close the page should still be at the upper right-hand corner, rather than within the inner pane
- I think the background color of the page should match the theme

andrewn:
I just cherry-picked the commits on top of `feature/public-api` so it's easier to review.

As I force-pushed, you'll probably want to delete any local copies of this branch to avoid things getting weird.

andrewn:
I just cherry-picked the commits on top of `feature/public-api` so it's easier to review.

As I force-pushed, you'll probably want to delete any local copies of this branch to avoid things getting weird.

andrewn:
>  * I think the "X" to close the page should still be at the upper right-hand corner, rather than within the inner pane

The designs have  [Back to the editor](https://scene.zeplin.io/project/55f746c54a02e1e50e0632c3/screen/59413d8dd21d1d37bd2618e0) call out instead of an "X" so I implemented this. 

What do you think?

Screenshot 2019-06-04 14 57 11

> * I think the background color of the page should match the theme

👍  ✅

catarak:
i think this looks a lot better, and is ready to merge!

PR #1081, opened 2019-05-22
Experiment with "Checks" in Travis output

andrewn:
~**Do not merge**~

I'm creating this PR to experiment with the types of information that appears in the [Checks](./1081/checks) tab.

Ideally we'd want:
- lint errors
- lint warnings?
- test failures

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,11 +33,10 @@ jobs:
         - gcloud --quiet components update
         - gcloud --quiet components update kubectl
       deploy:
+        if: branch = master
         - provider: script
           script: ./deploy.sh
           skip_cleanup: true
-          on:
-            branch: master
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,23 +16,28 @@ before_install:
 
 install: true
 
-script:
-  - docker-compose exec -T app npm run test:ci --verbose
-
-before_deploy:
-  - docker-compose stop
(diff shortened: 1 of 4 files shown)


andrewn:
I don't think Travis has any way to collect metadata from build steps to display in GitHub. So, although we'll know that checks have failed, we'll still need to go through the build log output on travis-ci.com to see what exactly failed. Very sad.

andrewn:
I've split the build into two [*stages*](https://docs.travis-ci.com/user/build-stages/): `test` and `deploy`.

`test` has two jobs *linting* and *tests* that will run in parallel. If either fail then `deploy` will not run (and will only run on `master`).

The [Checks](checks) tab now displays info about which job has failed. This isn't as good as the fine-grained items you get with Circle CI:

Screenshot 2019-05-23 11 01 34

catarak:
very cool! let me know when this is ready to merge!

andrewn:
I've reverted the linting error and fixed the failing test so the check should now pass.

It's a shame Travis doesn't send a fine-grained breakdown of what's failed to GitHub but I think it's good enough for now.

As an aside, 93c226c shows why I don't think snapshot tests are very useful. Something in the `Nav` component has changed, but the JSON diff doesn't really give me any information about whether that change was OK or not. Unit testing the same component would capture the intent around it's behaviour and let us see if a change is intentional or not.

andrewn:
It seems as though travis is running the `deploy` stage even on this branch... I'll try and figure out what's going on.

andrewn:
Ok, that seems to have worked. Travis won't run the top-level `deploy` command on PR builds.

@catarak I think this is ready to review/merge.

catarak:
yay this looks great! merging ✨

Issue #908, opened 2019-03-16
Nav item CSS:hover is applied before the nav item is even clickable


shanrauf:
#### Nature of issue?

- Found a bug

#### Details about the bug:
- Web browser and version:  On Chrome
- Operating System: Windows 10
- Steps to reproduce this bug: Hover over a nav item where your cursor is causing the :hover effect to be applied, but since your cursor is not hovering over the text, the nav item dropdown is still not clickable

#### New feature details:
It causes confusion when the CSS hover effect is applied and the user tries clicking a nav item, but they can't click it unless the cursor is above the text.

#### Solution:
Change line 52 of client\styles\components_nav.scss

from
`.nav__item:hover {`

to
`.nav__item button:hover {`

This way, the CSS hover effect (text changing color) will only be applied when the user's cursor is above the button/text and is able to click the dropdown.

AnkurIngale:
Are you working on it or can I solve this one?

AnkurIngale:
Also, I'll like to do it. :-P

shanrauf:
You're right, making the whole area clickable is a better solution. I've been trying to work on this for a while and am having one small problem.

I created a new class for the nav item button so that this CSS does not affect the` .nav__dropdown` CSS).
```
.nav__item-heading {
  width: 100%;
  height: 100%;
}
```
Nav.jsx:
```
<li className={navDropdownState.file}>
            <button
              onClick={this.toggleDropdown.bind(this, 'file')}
              onBlur={this.handleBlur}
              onFocus={this.clearHideTimeout}
              className="nav__item-heading"
            >
              <span className="nav__item-header">File</span>
              <InlineSVG className="nav__item-header-triangle" src={triangleUrl} />
            </button>
```

The problem with this is that the padding of the li elements is not clickable. So there are these small spaces (padding) between every nav item that are not clickable (I attached a photo - the highlighted portion is clickable, but the padding is not). Do you have any suggestions?

![53ed378e40ffcfcf9d51c2af05d25f5d](https://user-images.githubusercontent.com/40217167/54481034-ec127f80-47ec-11e9-91df-fe3828529569.png)

siddhant1:
Since we are changing NAV styles, I will try to tackle it with #910 , no need for an extra pr

siddhant1:
I have completed the implementation , thanks for bringing it though :)

PR #1085, opened 2019-05-25
Fix hover effect on Log in and Sign up nav items

shanrauf:

* [x] is descriptively named and links to an issue number: `Applies fixes from #908 to "Log in" and "Sign up"`

### DIFF
--- a/client/components/Nav.jsx
+++ b/client/components/Nav.jsx
@@ -531,11 +531,19 @@ class Nav extends React.PureComponent {
         { __process.env.LOGIN_ENABLED && !this.props.user.authenticated &&
           <ul className="nav__items-right" title="user-menu">
             <li className="nav__item">
-              <p>
-                <Link to="/login">Log in</Link>
-                <span className="nav__item-spacer">or</span>
-                <Link to="/signup">Sign up</Link>
-              </p>
+              <Link to="/login">
+                <button>
+                  <span className="nav__item-header">Log in</span>
+                </button>
+              </Link>
+            </li>
+            <span className="nav__item-spacer">or</span>
+            <li className="nav__item">
+              <Link to="/signup">
+                <button>
+                  <span className="nav__item-header">Sign up</span>
+                </button>
+              </Link>
             </li>
           </ul>}
         { __process.env.LOGIN_ENABLED && this.props.user.authenticated &&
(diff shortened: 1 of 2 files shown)


catarak:
thanks for working on this. the hover color is fixed, however, it appears the spacing is a little off:
Screen Shot 2019-05-28 at 6 06 01 PM
there's more space between "log in" and "or" than "sign up" and "or"

shanrauf:
My apologies for putting this off for a whole week... I fixed the unequal spacing between Log in and Sign up. Please let me know if there are any other issues with this commit.

shanrauf:
Apologies for the weird commit history, had trouble with Git... The unequal spacing was caused by a margin that was originally placed on all nav items seemingly to make the dropdowns look nicer. However, since Log in and Sign up don't have dropdowns, I just deleted the margin placed on all nav items, and added a margin to the triangles (which will only affect the necessary nav items). 

Please let me know if there are any other issues.

client/components/Nav.jsx
```diff
@@ -531,11 +531,19 @@ class Nav extends React.PureComponent {
         { __process.env.LOGIN_ENABLED && !this.props.user.authenticated &&
           <ul className="nav__items-right" title="user-menu">
             <li className="nav__item">
-              <p>
-                <Link to="/login">Log in</Link>
-                <span className="nav__item-spacer">or</span>
-                <Link to="/signup">Sign up</Link>
-              </p>
+              <Link to="/login">
```
catarak: hey, so this html is not semantically correct—rather than inserting a `<button>` within the link, the styles for these should be updated.

shanrauf: Got it, I'll keep this in mind and make changes.

shanrauf:
My new commit doesn't weirdly add a button in the <Link> element anymore. The hover effect I added to all nav__item-headers is a bit redundant because right above it there is something similar for the left nav, but I don't see any other option. I als oremoved the nav__item class on the right nav <li> elements, because f you keep that class, then there is the same problem from #908 where the hover effect is applied but the link is not even clickable. Please let me know if you want anything changed!

catarak:
there's something funky happening with the git history on this branch...

shanrauf:
Apologies once again... Everything should be fixed, so please let me know if anything else is wrong. I appreciate you dealing with this inefficiency, I'm learning a lot...

catarak:
@shanrauf no worries!

catarak:
this is working great for me. thanks for working on this!

Issue #1095, opened 2019-06-06
Create project via Public API


andrewn:
#### Nature of issue?

- New feature request

#### New feature details:

As part of #542, this is to implement an endpoint that implements the Create Project endpoint in the [Public API  #1076](https://github.com/processing/p5.js-web-editor/pull/1076).

This will allow users to create new projects using their personal access token.

PR #1096, opened 2019-06-06
Test create project server code (prep for #1095)

andrewn:
Implements tests for the existing `Project` model and `createProject` controller.

I tried a few different approaches, each with their own pros/cons:

### Write manual mocks

This involves implementing the API of mongoose which is time consuming and error-prone.

### Access a test instance of mongodb directly

This tests the whole stack at the integration-test level. It makes sure that everything is working as it would in a live scenario. But, there's a lot of boilerplate to ensure that the database is in the correct state e.g. a user must exist to be able to create a project. This will also slow the tests down.

### Use [`mockingoose`](https://github.com/alonronin/mockingoose)

This wraps the mongoose models and runs the pre-save hooks, virtuals etc which is great.
It doesn't allow us to check that the mongoose methods were called with certain arguments, which is bad.

`mockingoose` is good for testing that the mongoose models e.g. `Project` are doing the correct thing, without accessing the database.

### Use [sinon-mongoose](https://github.com/underscopeio/sinon-mongoose)

This allows us to mock the exact responses for mongoose methods and check that they were called with the correct arguments. It doesn't run the hooks or virtuals so we can only check that the data structures we define are being passed around correctly. This is good for the controllers.

## The result

So:
- use mockingoose for testing the models
- use sinon-mongoose for testing the controllers

It's annoying to use two libraries, but I think it's workable.

## Future

A friend suggested that the controllers would be easier to test if we extracted the project creation functionality into our own model objects.

So, we'd have:

```js
export function createProject(req, res) {
  let projectValues = {
    user: req.user._id
  };

  projectValues = Object.assign(projectValues, req.body);

  OurProjectModel.create(projectValues, (err, newProjectWithUser) => {
    if (err) {
      res.json({ success: false });
      return;
    }

    res.json(newProjectWithUser);
  });
}
```

We'd write unit tests for `OurProjectModel.create` to make sure it called `Project.create` and `Project.populate`. In the controller tests, we'd mock `OurProjectModel`to test that the correct data was being pulled out of the HTTP request and sent in the response.

Something to think about for the future.

## Checks


andrewn:
@catarak Would be great if you could look over this approach and see what you think. I'm going to continue adding `createProject` controller tests to this PR.

server/controllers/__test__/project.controller.test.js
```diff
@@ -0,0 +1,155 @@
...
+
+        done();
+      }
+
+      promise.then(expectations, expectations).catch(expectations);
+    });
+
+    it('extracts parameters from request body', (done) => {
+      const request = {
+        user: { _id: 'abc123' },
```
catarak: should `user` here be within `body`?

catarak:
this is looking great to me! also, as a sanity check, all of the tests are passing for me.

server/controllers/__test__/project.controller.test.js
andrewn: The [current controller](https://github.com/processing/p5.js-web-editor/blob/df1219b/server/controllers/project.controller.js#L17) extracts the user id from `req.user._id`. I think Passport attaches the user from the logged in session to `req.user`.

andrewn:
@catarak I think I've addressed your comment above so I'm going to "squash and merge" this so I can use the tests in my public API branch.

PR #1106, opened 2019-06-17
Public API: Create new project (fixes #1095)

### LINKED ISSUE #1095: Create project via Public API

andrewn:
This implements the following new endpoints that match the Public API spec (#1076). I've modified `npm run fetch-examples-ml5` so that it uses the new endpoints. It expects a user to exists and a Public Access Token to be provided to the script.

- [x] `GET /api/:user/sketches` to list all sketches
- [x] `DELETE /api/:user/sketches/:id` to delete a sketch
- [x] `POST /api/:user/sketches` to create a new sketch
- [x] Tests for the mapping of the API files data structure to the mongo models
- [x] Descriptive error messages returned by the API
- [x] Tests for the three controller functions

I need to add more tests around the controllers, but I wanted to share now to get feedback.


server/controllers/project.controller.js
```diff
@@ -175,22 +203,43 @@ export function getProjects(req, res) {
...
+          res.status(500).json({ message: 'Error fetching projects' });
+        }
+      });
   } else {
     // could just move this to client side
     res.json([]);
   }
 }
 
+export function apiGetProjectsForUser(req, res) {
```
andrewn: I don't really like duplicating these functions as they're so similar, but they return different objects.

server/domain-objects/Project.js
```diff
@@ -0,0 +1,100 @@
+import pick from 'lodash/pick';
+import Project from '../models/project';
+import createId from '../utils/createId';
+
+// objectID().toHexString();
```
andrewn: Need to remove. 🤦‍♂

server/server.js
```diff
@@ -95,6 +96,7 @@ app.use(session({
 
 app.use(passport.initialize());
 app.use(passport.session());
+app.use('/api', requestsOfTypeJSON(), api);
```
andrewn: I think these public endpoints should be mounted at `/api` and all the rest that only the editor uses should be mounted at `/editor`?

catarak: to me that sounds logical. do you think we should add the API versioning as well? i was trying to figure out what best practices are for this, but it seems like the options all have their own drawbacks (here's a [blog post](https://blog.restcase.com/restful-api-versioning-insights/) that breaks this down).

server/controllers/project.controller.js
catarak: agreed, though it feels like the way to merge them is awkward/possibly confusing.

server/controllers/project.controller/createProject.js
```diff
@@ -30,3 +31,27 @@ export default function createProject(req, res) {
     .then(populateUserData)
     .catch(sendFailure);
 }
+
+// TODO: What happens if you don't supply any files?
```
catarak: i think there are two options here: (1) create default files (the ones that are currently the default when you click sketch-> new) or (2) the project isn't valid. my instinct here is (1)...

server/utils/createApplicationErrorClass.js
```diff
@@ -0,0 +1,35 @@
+/**
```
catarak: thanks for writing this class and and moving towards better error handling/capturing ✨

server/controllers/project.controller.js
andrewn: Yeah, I think that's true. Where possible, I've been extracting the shared data fetching logic into a helper function that can be called from both. And then the API controller functions can be used to format the data into a specific JSON structure.

server/controllers/project.controller/createProject.js
andrewn: This is done. See ee6891d.

andrewn:
@catarak Just pushed some tests and a check for slug uniqueness for the user. If a slug is specified and it is not unique in the sketches that the user has created, then the ids of the project(s) with identical slugs are returned as an error.

Eventually, I think this should live as a pre-save hook on the Project mongoose model. But as we don't have a UI for editing slugs, then we'd need to discuss what that behaviour should be first.

server/server.js
andrewn: I've added URL-based versioning in 57ced9c. Although it's not ideal, I think it's more obvious for casual users that may write scripts to import data into the Editor.

I'll change the namespaces of the private and public APIs in another PR.

andrewn:
This is now ready for review @catarak 🎉

catarak:
I'm not able to successfully create a sketch that generates default values. What I'm doing is `POST localhost:8000/api/v1/:username/sketches` with the body
```json
{
  "name": "My cute sketch"
}
```
to which i get the response
```json
{
  "message":"Sketch Validation Failed",
  "detail":"Slug \"undefined\" is not unique. ..."
}
```
if i add a slug to the request body, i.e.
```json
{
  "name": "My cute sketch",
  "slug": "my_cute_sketch"
}
```
the sketch will be successfully created, but it won't have any files.

catarak:
i also noticed that if you send `Content-Type: application/json` with `GET /api/v1/:username/sketches`, you get an error.

catarak:
`DELETE /api/v1/:username/sketches/:id` is working great though! i'm sending with the Content-Type header and i'm getting a 200 response back.

andrewn:
> I'm not able to successfully create a sketch that generates default values. What I'm doing is `POST localhost:8000/api/v1/:username/sketches` with the body
> 
> ```json
> {
>   "name": "My cute sketch"
> }
> ```
> 
> to which i get the response
> 
> ```json
> {
>   "message":"Sketch Validation Failed",
>   "detail":"Slug \"undefined\" is not unique. ..."
> }
> ```
> 
> if i add a slug to the request body, i.e.
> 
> ```json
> {
>   "name": "My cute sketch",
>   "slug": "my_cute_sketch"
> }
> ```
> 
> the sketch will be successfully created, but it won't have any files.

### Creating default files

This is fixed in de85fea. The issue was that if `files` wasn't provided, then nothing file related would happen. I've made it mandatory to provide the `files` property, even if it's empty. I think this is needed because if the API user makes a typo (e.g. `filez`, it will _seem_ as if their files are being created, but they'll get the default files instead which could be confusing.

The following should create default files:

```json
{
    "name": "My cute sketch",
    "files": {}
}
```

### Slug validation

I only can't reproduce the slug validation issue, which  is weird.

```json
{
    "name": "My cute sketch",
    "files": {}
}
```

Not providing the slug as above always creates a sketch for me, but it's exposed an interesting issue with auto-generating slugs.

The slug is generated on `model.save()` if it's not explicitly defined, which bypasses the uniqueness test. This means that `POST`ing 3 times will create 3 different documents with the same slug. Some solutions:

1. Allow projects to have the same slug when it's auto-generated
2. Don't auto-generate slugs at all
3. Append a counter or unique string to the slug when it's not unique e.g. `My_cute_sketch_2`

I wonder if not generating a slug unless the API user explicitly provides one is the most obvious behaviour?

andrewn:
> i also noticed that if you send `Content-Type: application/json` with `GET /api/v1/:username/sketches`, the response is an error, but if you don't, you get the right response back.

I raised PR #1122 to tackle this as it's a general issue with the `requestsOfType` middleware.

catarak:
> I only can't reproduce the slug validation issue, which is weird.

which version of mongo are you using? i'm using 3.6.6, since that's what mlab uses.

catarak:
> Not providing the slug as above always creates a sketch for me, but it's exposed an interesting issue with auto-generating slugs.
> 
> The slug is generated on `model.save()` if it's not explicitly defined, which bypasses the uniqueness test. This means that `POST`ing 3 times will create 3 different documents with the same slug. Some solutions:
> 
> 1. Allow projects to have the same slug when it's auto-generated
> 2. Don't auto-generate slugs at all
> 3. Append a counter or unique string to the slug when it's not unique e.g. `My_cute_sketch_2`
> 
> I wonder if not generating a slug unless the API user explicitly provides one is the most obvious behaviour?

Looking at the bigger picture here, the point of adding a slug is to create a human-readable permalink, right? So I think it would be cool if they were still auto-generated from the title, and I think it would be okay if a unique string/number were appended if there were a UI in which a user could change this slug if they wanted to. Therefore I'm leaning towards (3).

andrewn:
> > I only can't reproduce the slug validation issue, which is weird.
> 
> which version of mongo are you using? i'm using 3.6.6, since that's what mlab uses.

@catarak I still can't reproduce this. I dropped the `p5js-web-editor` mongo database and switched to using 3.6:

```
$ ps aux | grep mongo
andrew           11711   0.3  0.1  5106368  13004   ??  S    Mon09am  12:23.74 /usr/local/opt/mongodb-community@3.6/bin/mongod --config /usr/local/etc/mongod.conf
```

I created a new user (andrewn), with a new personal access token and POSTed to create a new sketch:

```
* Preparing request to http://localhost:8000/api/v1/andrewn/sketches
* Using libcurl/7.54.0 LibreSSL/2.6.5 zlib/1.2.11 nghttp2/1.24.1
* Current time is 2019-08-03T13:29:15.247Z
* Disable timeout
* Enable automatic URL encoding
* Disable SSL validation
* Enable cookie sending with jar of 0 cookies
* Found bundle for host localhost: 0x7fb7e1f28530 [can pipeline]
* Re-using existing connection! (#4) with host localhost
* Connected to localhost (::1) port 8000 (#4)
* Server auth using Basic with user 'andrewn'

> POST /api/v1/andrewn/sketches HTTP/1.1
> Host: localhost:8000
> Authorization: Basic YW5kcmV3bjpORFZrWVRCbU0ySXhOV1ExWTJVM05tTmxaamhrT1RWbVkyTXhPVEV6WkRJNU1EZzNPRGhqT0E9PQ==
> User-Agent: insomnia/6.5.4
> Content-Type: application/json
> Accept: */*
> Content-Length: 44

| {
|   "name": "My cute sketch",
| 	"files": {}
| }

* upload completely sent off: 44 out of 44 bytes

< HTTP/1.1 201 Created
< X-Powered-By: Express
< Vary: Origin
< Access-Control-Allow-Credentials: true
< Content-Type: application/json; charset=utf-8
< Content-Length: 18
< ETag: W/"12-7Z0Dd/UqbdLyYIPp5BjeLo/dpLU"
< Date: Sat, 03 Aug 2019 13:29:15 GMT
< Connection: keep-alive


* Received 18 B chunk
* Connection #4 to host localhost left intact
```

`mongoose` is at v 4.13.18:

```
$ npm ls mongoose
p5.js-web-editor@0.0.1 /Users/andrew/Projects/oss/processing/p5.js-web-editor
└── mongoose@4.13.18
```

Perhaps my `mongod.conf` is different?

catarak:
let me try testing this again to sanity check, it's super weird!

catarak:
i'm using postman to test this, maybe that has to do with it?

catarak:
In testing this again, I found a bug. I made a POST to `/api/v1/garbage_username/sketches` with the correct authorization header (my real username, `cassie` + the API key), and a sketch body (`{
	"name": "My cute sketch",
	"files": {}
}`) and it successfully created a sketch under my username.

andrewn:
> In testing this again, I found a bug. I made a POST to `/api/v1/garbage_username/sketches` with the correct authorization header (my real username, `cassie` + the API key), and a sketch body (`{ "name": "My cute sketch", "files": {} }`) and it successfully created a sketch under my username.

This is not fixed. An `UNAUTHORIZED` response code is returned with the username of the authenticated user, and the username specified in the URL.

catarak:
sweet, thanks for fixing that! i don't think there's any other issues so i'll merge this in.

PR #1111, opened 2019-06-30
for #989, files cannot be renamed to empty string

raclim:

### DIFF
--- a/client/modules/IDE/components/FileNode.jsx
+++ b/client/modules/IDE/components/FileNode.jsx
@@ -68,20 +68,12 @@ export class FileNode extends React.Component {
     const oldFileExtension = this.originalFileName.match(/\.[0-9a-z]+$/i);
     const newFileExtension = this.props.name.match(/\.[0-9a-z]+$/i);
     const newFileName = this.props.name;
-    if (oldFileExtension && !newFileExtension) {
-      this.props.updateFileName(this.props.id, this.originalFileName);
-    }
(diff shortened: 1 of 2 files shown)


client/modules/IDE/components/FileNode.jsx
```diff
@@ -77,6 +78,12 @@ export class FileNode extends React.Component {
     ) {
       this.props.updateFileName(this.props.id, this.originalFileName);
     }
+    if (newFileName === '') {
```
andrewn: Since the each clause of these four `if` statements have the same effect, you could combine them into a single statement. 

Maybe assign the conditionals to variables, like:

```js
const hasNoExtension = oldFileExtension && !newFileExtension;
const hasSameExtension = oldFileExtension && newFileExtension && oldFileExtension[0].toLowerCase() !== newFileExtension[0].toLowerCase();
const hasEmptyFilename = newFileName === '';
const hasOnlyExtension = newFileName === newFileExtension[0];

if (hasNoExtension || hasSameExtension || hasEmptyFilename || hasOnlyExtension) {
}
```

catarak: i like this suggestion! i think it is easier to read and understand.

catarak:
this looks great to me! i added a related, but different fix, just because i was already looking at this code, which prevents users from adding a file extension to a folder. i figured this would save users from themselves 😸

PR #1122, opened 2019-07-21
Fixes bug where requestsOfType() would fail if no body

andrewn:
The `requestsOfType()` middleware wasn't behaving correctly when an incoming request had no body. When there is no body, the request should succeed.

Express' [`request.is(type)`](https://expressjs.com/en/api.html#req.is) function can return three things:
- the content-type of the request if it matches the specified `type` e.g. `application/json`
- `null` if the request has no body
- `false` if it doesn't match

If the request does have a body, then we check that the `Content-Type` matches the `type` provided, and fail if it doesn't.

I also added some error middleware that will return a JSON object with the correct status code from the API and will log a more descriptive error in development.

--

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -132,6 +132,17 @@ app.get('/', (req, res) => {
   res.sendFile(renderIndex());
 });
 
+// Handle API errors
+app.use('/api', (error, req, res, next) => {
+  if (error && error.code && !res.headersSent) {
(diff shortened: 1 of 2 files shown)


catarak:
this is great. i had started working on the api route for uploading files, and noticed some weirdness with this middleware (since file uploads are of type `multipart/form-data`).

Issue #1088, opened 2019-05-29
Viewing a Map in the console crashes the session


Dollar-500:
#### Nature of issue?

- Found a bug

#### Details about the bug: 

- Web browser and version: Google Chrome Version 74.0.3729.169
- Operating System: Windows
- Steps to reproduce this bug: 
-- Copy and paste the linked code into the web editor
-- Open up the console
-- Click the disclosure triangle on the Map object that was printed

Here is the code (it's short, don't worry): https://pastebin.com/WWCC0yqu

Here is the window + dev tools window *before* I clicked the disclosure triangle:
<img width="960" alt="BeforeClick" src="https://user-images.githubusercontent.com/39846627/58595211-805d8180-8235-11e9-8621-8b90750e1740.PNG">

Here is the window + dev tools window *after* I clicked the disclosure triangle:
<img width="959" alt="AfterClick" src="https://user-images.githubusercontent.com/39846627/58595212-84899f00-8235-11e9-968b-ade5e5566d9d.PNG">

catarak:
thanks for reporting!

catarak:
i tried running this locally to get better errors, and this is what i'm seeing:
```
checkPropTypes.js?21ee:20 Warning: Failed prop type: Invalid prop `name` of type `object` supplied to `ConnectedTreeNode`, expected `string`.
    in ConnectedTreeNode (created by ConnectedTreeNode)
    in ConnectedTreeNode (created by TreeView)
    in TreeView (created by ObjectInspector)
    in ThemeProvider (created by ObjectInspector)
    in ObjectInspector (created by Inspector)
    in Inspector (created by CustomInspector)
    in div (created by Styled(div))
    in Styled(div) (created by CustomInspector)
    in CustomInspector (created by WithTheme(CustomInspector))
    in WithTheme(CustomInspector) (created by ObjectTree)
    in ObjectTree (created by WithTheme(ObjectTree))
    in WithTheme(ObjectTree) (created by ConsoleMessage)
    in div (created by Styled(div))
    in Styled(div) (created by ConsoleMessage)
    in div (created by Styled(div))
    in Styled(div) (created by ConsoleMessage)
    in ThemeProvider (created by ConsoleMessage)
    in ConsoleMessage (created by Console)
    in div (created by Styled(div))
    in Styled(div) (created by Console)
    in ThemeProvider (created by Console)
    in Console (created by Console)
    in div (created by Console)
    in div (created by Console)
    in div (created by Console)
[...]

PR #1126, opened 2019-07-31
Wraps console external component in error boundary

bomoko:
addresses #1088 

Demonstration of the kind of ErrorBoundary that can be set up to stop the console-feed component crashing the entire IDE


catarak:
do you think that if there's an error and the console component crashes, it would be possible to mount a new one?

bomoko:
I'll give that a shot! I guess the only thing we really want to avoid is the console going into some weird mounting/unmounting loop (assuming we're passing through the same data to the component, maybe it'll be prudent to clear out the console items on unmount ... don't really have any solid ideas here, just speaking out loud and getting around to saying --->) I'll see what I can do.

Issue #588, opened 2018-03-02
Console errors highlight wrong file


catarak:
#### Nature of issue?

- [X] Found a bug

#### Details about the bug: 

If you run a sketch and it throws an error, the line of any file that is selected will be highlighted, rather than switching to the file that has the error. The expected behavior is that it would switch to the file that threw the error.

- Web browser and version: Chrome (also this is really N/A)
- Operating System: Mac OSX (also N/A)
- Steps to reproduce this:

1. open a new sketch
2. make a syntax error in `sketch.js`
3. switch to `index.html`
4. click play and run the sketch
5. `index.html` will still be selected and the line number that threw the error in `sketch.js` will be highlighted.

shinytang6:
would like to take this up :smile:

siddhant1:
Anything left here? @catarak

catarak:
@siddhant1 this bug has not been fixed and is still happening!

siddhant1:
Ok , I will send a fix :)

jpgualdarrama:
I'm new here, but I would like to take this on if it hasn't been fixed yet. @catarak

PR #1136, opened 2019-08-17
Fix Issue #588. Only highlight runtime errors at the appropriate line…

jpgualdarrama:
… for file in which the error occurred.

### DIFF
--- a/client/modules/IDE/components/Editor.jsx
+++ b/client/modules/IDE/components/Editor.jsx
@@ -197,7 +197,7 @@ class Editor extends React.Component {
             consoleEvent.data[0].indexOf(')') > -1) {
             const n = consoleEvent.data[0].replace(')', '').split(' ');
             const lineNumber = parseInt(n[n.length - 1], 10) - 1;
-            if (!Number.isNaN(lineNumber)) {
+            if (!Number.isNaN(lineNumber) && `${consoleEvent.source}.js` === this.props.file.name) {
               this._cm.addLineClass(lineNumber, 'background', 'line-runtime-error');
             }
           }


client/modules/IDE/components/Editor.jsx
```diff
@@ -197,7 +197,7 @@ class Editor extends React.Component {
             consoleEvent.data[0].indexOf(')') > -1) {
             const n = consoleEvent.data[0].replace(')', '').split(' ');
             const lineNumber = parseInt(n[n.length - 1], 10) - 1;
-            if (!Number.isNaN(lineNumber)) {
+            if (!Number.isNaN(lineNumber) && `${consoleEvent.source}.js` === this.props.file.name) {
```
catarak: what if the error is in javascript that is in a script tag in the html file?

jpgualdarrama: I'll investigate what I can do in that case, but from what I found so far, there may be other fixes required to correctly report errors from javascript in a script tag anyway. When a runtime error is found in a <script> tag in the html, the filename is not reported and the line number is not reported correctly. I made a sketch to demonstrate: https://editor.p5js.org/jon_gualdarrama/sketches/NXb-s08x1

jpgualdarrama: Commit 39efd19 includes part of what I believe has to be done in order to highlight error lines in a <script> tag. It isn't complete yet, as the line number reported in the error will not match the line number in the html file on which the error actually occurred, so I set the line class on the correct line of the html file. After some googling, I've convinced myself that it has something to do with source maps - https://trackjs.com/blog/debugging-with-sourcemaps/. Will continue to investigate.

catarak: i don't think it has to do with source maps—the sketch gets compiled into one long html file, which messes with the line numbers, and the necessary offsets are calculated by a function called `getAllScriptOffsets`, which is in `client/utils/consoleUtils.js`. looking at this function now, i see that it's only capturing the offsets for js files, and not inline js.

catarak:
another potential solution to this PR is to change the editor focus to the file that has the error—what do you think about this flow, versus just showing the error in the console if the file isn't selected?

jpgualdarrama:
I'm not sure how that would resolve this PR; would that help resolve the problem with the html files?

Also, I don't think just changing the focus to the be the file in which the error occurred would be sufficient to resolve the issue, as the problem is that once the error occurs on line _X_ of some file, no matter what file is currently active in the editor, line _X_ will always be highlighted, assuming the file has at least _X_ lines in it. If the user switches files while the error still exists, the same issue will continue to occur.

jpgualdarrama:
I think I've fixed the issue in `getAllScriptOffsets` to include index.html and updated `Editor.jsx` to highlight the correct line in index.html @catarak

Issue #1169, opened 2019-10-01
Action should a plain object


shakti97:
#### Nature of issue?

- Found a bug

#### Details about the bug: 

- Web browser and version: Chrome/FireFox/Opera 
- Operating System: Windows/MacOSX/Linux 
- Steps to reproduce this bug:

Steps to reproduce:-
1. first login and save the opening default file from the File menu
2. Now download the file
3. Move to the console it will raise the error

![Error](https://user-images.githubusercontent.com/30585570/65987582-bd67a800-e4a3-11e9-92ae-9d9e0a756659.PNG)

shakti97:
I working on it

shakti97:
Issue #1137 for downloading issue - file not downloading completely this issue 
This issue for removing the above shown error

PR #1170, opened 2019-10-01
Fixed action error while downloading

shakti97:

* It fixes the issue #1169


catarak:
thanks for working on this issue. i don't think that this is a stable solution, because it doesn't address the issue directly. instead, this dispatches a fake action to quiet an error, whereas the function itself should not be dispatching an action.

Issue #1196, opened 2019-10-17
loadImage() broken when using Editor with p5 0.10.2


stalgiag:
#### Nature of issue?

- Found a bug

#### Details about the bug: 

- Web browser and version:  Chrome 77.0.3865
- Operating System: Windows 
- Steps to reproduce this bug: [Go to this example](https://editor.p5js.org/stalgiag/sketches/AEa__RUdZ)

I am using the web editor to host [some examples for p5.xr](https://stalgiag.github.io/p5.xr/#/), so I have been **manually loading in v0.10.2 of p5 into the editor** since it isn't default linked yet. This is lucky because in doing so, I found that `loadImage()` silently fails (or silently to the editor console, not to the browser console).

The error is :
```
TypeError: Cannot read property 'includes' of null
```
for [this line](https://github.com/processing/p5.js/blob/master/src/image/loading_displaying.js#L77). So `fetchResponse.headers.get('content-type')` is returning `null`. Note that this works when the sketch is run locally.

_Funny enough,_ I actually wrote the line that is giving the bug. I can fix it on the p5 side if that makes sense, but I am wondering if there is any idea over here as to why this would throw an error on the editor in specific?

catarak:
when you're calling `loadImage()`, is the path to a file uploaded to the web editor, or is it hosted elsewhere? my instinct here is that it's a CORS issue that's throwing an error.

stalgiag:
The path is to a file uploaded to the web editor. In [this example](https://editor.p5js.org/stalgiag/sketches/AEa__RUdZ) at least.

catarak:
with some digging, i think it's related to #496... since `response.headers.get('content-type')` is null

stalgiag:
Oops! I should've caught that. Thanks for making the connection!

I am happy closing this as a duplicate but maybe I should make a note in the other issue pointing out that the issue should be resolved before updating the editor link to the latest version of p5 since the GIF feature relies on a content-type check?

catarak:
makes sense to me to get it fixed before updating the p5 version!

PR #1200, opened 2019-10-21
Add content-type to assets uploaded to s3

### LINKED ISSUE #496: Blank Content-Type set for uploaded assets in S3
#### Nature of issue?

- [x] Found a bug
- [ ] New feature request

#### Details about the bug: 

Firefox 57.0.1/macOS 10.13.1
- open the editor, sign in to an account
- open the sidebar for a sketch, open the new file modal
- drag-and-drop a file (i.e. a .jpg image) and let it finish uploading
- in the AWS S3 console, or by using a ```curl -I``` on the asset url, observe that the Content-Type is blank
- if you have the network inspector open while uploading, you'll see that the Content-Type field in the POST to S3 is blank

stalgiag:

resolves #496 
addresses #1196 

This adds the relevant `Content-Type` metadata to files uploaded to AWS. This fixes a bug introduced with p5 v0.10.2 which checks the header's content-type to determine whether an asset is a GIF.

I have tested this locally with a personal S3 bucket.

Potential issues - I think this only handles assets uploaded through the drag+drop interface. I assume that this is the only way that a user will upload a media asset but I could be wrong.

### DIFF
--- a/client/modules/IDE/actions/uploader.js
+++ b/client/modules/IDE/actions/uploader.js
@@ -82,7 +82,7 @@ export function dropzoneSendingCallback(file, xhr, formData) {
       Object.keys(file.postData).forEach((key) => {
         formData.append(key, file.postData[key]);
       });
-      formData.append('Content-type', '');
+      formData.append('Content-type', file.type);
       formData.append('Content-length', '');
       formData.append('acl', 'public-read');
     }


catarak:
yay thank you! 

eventually, it will be possible to upload assets via the api (#542), but that can be handled separately from this PR.

catarak:
one major issue is that all of the assets currently uploaded to S3 don't have a content type, so we'll need to write a script to update them. i'm looking at https://stackoverflow.com/questions/41015200/how-to-change-content-type-of-amazon-s3-objects

catarak:
anyway, this is working great for me and i'm going to merge it.

stalgiag:
Oh yeah! I didn't think about that. Perhaps this will be an edge-case? Updating the version on existing projects seems rare, but I may be missing something. If writing a script to update the existing assets proves difficult, then we could add something on the p5 side to catch this and at least output a helpful warning.

catarak:
i imagine that folks will rarely update the p5.js version in existing sketches, so yeah, i think it's more of an edge case. it would be cool if p5.js could throw out a useful warning if possible!

stalgiag:
I agree! Opened a PR on the p5 library side :-)

PR #1230, opened 2019-12-29
Update IDEView.jsx

gabrielsroka:
`<Overlay...` has a default `title` of `'Modal'`
https://github.com/processing/p5.js-web-editor/blob/master/client/modules/App/components/Overlay.jsx#L102

many modals are missing a title, so i added some. i made them basically the same as the `ariaLabel` (but capitalized, changed `-` to ` `, etc)


catarak:
thank you! in the future it would be helpful if you opened an issue before opening a pull requests—it's better for documentation bugs/features in the project.

PR #1233, opened 2020-01-06
Fix Issue #1199, Ensure login button is disabled on page load for mob…

### LINKED ISSUE #1199: In mobile display can't identify if button is disabled or not.
#### Nature of issue?

#### Details about the bug: 

- Operating System: LInux

Screenshot from 2019-10-21 08-37-55
In desktop when we hover on the button we can see an icon which shows it is disabled until all the details are not entered correctly.

Whereas in the case of mobile, It is really difficult for a user to figure out if the button is disabled or not. 
Screenshot from 2019-10-21 08-37-19

#### New feature details:
Change the CSS properties of button when the screen size is reduced to 500px or so.

tanhengyeow:
Fixes #1199 

### DIFF
--- a/client/modules/User/components/LoginForm.jsx
+++ b/client/modules/User/components/LoginForm.jsx
@@ -30,7 +30,13 @@ function LoginForm(props) {
         />
         {password.touched && password.error && <span className="form-error">{password.error}</span>}
       </p>
-      <input type="submit" disabled={submitting || pristine} value="Log In" aria-label="login" />
+      <input
+        type="submit"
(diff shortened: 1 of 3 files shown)


catarak:
thanks for submitting this PR! i don't think there's any reason to differentiate this button's design _only_ on mobile—why not add the same disabled style on desktop as well?

also, rather than using raw colors, this project uses a theming mixin, see https://github.com/processing/p5.js-web-editor/blob/master/client/styles/abstracts/_placeholders.scss as an example

PR #1257, opened 2020-02-13
[Experimental] Standalone Frontend Dev Server

hydrosquall:
## Motivation

This is an experiment to to make it easier for people to contribute to the frontend by reducing the amount of setup steps (no mongodb) if they're making a frontend-only change. If we like the dev experience, it might be worthwhile to extract the "client" portion of the app to its own package, which the "server" can import.

## Changes

- Changed the references to `process` based on the simplified definition of `index.html` (I didn't bind the process variables to the window).
- Parcel is a "zero config" dev server that we could used instead of running a full dev server. You can find [docs](https://github.com/parcel-bundler/parcel) here. Next.js is another option that could be worth investigating. 
- I am running `12.16.0` locally since I use `nodenv` rather than `nvm`, and `12.16.1` wasn't available. However, hopefully there shouldn't be significant differences.

## Testing

1. Clone this branch
2. `npm run parcel`
3. `cp .env.example .env`
4. The app should work on localhost port `1234`.

A live version of this app was auto-deployed to `now.sh` at this url:

- Branch specific URL: https://p5-js-web-editor-jhljph7vi.now.sh/
- Vanity URL works too: https://p5-js-web-editor.now.sh/

(Login and other stateful features are disabled, since I didn't supply any environment variables. However, if supplied locally, those should work too. The API_URL will probably not work out-of-the-box, since it looks like CORS is disabled with the production API.

The now Github integration is capable of automatically building branch previews every time someone opens a PR, but I manually supplied the link here since this is still early-stage.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -23,7 +23,6 @@
             "loose": true
           }
         ],
-        "@babel/plugin-proposal-json-strings",
         "@babel/plugin-proposal-function-sent",
         "@babel/plugin-proposal-export-namespace-from",
         "@babel/plugin-proposal-numeric-separator",
@@ -36,10 +35,7 @@
           {
             "proposal": "minimal"
           }
-        ],
-        "@babel/plugin-proposal-nullish-coalescing-operator",
-        "@babel/plugin-proposal-do-expressions",
-        "@babel/plugin-proposal-function-bind"
+        ]
       ],
       "presets": [
         "@babel/preset-env",
@@ -53,7 +49,6 @@
     }
   },
   "plugins": [
-    "@babel/plugin-syntax-dynamic-import",
     "@babel/plugin-syntax-import-meta",
     [
       "@babel/plugin-proposal-decorators",
@@ -67,22 +62,17 @@
         "loose": true
       }
     ],
-    "@babel/plugin-proposal-json-strings",
     "@babel/plugin-proposal-function-sent",
     "@babel/plugin-proposal-export-namespace-from",
     "@babel/plugin-proposal-numeric-separator",
     "@babel/plugin-proposal-throw-expressions",
     "@babel/plugin-proposal-export-default-from",
     "@babel/plugin-proposal-logical-assignment-operators",
-    "@babel/plugin-proposal-optional-chaining",
     [
       "@babel/plugin-proposal-pipeline-operator",
       {
         "proposal": "minimal"
       }
-    ],
-    "@babel/plugin-proposal-nullish-coalescing-operator",
-    "@babel/plugin-proposal-do-expressions",
-    "@babel/plugin-proposal-function-bind"
+    ]
   ]
 }
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ cert_chain.crt
 localhost.crt
 localhost.key
 privkey.pem
+.cache/
--- a/client/components/Nav.jsx
+++ b/client/components/Nav.jsx
@@ -16,7 +16,7 @@ import { metaKeyName, } from '../utils/metaKey';
 const triangleUrl = require('../images/down-filled-triangle.svg');
 const logoUrl = require('../images/p5js-logo-small.svg');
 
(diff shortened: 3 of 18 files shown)


catarak:
Thanks for working on this! Question about using Parcel—is it weird to have both Webpack and Parcel in a project? Does this circumvent Webpack?

hydrosquall:
If we like Parcel enough, it could replace webpack- I just didn’t want to
prematurely remove it. In either case, both bundlers were reading from the
same .babelrc, hence the changes to that file.

If there are custom things that we need our webpack  to do, then switching
over may not make sense. If everything is covered by parcel’s defaults, we
would hopefully have a smaller build related maintenance burden over time
(see
https://flaviocopes.com/parcel/ )

andrewn:
I agree that not having to run the API and mongodb to make front-end changes would be great!

I think we should use the same packager and config for all development environments. This will ensure making a change in the front-end-only env will be the same as the full-stack env and avoid weird bugs with the different packagers.

Perhaps a good way forward is separating the webpack dev server from the API server so that the can be run standalone? Then we could also switch to Parcel if we wanted.

hydrosquall:
Hi @andrewn, thanks for the idea! I've filed this as an issue to take on separately from this branch: https://github.com/processing/p5.js-web-editor/issues/1348

Do you have any advice about where to start on this task? I'll see if I can budget some time for it in upcoming weeks.

catarak:
Thanks @hydrosquall!

I would be down to explore if it makes sense to move from webpack to parcel. There's really nothing fancy happening with the bundling in this app, and maybe parcel is good enough.

hydrosquall:
I took a look another at trying with the latest Parcel 2 and the latest head of `develop`.

I ran into a few hurdles due to changes to both this codebase and Parcel itself, some of which seemed time effective to solve, others which do not. 

1. [x] The original index.html in this file did not work. The script type needs to be changed to `module`. See https://github.com/fregante/browser-extension-template/issues/51 . 
2. [x] Parcel 2 did not handle importing audio files (mp3) by default. This was fixable by following https://github.com/parcel-bundler/parcel/issues/1911 , I chose to create a `.parcelrc` instead of doing the `url:` prefix method to avoid breaking the existing code.
3. [ ] Parcel 2 does not handle `resourceQuery`, and there are no plans to do so. ( see https://github.com/parcel-bundler/parcel/issues/2186 , https://github.com/parcel-bundler/parcel/issues/1084 ). This is an issue because the P5 editor needs this to load SVG files in 3 different ways:
    a. As raw text files, to inline (`?byContent`), as part of the [replace](https://github.com/processing/p5.js-web-editor/pull/1552) feature
    b. As URL paths, to inject into stylesheets (`?byUrl`) (added as part of https://github.com/processing/p5.js-web-editor/pull/1420 )
    c. As React components (with the SVGR Loader) (Seemed to mostly change as part of `b` too).

https://github.com/processing/p5.js-web-editor/blob/f373938bc8554694cb8d66bafbbce93e63e9d90d/webpack/config.prod.js#L119-L140

I look at the code that actually reads these SVGs. It didn't seem worthwhile to rewrite this code such that all SVGs are loaded in 1 way, since there seemed to be legitimate reasons to be loading icons in 3 different ways in the editor for now.

I didn't have confidence that this would be the end of Parcel configuration issues (inability to load relative URLs looked problematic too https://github.com/parcel-bundler/parcel/issues/206 ), so I think may be best to retire the idea of using Parcel 2 to bundle this project, given that the current webpack config has a few customized asset loaders. 

It might be more practical to get the standard webpack dev server working instead, so I'm going to try that in a separate PR.

## References

- Proof-of-concept branch with Parcel 2 (build fails because it's not sure how to handle SVGs with resourceQuery strings, but it solves 1 and 2): https://github.com/hydrosquall/p5.js-web-editor/tree/cameron.yick/experiment/parcel-frontend-only

hydrosquall:
Closing this PR as it will be replaced by https://github.com/processing/p5.js-web-editor/pull/1999 .

PR #1260, opened 2020-02-13
Fix svg title

### LINKED ISSUE #1259: Every InlineSVG is missing an aria-label
- Found a bug
Right Now Every SVG used is missing an aria-label/tooltip when we hover over an SVG it says triangle, a wrapper over svgs, like button has an aria-label but hovering over the  SVG shows the aria-label on the SVG which is by default set to triangle.
This feature can be easily added by providing title as a prop on the InlineSVG component.

ashu8912:

* [x] is descriptively named and links to an issue number, i.e. 
Fixes #1259


ashu8912:
Though This Might Not Be the Best Feature But This certainly gives boost to accessibility earlier every svg was having a wrong title but now they have correct title and helps screen readers a lot

PR #1281, opened 2020-02-23
made settings tab mobile responsive

### LINKED ISSUE #1036: Responsiveness for settings overlay

devashish1099:

* [x] is descriptively named and links to an issue number, i.e. `Fixes #1036 `

no more overflow :
Screenshot_20200223_225104

no more broken width:
Screenshot_20200223_224934


client/styles/components/_overlay.scss
```diff
@@ -42,3 +43,15 @@
   @include icon();
   padding: #{3 / $base-font-size}rem 0;
 }
+
+@media only screen and (max-width: 430px){
```
catarak, line 60: Please don't just add random numbers for the breakpoints.  They should be variables used in Sass functions.

PR #1330, opened 2020-03-17
fix(file.controller.js): set headers for file extensions from MIME -I1218

### LINKED ISSUE #1218: Text assets's content-type is always text/html
#### Nature of issue?
- Found a bug

#### Details about the bug: 

- Operating System: Windows

I found getFileContent function at /p5.js-web-editor/server/controllers/file.controller.js
But there are nothing considered about content-type and rely on default behavior.

Express official document says
> When the parameter is a Buffer object, the method sets the Content-Type response header field to “application/octet-stream”

> When the parameter is a String, the method sets the Content-Type to “text/html”

https://expressjs.com/en/api.html#res.send

I think that it should be text/javascript(or application/javascript) for .js, application/json for .json and so on.

Thank you.

elit-altum:

Fixes #1218 

- Adds [mime-types](https://www.npmjs.com/package/mime-types) as a dependency to check for the file's MIME and then set an appropriate header.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13939,16 +13939,16 @@
       "integrity": "sha512-x0Vn8spI+wuJ1O6S7gnbaQg8Pxh4NNHb7KSINmEWKiPE4RKOplvijn+NkmYmmRgP68mc70j2EbeTFRsrswaQeg=="
     },
     "mime-db": {
-      "version": "1.40.0",
-      "resolved": "https://registry.npmjs.org/mime-db/-/mime-db-1.40.0.tgz",
-      "integrity": "sha512-jYdeOMPy9vnxEqFRRo6ZvTZ8d9oPb+k18PKoYNYUe2stVEBPPwsln/qWzdbmaIvnhZ9v2P+CuecK+fpUfsV2mA=="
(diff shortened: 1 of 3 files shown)


elit-altum:
@stalgiag Is this the right approach ? Please review!

stalgiag:
Hi @elit-altum one thing that I am noticing is that this is failing the linting test with the error `Absolute imports should come before relative imports`. Try moving up the import for mime-types.`

Other than that, this looks like it should work but I don't have the opportunity to test currently. Are you able to test successfully with the appropriate file types?

elit-altum:
Oh that's my bad @stalgiag. Thanks for pointing it out. I am really sorry
I tried logging the output of the return from ```mime.lookup``` and that seems to be working fine. But I am new to this so wanted a second opinion

andrewn:
Thanks for opening this PR!

`resolvedFile` is an object, and `mime.lookup()` requires a filename string so I've changed this to `resolvedFile.name`.

Also, `lookup()` returns `false` if no mime-type is found, so I've added a fallback to `application/octet-stream` as recommended by the library otherwise the server crashes.

PR #1372, opened 2020-04-07
Fixes #681 - File name update triggers editor change detection

### LINKED ISSUE #681: When changing a file name, sketch does not mark itself as unsaved

ghalestrilo:
### Problem (#681)

Editor would not detect filename changes as project changes. This is a minor issue since usually filenames will be referred to by other files, which require changing said files, therefore triggering a change in the project. Still, it's a nice detail to have.

### Solution

Detecting these changes require comparing the original filename to the new one, so a new `updatedName` field was created in the `files` reducer. The `updatedFileName` action should detect if they are different before triggering changes, requiring the ability to dispatch more than one action at once, so it was refactored to a thunk. Finally, the saveProject action must commit these name changes, and clear their `updatedNames`, so this was added right before sending the request.


catarak:
Thanks for working on this! While this approach works, I'm not sure it's the best one. I think it can be super confusing when to store the state in React vs. Redux, and this is a case in which I think it makes sense to store it in React.

I would suggest taking a similar approach, but storing `updatedName` in the state of `client/modules/IDE/components/FileNode.jsx`, and then only calling `this.props.updateFileName` if `state.updatedName !== props.name`. Then the action `updateFileName` can always ` dispatch(setUnsavedChanges(true));`.

ghalestrilo:
@catarak Agreed, thanks for the input. I followed your suggestion, and it's a bit clearer now

client/modules/IDE/components/FileNode.jsx
```diff
@@ -103,6 +118,7 @@ export class FileNode extends React.Component {
 
   hideEditFileName() {
     this.setState({ isEditingName: false });
+    this.commitFileNameChange();
```
catarak, line 121: Because this is called `onBlur`, along with `validateFileName`, adding `this.commitFileNameChange()` here is actually causing `this.props.updateFilename` to be called multiple times. i think this line and function is unnecessary!

client/modules/IDE/components/FileNode.jsx
```diff
@@ -47,15 +48,29 @@ export class FileNode extends React.Component {
     }, 200);
   }
 
+  getName() {
```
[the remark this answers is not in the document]
ghalestrilo: True, I made a getter for incapsulation. Accessing this.state all over the place can be problematic

client/modules/IDE/components/FileNode.jsx
ghalestrilo: This call is actually really important, It's what triggers change detection!
I noticed `validateFileName` only calls `this.props.updateFilename` when the filename is problematic, to reset it to the original value.
Keeping this line as it is would skip filename validation entirely, which is very problematic, so I moved it into `validateFileName`. Now it's only called when the filename is valid, and no duplicate calls occur :)

ghalestrilo:
I chose to write a quick test for that filename check, hope that's ok
This is an important component to test - I want to write more cases on a new branch

catarak:
I couldn't help myself and ended up hijacking this PR 🤦 Ended up with a great refactor though!

PR #1381, opened 2020-04-08
resolve path present assets #1268 

### LINKED ISSUE #1268: A Problem about the Present Mode
#### Nature of issue?

- Found a  bug (not sure if it's a bug)

#### Details about the bug: 

- Web browser and version: Safari 13.0.3 and Chrome 79.0.3945.130
- Operating System: macOS Catalina 15.10.1 
- Problem: I use a for loop like this to load images
```
for (let i=0;i<4;i++){
 let tem = "filename"+str(i+1)+".jpg";
imageArray.push(loadImage(tem));
}
```
and it work good as a sketch, as an embed and being run locally. But It doesn't work on the present page (aka on editor.p5js.org/account/present/projectcode) and it return a 404 error of wrong image path.
And if I use the normal method it work well on every mode.

davidbmx:

* [x] is from a uniquely-named feature branch and has been rebased on top of the latest master.
* [x] is descriptively named and links to an issue number, i.e. `Fixes #1268`

### DIFF
--- a/client/components/Nav.jsx
+++ b/client/components/Nav.jsx
@@ -227,7 +227,7 @@ class Nav extends React.PureComponent {
 
   renderDashboardMenu(navDropdownState) {
     return (
-      <ul className="nav__items-left" title="project-menu">
+      <ul className="nav__items-left">
         <li className="nav__item-logo">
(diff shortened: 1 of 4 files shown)


catarak:
Found another issue that was causing this not to work—works locally now though!

Issue #1348, opened 2020-03-27
[dev-task] Separate Webpack Dev server from API server


hydrosquall:
#### Nature of issue?

- Refactoring to enable devs to make frontend changes without running an API server or database, as suggested by @andrewn 

#### New feature details:

> Perhaps a good way forward is separating the webpack dev server from the API server so that the can be run standalone? Then we could also switch to Parcel if we wanted.

- https://github.com/processing/p5.js-web-editor/pull/1257#issuecomment-604924715

andrewn:
@hydrosquall I started thinking about how we might do this and [ended up with a branch](https://github.com/andrewn/p5.js-web-editor/tree/chore/extract-dev-server) that mostly does it...

The general idea is:
- Totally remove webpack code from API server
- Create a new server that runs the dev webpack config on port 3000 and serves the page routes
- Print a message telling the user to set `API_URL` in .env to the correct API server instance (localhost or production)

The missing piece is being able to contact the deployed API. At the moment we restrict CORS origins to only `p5js.org`. I've opened PR #1398 to address this.

To run the client only, you `npm run start:client` and to run both `npm start`.

Once we have this separation, then we can start experimenting with other packagers.

What do you think?

hydrosquall:
Hi @andrewn, I think that sounds like a clean approach, as long as there aren't any particular security motivations to keep the CORS restrictions. I think this should be fine especially if someone just wants to be running the site in a "stateless" mode.

PR #1398, opened 2020-04-19
Always allow localhost CORS requests

andrewn:
In order to allow developers to run only the client app code on their machines (see #1348), they need to be able to send requests from the app running on `localhost` to a remote instance of the Editor API.

Our CORS restrictions currently prohibit `localhost` in production so this PR removes that restriction.

I've thought about the security implications of this, and I think it's OK as the API will still require a user to be authenticated before doing anything.

It _might_ be an idea to parallel deploy to a staging server i.e. `staging.editor.p5js.org` and only allow localhost CORS there to protect developers from accidentally doing anything weird. We'd also probably want a copy of mongo too?

--

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,6 +2,7 @@ API_URL=/editor
 AWS_ACCESS_KEY=<your-aws-access-key>
 AWS_REGION=<your-aws-region>
 AWS_SECRET_KEY=<your-aws-secret-key>
+CORS_ALLOW_LOCALHOST=true
 EMAIL_SENDER=<transactional-email-sender>
 EMAIL_VERIFY_SECRET_TOKEN=whatever_you_want_this_to_be_it_only_matters_for_production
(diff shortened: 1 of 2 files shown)


catarak:
Awesome!! Agree that maybe it would make sense to only allow localhost from a staging server. Maybe this could be configured with an environment variable then?

andrewn:
I've added the `CORS_ALLOW_LOCALHOST` env var to handle this. When set to the string `"true"` it will be enabled.

catarak:
> I've added the `CORS_ALLOW_LOCALHOST` env var to handle this. When set to the string `"true"` it will be enabled.

Thank you!! I think you didn't push this change as I'm only seeing one commit on this PR?

andrewn:
@catarak Sorry. Pushed it now!

catarak:
perfect :) i'll add this env to the staging server!

PR #1403, opened 2020-04-23
Enhancement(Esc key hide modal)

### LINKED ISSUE #1312: esc key should hide modal box.
#### Nature of issue?

The box that appears on clicking the add folder/file option onn the editor should close on clicking "esc" key which currently doesn't happens.

shakti97:

Fixes #1312 

I have 2 solution for this enhancement, 
1. add keylistener in every component (File/Folder) and then trigger close modal from those component
2. add global listener for ESC key and close modal from there

I have taken the 2nd approach, as the first looks to add redundant code in all component.
@catarak have a look and suggest changes if any :)

### DIFF
--- a/client/modules/IDE/pages/IDEView.jsx
+++ b/client/modules/IDE/pages/IDEView.jsx
@@ -170,6 +170,14 @@ class IDEView extends React.Component {
       } else {
         this.props.expandConsole();
       }
+    } else if (e.keyCode === 27) {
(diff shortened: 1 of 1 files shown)


catarak:
Thank you!! Looks good to me. I think this is the right approach. 

I think that this state (whether or not these overlays are visible) should be moved from Redux to React, and then it would make sense to put the keyboard handler in the component, but for now, this is okay.

PR #1410, opened 2020-05-01
Write tests for name verification logic (<ToolBar />, <FileNode />)

ghalestrilo:
Some additional tests for these two components, just to make sure the logic works as intended, and to prevent it from breaking in the future.


andrewn:
This is really cool! 🎉 

Have you ever used [React Testing Library](https://testing-library.com/) to write tests instead of Enzyme? I've used it in a couple of projects now and it really helps me to write more accessible components. It moves focus into testing how the user experiences the app via labels and text, rather than testing internal details like state or props.

Was thinking it might be worth moving to this while we don't have a huge number of UI tests in the project.

@catarak?

ghalestrilo:
That's a great idea, I'm all for it!
React Testing Library also seems a lot more future-proof

catarak:
React Testing Library looks awesome! I think it would be worth trying it out in a separate PR and see how it looks compared to the enzyme tests.

catarak:
these tests look good to me for right now!

catarak:
Opened an issue about React Test Framework: #1432

PR #1411, opened 2020-05-01
Fix(overflow option over sticky head issue)

shakti97:

actually its a very small issue so don't want to create a new issue

Issue
In the Example table, the last option has a button that has an arrow-down icon which is scrolling over the header i.e. not obeying the normal behavior which the others are following.
OptionScrolled


catarak:
thank you!!

PR #1415, opened 2020-05-03
Separate Webpack Dev server from API server (fixes #1348)

### LINKED ISSUE #1348: [dev-task] Separate Webpack Dev server from API server

andrewn:
This extracts the webpack dev server into it's own process, run with `npm run start:client`. This allows contributors to just run the web app.

There are a couple of others parts to this:

- [x] staging server URL needs to be configured
- [x] CORS needs to be configured for staging (PR #1398)

### DIFF
--- a/developer_docs/installation.md
+++ b/developer_docs/installation.md
@@ -4,6 +4,41 @@ Follow these instructions to set up your development environment, which you need
 
 ## Manual Installation
 
+There are two types of manual installation. 
+
+The simplest is to run the front-end web app only (A). This connects to a staging API and allows you to make changes to the web app without having to worry about installing the backend API and database.
+
+If you need to change the API, then you'll need to run the full-stack (B).
+
+### A. Front-end web app only
+
+_Note_: The installation steps assume you are using a Unix-like shell. If you are using Windows, you will need to use `copy` instead of `cp`.
+
+1. Install Node.js. The recommended way is to Node through [nvm](https://github.com/nvm-sh/nvm). You can also install [node.js](https://nodejs.org/download/release/v12.16.1/) version 12.16.1 directly from the Node.js website.
+2. [Fork](https://help.github.com/articles/fork-a-repo) the [p5.js Web Editor repository](https://github.com/processing/p5.js-web-editor) into your own GitHub account.
+3. [Clone](https://help.github.com/articles/cloning-a-repository/) your new fork of the repository from GitHub onto your local computer.
+
+   ```
+   $ git clone https://github.com/YOUR_USERNAME/p5.js-web-editor.git
+   ```
+
+4. If you are using nvm, run `$ nvm use` to set your Node version to 12.16.1
+5. Navigate into the project folder and install all its necessary dependencies with npm.
+
+   ```
+   $ cd p5.js-web-editor
+   $ npm install
+   ```
+
+7. `$ cp .env.example .env`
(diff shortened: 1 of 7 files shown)


hydrosquall:
Hi @catarak, finally having some time to look at this project again due to the holiday - what do you think about this change? 

Once this is in place, I could continue with #1257 . This would open the door to us being able to easily have per-branch preview deploys, which should hopefully make it easier for maintainers to review the functionality of PRs without needing to copy the changes locally every time.

catarak:
Thanks for bumping this! Right now, I have the staging server set up with basic auth, which works great with a relatively small number of people testing changes on `develop` but is not super compatible with running the front-end locally and using the staging server API. But I guess this change would allow the two to run separately, so that the front-end could be password protected, and the back-end could be CORS protected.

So I think what needs to happen is that this PR needs to be updated so that `server-webpack-only.js` can be password protected with basic auth, and basic auth needs to be removed from `server.js`. Maybe also the names of these files should be changed—do you know what the conventions are here? Like `server.js` and `app.js`? `api-server.js` and `app-server.js`?

andrewn:
> So I think what needs to happen is that this PR needs to be updated so that `server-webpack-only.js` can be password protected with basic auth, and basic auth needs to be removed from `server.js`. Maybe also the names of these files should be changed—do you know what the conventions are here? Like `server.js` and `app.js`? `api-server.js` and `app-server.js`?

`server-webpack-only.js` would only ever be used when developing the Web Editor without running the API server locally. It runs the Webpack Dev Server, and wouldn't be used in production. When running in production, the server and app would continue to run as it does now, using `'./server/server'`. I don't think `app.js` makes sense as a name, since it's only used for development... perhaps `server/app-dev.js`?

I'm not sure that we'd need `server-webpack-only.js` to have basic auth enabled since it won't be run on the staging server.

I think what we want is:
  1. to have basic auth in front of the staging web app UI - so only those with access can log in to see new features;
  2. the staging API endpoint should be accessible by via CORS, either for the the staging web app or for local developer.

Both of these changes need to be made in `./server/server`.

catarak:
> `server-webpack-only.js` would only ever be used when developing the Web Editor without running the API server locally. It runs the Webpack Dev Server, and wouldn't be used in production. 

oh duh of course! 


> When running in production, the server and app would continue to run as it does now, using `'./server/server'`. I don't think `app.js` makes sense as a name, since it's only used for development... perhaps `server/app-dev.js`?

`app-dev.js` makes sense to me!

> 1. to have basic auth in front of the staging web app UI - so only those with access can log in to see new features;
> 2. the staging API endpoint should be accessible by via CORS, either for the the staging web app or for local developer.
> 
> Both of these changes need to be made in `./server/server`.

This approach makes sense to me!

I've also been in contact with a new tool called [Release](https://releaseapp.io/) which will deploy pull request builds (and do other stuff too), but I still think it's worth it to get this in so that folks can do development and only run the front-end application.

raclim:

Issue #91, opened 2016-09-07
privacy messaging and private/public flag on sketches


aparrish:
I love that there's a publicly-visible list of [sketches for particular users](https://alpha.editor.p5js.org/aparrish/sketches)! But it's not necessarily intuitive for new users that any work they do in the web editor will be public by default. So there should be messaging that tells users that sketches they save might be visible to anyone.

Alternatively/additionally, it would be great to have a public/private (or draft/published, or whatever language you want to use) flag on sketches that determines whether the sketch appears in the user's list of sketches in the public view.

Of course this ticket runs the risk of spiraling out into a full-blown sketch authorization thing, and it's probably worth addressing at some point (if this hasn't already been addressed) things like making sketches that are only visible to the owner of the sketch, or to a group of users specified by the owner, or only visible to authenticated users, etc etc etc. But a public/private flag is the bare minimum necessary in my opinion!

JithinKS97:
Hey @catarak I'm interested in taking up this issue and discuss more about the design aspects of implementing it. Can you provide me some ideas on how should I start and move forward?

nik72619c:
Hi , I am interested in working on this one . @catarak can you please guide me through how shall we start implementing it up ? What is the plan of action and which subtasks are at priority ?

pcgamer1:
I am interested to work on this feature. I would like to know more about the exact requirements and the location of required change on Github.

aldrinjenson:
Seeing that the issue is over 3 and a half years old, it would be really helpful if someone could provide a bit more information on proceeding to fix this. 
Thanks

catarak:
This is being actively worked on by @Khensura21.

dhowe:
Status on this? I see some commits but nothing since May

catarak:
This is coming soon, within the next few months!

basicvisual:
Hi , curious to know if there has been any updates on this ?My idea is to display my sketches without the jibberish experimental code :+1:

SFUMisha:
Hello, has been a little bit since the last activity here. Just wondering if any functionality for privateing sketches has been released yet? or maybe an eta when it could be possible? Would really love this feature, thank you!

EmmanuelPil:
@SFMischa "Would really love this feature, thank you!" Same here!

cgortonpenguinhall:
I would also like to see this completed. Right now, my students can access my entire portfolio, which is not ideal.

Thanks

dhowe:
[...]

PR #1433, opened 2020-05-24
Add private/public flag on sketches (fixes issue #91)

Khensura21:


server/controllers/project.controller.js
```diff
@@ -2,28 +2,28 @@ import archiver from 'archiver';
...
   Project.findById(req.params.project_id, (findProjectErr, project) => {
-    if (!project.user.equals(req.user._id)) {
-      res.status(403).send({ success: false, message: 'Session does not match owner of project.' });
-      return;
-    }
-    if (req.body.updatedAt && isAfter(new Date(project.updatedAt), req.body.updatedAt)) {
-      res.status(409).send({ success: false, message: 'Attempted to save stale version of project.' });
-      return;
-    }
+    // if (!project.user.equals(req.user._id)) {
```
catarak, line 19: Since you commented this out for testing, it should be uncommented for this PR!

server/controllers/project.controller.js
```diff
@@ -140,14 +140,21 @@ export function projectExists(projectId, callback) {
   ));
 }
 
-export function projectForUserExists(username, projectId, callback) {
+export function projectForUserExists(username, authenicatedUser, projectId, callback) {
+  console.log('Just got called');
```
catarak, line 144: Also I'd remove this console.log, sometimes they get left in for testing but I try to remove all of them.

server/routes/project.routes.js
```diff
@@ -6,7 +6,9 @@ const router = new Router();
 
 router.post('/projects', isAuthenticated, ProjectController.createProject);
 
-router.put('/projects/:project_id', isAuthenticated, ProjectController.updateProject);
+// router.put('/projects/:project_id', isAuthenticated, ProjectController.updateProject);
```
catarak, line 9: I'd also uncomment this and remove line 11!

server/routes/server.routes.js
```diff
@@ -34,7 +34,7 @@ router.get('/:username/sketches/:project_id/add-to-collection', (req, res) => {
 });
 
 router.get('/:username/sketches/:project_id', (req, res) => {
-  projectForUserExists(req.params.username, req.params.project_id, exists => (
+  projectForUserExists(req.params.username, req.user, req.params.project_id, exists => (
```
catarak, line 37: `projectForUserExists` is called a few times in the application, have the other times that it's called been updated to work with the new parameters?

server/controllers/project.controller/getProjectsForUser.js
```diff
@@ -17,16 +17,26 @@ function getProjectsForUserName(username) {
         reject(new UserNotFoundError());
         return;
       }
-
-      Project.find({ user: user._id })
+      const isUserAuthorized = authenicatedUser && username === authenicatedUser.username;
+      let query;
+      if (!isUserAuthorized) {
```
jamesrweb, line 22: I would remove this whole `if` to just make it a `const` since the whole point of the if is toggling one field to false, why not just make it true if the user is authorised and false if not which is basically what you're doing anyway:

```js
const query = {
      user: user._id,
      isPrivate: isUserAuthorized
};
```

client/modules/IDE/components/Preferences.jsx
```diff
@@ -229,6 +230,25 @@ class Preferences extends React.Component {
...
+                <input
+                  type="checkbox"
+                  onChange={(event) => {
+                    this.props.setIsPrivate(event.target.checked);
+                  }}
+                  aria-label="privacy output on"
+                  name="privacy output"
+                  id="privacy-output-on"
+                  value="On"
+                  checked={(this.props.isPrivate)}
```
jamesrweb, line 246: Unnecessary `()` brackets, consider removing them or alternatively wrapping in a `Boolean()` call if it's really necessary.

client/modules/IDE/actions/project.js
```diff
@@ -66,6 +59,30 @@ export function getProject(id, username) {
...
+    const { id } = state.project;
+    console.log(value);
+    axios.put(`${ROOT_URL}/projects/${id}`, { isPrivate: value }, { withCredentials: true })
+      .then((response) => {
+        if (response.status === 200) {
+          console.log('success');
+        }
+      })
+      .catch((response) => {
+        console.log(response);
```
jamesrweb, line 78: Extra `console.log` calls should be removed such as here, line 74 and line 70 for example.

Khensura21:
Will get to this today, and get back to you.

Thanks,

Khensu-Ra

server/routes/server.routes.js
Khensura21: I am not sure, how should I check that?

server/controllers/project.controller/getProjectsForUser.js
jamesrweb: Exactly 😄

catarak: I agree with you @Khensura21, that I don't think @jamesrweb's suggestion works for the behavior we want!

jamesrweb: But the condition is just saying if the user is authorised then the isPrivate will be false, otherwise it should be true, right? Seems strange to me to have a conditional to remove a key instead of flipping it’s value, or?

server/routes/server.routes.js
catarak: What I would do is search all files for instances where this function is being called. I do this in VSCode by hitting <kbd>Shift</kbd>+<kbd>Command</kbd>+<kbd>F</kbd>, then typing in the function name `projectForUserExists`, and then it should list all of the instances!

server/controllers/project.controller/getProjectsForUser.js
catarak: This code is a MongoDB query, not setting or unsetting any values. Taking a close look at this again, I think there's actually a bug that's probably my fault from when we were working on this together. On line 25, I think we actually want to return all sketches where `isPrivate` is false or undefined, therefore (i think!!! definitely test this) lines 23-26 should be:
```js
        query = {
          user: user._id,
          $or: [
                { isPrivate: false },
                { isPrivate: { $exists: false } }
           ]
        };
```

jamesrweb: Fair enough, sounds fair with that being the case, good to know! 😁

server/routes/server.routes.js
Khensura21: I realized `projectForUserExists` function definition does not include the parameter that handles the `req.user` argument that's being passed on line 37. Not sure what happened to the code that was written to handle it

server/controllers/project.controller/getProjectsForUser.js
Khensura21: Sounds good. I will look into this and let you know!

Khensura21: Sounds good. I will check this out and let you know!

raclim:

PR #1436, opened 2020-05-26
Fixes #1409 - adds username to url when fetching sketch from editor API

### LINKED ISSUE #1409: When fetching sketches via slug, API returns first sketch that matches slug
#### Nature of issue?

- Found a bug

#### Details about the bug: 

- Web browser and version:  Chrome
- Operating System:  MacOSX

As reported in ml5js/ml5-library#927:

> When one navigates to https://editor.p5js.org/ml5/sketches/ImageModel_TM which is the URL that should open up the default teachable machine URL from the ml5 account, for some reason the URL opens up to another user's project 😬
>
> I think this may have to do with the way that the URL handling occurs on the p5 web editor... my hunch is that if we run our batch update/upload script and another project in the web editor database has the same name, then that existing project takes that named URL and not the typical id string maybe?
>
> We definitely should try to figure out why this is occurring so that people using teachable machine don't get confused as to why they are being sent to this existing project!

Solution determined in https://github.com/ml5js/ml5-library/issues/927#issuecomment-622175396:

>I noticed that the web editor is making an API request to `https://editor.p5js.org/editor/projects/ImageModel_TM` to fetch the sketch. So it would make sense that it would be finding the first sketch with the name `ImageModel_TM`.
>
> I think the way to fix this would be to change the API request so that the username is in the request URL. This hadn't come up before since most sketches are resolved by ID, but by exposing some URL via slug this issue shows itself :). I think the URL should be changed to `/<username>/projects/<slug>`.

catarak:

### DIFF
--- a/client/modules/IDE/actions/project.js
+++ b/client/modules/IDE/actions/project.js
@@ -49,10 +49,10 @@ export function setNewProject(project) {
   };
 }
 
-export function getProject(id) {
+export function getProject(id, username) {
   return (dispatch, getState) => {
     dispatch(justOpenedProject());
-    axios.get(`${ROOT_URL}/projects/${id}`, { withCredentials: true })
+    axios.get(`${ROOT_URL}/${username}/projects/${id}`, { withCredentials: true })
       .then((response) => {
         dispatch(setProject(response.data));
         dispatch(setUnsavedChanges(false));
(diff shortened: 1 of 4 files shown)


catarak:
thanks for testing :) i'm gonna merge this!!

joeyklee:
Bravo!!! 🚀💯🥳

PR #1452, opened 2020-06-08
Helpers for config and API client

andrewn:
I'm trying to add Storybook stories for the `FileNode` component but this isn't possible because it imports Redux actions which rely on reading `window.process.env`, which throws an error in the storybook environment. 😢 

I noticed that the code for reading the env var is duplicated throughout the app, so I've refactored into a helper called `getClient(key)`. This also abstracts where the var is being read from, so in future we can read it from elsewhere, or try a few different places (local storage, then the env) without changing the calling code.

The biggest user of the env var is the `API_URL` in the redux actions. I've created `apiClient` which is an instance of `axios` that provides a single place to configure the `API_URL` and basic config that we want to apply to all communication to the editor API. 

### DIFF
--- a/client/components/Nav.jsx
+++ b/client/components/Nav.jsx
@@ -10,13 +10,12 @@ import * as projectActions from '../modules/IDE/actions/project';
 import { setAllAccessibleOutput } from '../modules/IDE/actions/preferences';
 import { logoutUser } from '../modules/User/actions';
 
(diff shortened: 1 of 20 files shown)


andrewn:
I've changed `getConfig` so it doesn't warn when in the test environment. 

I think this is good enough until we start writing tests that require these config values.

catarak:
works for me!

PR #1459, opened 2020-06-15
Add Editor component to the Mobile IDE View

ghalestrilo:
This PR mounts the `<Editor />` component on the Mobile IDE View (`/mobile`), and places a button for exiting the mobile view.

Functionality is only partially implement, as expected: running the code will require a new overlay component.

### DIFF
--- a/client/modules/IDE/pages/IDEViewMobile.jsx
+++ b/client/modules/IDE/pages/IDEViewMobile.jsx
@@ -1,14 +1,33 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
-import { Link } from 'react-router';
+// import { Link } from 'react-router';
+import { connect } from 'react-redux';
+import { withRouter } from 'react-router';
+import { useState } from 'react';
 
+// Imports to be Refactored
+import { bindActionCreators } from 'redux';
+import * as FileActions from '../actions/files';
+import * as IDEActions from '../actions/ide';
+import * as ProjectActions from '../actions/project';
+import * as EditorAccessibilityActions from '../actions/editorAccessibility';
+import * as PreferencesActions from '../actions/preferences';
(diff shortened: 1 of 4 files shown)


client/modules/IDE/pages/IDEViewMobile.jsx
```diff
@@ -33,16 +79,217 @@ Screen.propTypes = {
...
-  </Screen>
-);
+  return (
+    <Screen>
+      <Header>
+        <div>
+          <h2>{project.name}</h2>
+          <h3>{selectedFile.name}</h3>
+        </div>
+        <Icon href="/">
```
catarak, line 99: You should import `CloseIcon` from `client/common/icons.jsx`, and wrap it in a `<Link>` or `<button>` depending on what you want it to do. And then, because this icon has no text but performs an action, it should have the attribute `aria-label="close header"` (or whatever the alt text should be.

ghalestrilo: Ok, will do that!

client/modules/IDE/pages/IDEViewMobile.jsx
```diff
@@ -92,18 +91,14 @@ const IDEViewMobile = (props) => {
   return (
     <Screen>
       <Header>
+        <Link to="/" style={{ width: '3rem', marginRight: '1.25rem' }}>
```
catarak, line 94: Since we're using styled components, I think it makes sense to not use any inline CSS like this! You also shouldn't need `aria-label` or `aria-hidden` on the `CloseIcon`, as `aria-label` should be added to the `<Link>`, see https://www.scottohara.me/blog/2019/05/22/contextual-images-svgs-and-a11y.html

ghalestrilo: Makes sense. Just updated this!

PR #1467, opened 2020-06-19
Mobile Sketch Preview Screen

ghalestrilo:
Part of the Mobile UI project. This PR adds a (working!) Mobile Sketch Preview screen (/mobile/preview), which renders the canvas with the project files and resizes it to properly fit the mobile view.

How to test:
1. Run the app locally with the `MOBILE_ENABLED` flag set to `true` (`MOBILE_ENABLED=true npm start`)
  - You should be able to access `/mobile`
  - You should be able to edit code
  - Clicking the play icon should run the sketch and you to the preview screen screen
  - You should be able to go back to the code screen, and repeat the process without errors
  - Clicking the exit icon should take you to the default editor
2. Run the app without the flag enabled
  - It should work as usual
  - `/mobile` routes should not be accessible


client/modules/Mobile/MobileSketchView.jsx
```diff
@@ -0,0 +1,200 @@
...
+  const {
+    setTextOutput, setGridOutput, setSoundOutput,
+    endSketchRefresh, stopSketch,
+    dispatchConsoleEvent, expandConsole, clearConsole,
+    setBlobUrl,
+  } = props;
+
+  const { preferences, ide } = props;
+
+  useEffect(() => {
```
andrewn, line 56: This should be removed if the effect isn't doing anything?

client/modules/Mobile/MobileSketchView.jsx
```diff
@@ -0,0 +1,200 @@
...
+  //   console.log(params);
+  //   getProject(params.project_id, params.username);
+  });
+
+  return (
+    <Screen>
+      <Header>
+        <IconLinkWrapper to="/mobile" aria-label="Return to original editor">
```
andrewn, line 64: Can we use the common Button component?

client/modules/Mobile/MobileSketchView.jsx
```diff
@@ -0,0 +1,200 @@
...
+          <ExitIcon viewBox="0 0 16 16" />
+        </IconLinkWrapper>
+        <div>
+          <h2>{projectName}</h2>
+          <h3><br /></h3>
```
andrewn, line 69: What's the purpose of this mark-up?

`<h3><br /></h3>`

client/routes.jsx
```diff
@@ -21,7 +22,11 @@ const checkAuth = (store) => {
   store.dispatch(getUser());
 };
 
+// This short-circuit seems unnecessary - using the mobile <Switch /> navigator (future) should prevent this from being called
```
andrewn, line 25: Can we mark this comment as "TODO" so we can find it later?

andrewn:
I ran it on my iPhone and it works great! But the preferences and play icons are a bit small compared the the exit icon. They're a bit hard to tap

iphone

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -22,6 +22,24 @@ import {
 import { hijackConsoleErrorsScript, startTag, getAllScriptOffsets }
   from '../../../utils/consoleUtils';
 
+
+// Kept inside class just for linter's
```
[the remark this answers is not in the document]
ghalestrilo: That shouldn't be there! Thanks for catching it.
It was a comment on a cryptic eslint rule which prevented me from removing this method from inside the class (outdated)

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -30,46 +48,17 @@ class PreviewFrame extends React.Component {
...
-      this.renderSketch();
-      return;
-    }
-
-    if (this.props.fullView && this.props.files[0].id !== prevProps.files[0].id) {
-      this.renderSketch();
-    }
-
+    if (shouldRenderSketch(this.props, prevProps)) this.renderSketch();
     // small bug - if autorefresh is on, and the usr changes files
```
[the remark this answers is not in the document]
ghalestrilo: I think the commend holds, since I merely moved code around. The logic is pretty much the same

client/modules/Mobile/MobileSketchView.jsx
ghalestrilo: It just heightens the header by an `h3` height, to keep the `h2` tag aligned with the previous screen - couldn't think of a better way to do this...

client/routes.jsx
ghalestrilo: Will do!

client/modules/IDE/pages/IDEViewMobile.jsx
```diff
@@ -20,80 +21,34 @@ import { getHTMLFile } from '../reducers/files';
...
-  background: ${background};
-  color: ${textColor};
-  padding: ${remSize(12)};
-  padding-left: ${remSize(32)};
-  z-index: 1;
-
-  bottom: 0;
-`;
+import PreferencesIcon from '../../../images/preferences.svg';
+import PlayIcon from '../../../images/triangle-arrow-right.svg';
```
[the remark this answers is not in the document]
ghalestrilo: True!

client/modules/IDE/pages/IDEViewMobile.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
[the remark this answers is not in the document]
ghalestrilo: Wow, thanks for seeing this. Two warnings will be fixed in future PRs, I'll ignore them in particular

client/components/mobile/Header.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import styled from 'styled-components';
+import { prop, remSize } from '../../theme';
+
+const background = prop('Button.default.background');
```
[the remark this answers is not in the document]
ghalestrilo: Great point. I made a `Panel` color prop

client/components/mobile/IconButton.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
```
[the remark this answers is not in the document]
ghalestrilo: Yes, and it even looks better now

client/components/mobile/Header.jsx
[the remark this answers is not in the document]
ghalestrilo: Great Point. I made a Panel object in theme.js for that

client/modules/Mobile/MobileSketchView.jsx
andrewn: We might have to use CSS to add padding to the h2? And maybe set the previous screens header to a know height?

Using `h3` is this way is not semantic especially as screen-readers allow the user to list and navigate via the headings in the page, which will be empty.

client/theme.js
```diff
@@ -88,6 +88,13 @@ export default {
     Icon: {
       default: grays.middleGray,
       hover: grays.darker
+    },
+    Panel: {
```
andrewn, line 92: These are same colours as the buttons, so we'll have a problem if someone later puts a Button on top of a Panel as they'll be the same.

Shall we take the colours [from the console header component](https://github.com/processing/p5.js-web-editor/blob/develop/client/styles/components/_console.scss)? Cassie made accessibility colour changes a a while back to ensure the contrast worked across all the themes.

So:

`Panel.default.foreground` should be the value of `console-header-color`
`Panel.default.background` should be the value of  `console-header-background-color`
`Panel.default.border` should be the value of  `ide-border-color`

The [colours are here](https://github.com/processing/p5.js-web-editor/blob/67f98246423a9bd87fb671d7f2f1e9929c27c5e8/client/styles/abstracts/_variables.scss).

I know it's a bit of a pain to do this, but it'll really help with the move to styled-components and away from SCSS and helping to make the design more consistent.

ghalestrilo: True, but these are the actual colors of the mobile UI. I think I'll create a `Panel: { mobile: {} }` attribute and keep these colors there, so the default `Panel` colors can be changed.
Do you think that makes sense?

andrewn: If possible, it would be great to have the same set of colours work for mobile and the legacy UI. My feeling is branching the theme into _Panel.mobile_ and _Panel.notMobile_ creates more code for us to maintain.

Your right that maybe this isn't a reusable panel colour, so if it's not possible to find common colours, then my preference is that we name the variable something other than Panel.

@catarak, what do you think?

ghalestrilo: I investigated this a bit, and It's an important decision that goes beyond the scope of the PR, so I'll create an issue and we can discuss it there!

client/modules/Mobile/MobileSketchView.jsx
ghalestrilo: Good point. I'm fixing this on the next PR, and improving the `<Header />` component as well

client/theme.js
andrewn: Ok, good idea.

So shall we call this MobilePanel or something for now and we can get this merged in?

catarak: I agree with @andrewn, that it would be helpful if possible to not have mobile specific colors. 

@ghalestrilo and I talked this over and thought it would make sense to open a new issue/PR with respect to the panel and button color contrasts!

client/components/mobile/IconButton.jsx
```diff
@@ -0,0 +1,24 @@
...
+> svg {
+  width: 100%;
+  height: 100%;
+}
+`;
+
+const IconButton = props => (<ButtonWrapper
+  iconBefore={props.children}
+  kind={Button.kinds.inline}
+  {...{ ...props, children: null }}
```
catarak, line 17: Do you need to put `children: null` here? Isn't `children` null here if you just don't put anything in between `<ButtonWrapper></ButtonWrapper>`?

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -212,6 +200,12 @@ class PreviewFrame extends React.Component {
     this.addLoopProtect(sketchDoc);
     sketchDoc.head.insertBefore(consoleErrorsScript, sketchDoc.head.firstElement);
 
+    if (this.props.forceFullWidth) {
```
catarak, line 203: I saw you reversed this, but maybe this should be removed entirely? Including the `forceFullWidth` prop?

client/modules/IDE/pages/IDEViewMobile.jsx
```diff
@@ -20,93 +20,61 @@ import { getHTMLFile } from '../reducers/files';
...
 
   return (
     <Screen>
       <Header>
-        <IconLinkWrapper to="/" aria-label="Return to original editor">
-          <ExitIcon />
-        </IconLinkWrapper>
-        <div>
+        <IconButton to="/mobile" aria-label="Return to original editor">
+          <ExitIcon viewBox="0 0 16 16" />
```
catarak, line 60: Are you using viewbox here to change the size of the icon? I would prefer if you explicitly set the width and height, and maybe this needs to be added to `client/common/icons.jsx`, which could be done in a separate PR.

client/modules/IDE/pages/IDEViewMobile.jsx
```diff
@@ -20,93 +20,61 @@ import { getHTMLFile } from '../reducers/files';
...
+        </IconButton>
+        <div style={{ marginLeft: '1rem' }}>
```
catarak, line 62: if possible i would prefer to not have inline styles!

client/theme.js
ghalestrilo: Nice! I'll update this :)

client/components/mobile/IconButton.jsx
ghalestrilo: Yes, it's null if I don't pass anything, but in this case I wanted to pass `children` into the `iconBefore` prop, and not into the body of the component.
I thought of a better solution, so I'll change this soon

client/modules/IDE/components/PreviewFrame.jsx
ghalestrilo: Not sure why it's still there, I thought I had already removed it!

client/modules/IDE/pages/IDEViewMobile.jsx
ghalestrilo: The [next PR](https://github.com/processing/p5.js-web-editor/pull/1472/files#diff-d28178cfe0c42c2d9779fdbe52f4ad3c) fixes this by improving the structure of the `<Header />` component

client/modules/IDE/pages/IDEViewMobile.jsx
ghalestrilo: I'll remove the `viewbox` prop for now, and find a way to resize the icon in a future PR, to minimize conflicts with the other branch

client/components/mobile/IconButton.jsx
ghalestrilo: I introduced an `icon` prop, now it looks a lot cleaner

PR #1472, opened 2020-06-23
Mobile Preferences Screen Prototype

ghalestrilo:
Part of the Mobile UI project. This PR implements a screen to access and control editor settings.

This PR requires #1467 to be merged

How to test:
1. Run the app locally with the `MOBILE_ENABLED` flag set to `true` (`MOBILE_ENABLED=true npm start`)
  - You should be able to access `/mobile`
  - You should be able to edit code
  - Clicking the gear icon should take you to the settings screen
  - You should be able to change settings and go back to the editor
  - Preview Screen should be working as well
2. Run the app without the flag enabled
  - It should work as usual
  - `/mobile` routes should not be accessible


client/components/mobile/IconButton.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import styled from 'styled-components';
+import { prop, remSize } from '../../theme';
+
+const textColor = prop('primaryTextColor');
+
+const IconButton = styled.button`
+width: 3rem;
```
catarak, line 8: Would it make sense to extend from `client/common/Button`? No worries if that's overcomplicated. Either way, I'd like this to use `remSize` so the measurement can easily be read as pixels.

client/components/mobile/Selector.jsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
```
catarak, line 1: I think this file could have a more descriptive/specific name since "Selector" feels pretty generic, where this component is specifically being used for a Preference selection.

client/modules/IDE/components/PreviewFrame.jsx
```diff
@@ -212,6 +201,12 @@ class PreviewFrame extends React.Component {
     this.addLoopProtect(sketchDoc);
     sketchDoc.head.insertBefore(consoleErrorsScript, sketchDoc.head.firstElement);
 
+    if (this.props.forceFullWidth) {
```
catarak, line 204: I understand where you're coming from with this, but I'm not sure it's a good idea. It will distort the canvas and I think folks will be really confused as to what's happening. I think ideally, the default canvas size for mobile would be `windowWidth` or `windowHeight` but I'm not sure the details on how to implement that in a way that doesn't feel confusing to users.

client/modules/IDE/pages/MobileIDEView.jsx
```diff
@@ -104,9 +58,23 @@ const IDEViewMobile = (props) => {
           <h2>{project.name}</h2>
           <h3>{selectedFile.name}</h3>
         </div>
+
+        <div style={{ marginLeft: '2rem' }}>
+          <Link to="/mobile/preferences">
+            <IconButton onClick={() => setOverlay('preferences')}>
+              <PreferencesIcon focusable="false" aria-hidden="true" />
```
catarak, line 65: You should import this icon to  ` client/common/icons` so you don't have to manually put in the accessibility attributes! Same with the `PlayIcon`.

client/components/mobile/IconButton.jsx
ghalestrilo: Sure! I updated it on

client/components/mobile/Selector.jsx
ghalestrilo: True. I renamed it to `<PreferencePicker />` to be more descriptive

client/modules/IDE/components/PreviewFrame.jsx
ghalestrilo: I think it's best to leave this behavior unimplemented for now while we work out a better strategy. The preview screen will render the sketch with original dimensions until then.

catarak:
When trying this on my phone, the "On" and "Off" text felt too small. In the light theme, the contrast between "On" being selected or "Off" being selected feels too subtle, though maybe that needs to be addressed in a separate PR. 

A design idea would be to make the "Settings        X" at the top into a bar, with a separate background color, maybe the same color as on the main page (the panel color).

ghalestrilo:
I increased the option label font-size to `16px` on the mobile view, that should be a bit better
About the `Header`, I think we need to test it on the second iteration. I thought leaving it transparent conveyed the idea that it's a "temporary" Screen / Modal, with a different purpose from the ones with solid headers.

catarak:
> About the `Header`, I think we need to test it on the second iteration. I thought leaving it transparent conveyed the idea that it's a "temporary" Screen / Modal, with a different purpose from the ones with solid headers.

Sounds good! I think it's worth testing, as when I opened the screen, the hierarchy between "Settings" and "General Settings" felt confusing to me as they are both the same font size and styling.

catarak:
@ghalestrilo looks great! I just pushed a commit that re-added the Redux DevTools to App.jsx... maybe this could be done with an environment variable so that you don't have to comment/uncomment it all of the time?

PR #1477, opened 2020-06-28
Use React Testing Library instead of Enzyme for tests

### LINKED ISSUE #1432: Investigate replacing Enzyme with React Testing Library
#### Nature of issue?

There are still a relatively small number of tests in this project. Because of this, it makes sense to investigate alternative testing frameworks such as [React Testing Library](https://testing-library.com/docs/react-testing-library/intro) as they might be more stable in the long run.

andrewn:
Fixes #1432.

This replaces [enzyme](https://enzymejs.github.io/enzyme/) with [React Testing Library](https://testing-library.com/docs/react-testing-library/intro). 

The `server` and `client` code is run as separate [Jest projects](https://jestjs.io/docs/en/configuration#projects-arraystring--projectconfig) which means:
- server code uses the node.js Jest environment
- server tests are configured via `server/jest.setup.js`
- client code uses the jsdom Jest environment (emulating a browser)
- client test are configured via `client/jest.setup.js`

I've ported the Toolbar, Nav and FileNode tests to use React Testing Library. By design, it forces a focus on how a user will interact with components. e.g. selecting buttons by text content or associated <label>.

I've also flattened the nested `describe` blocks as I found it difficult to follow the flow of tests.


package.json
```diff
@@ -13,6 +13,9 @@
     "build:server": "cross-env NODE_ENV=production webpack --config webpack/config.server.js",
     "build:examples": "cross-env NODE_ENV=production webpack --config webpack/config.examples.js",
     "test": "jest",
+    "xxtest": "npm run test:server && npm run test:client",
```
catarak, line 16: Is this the old command to run the tests? Should this be removed?

catarak:
This looks great to me! I do really like the switch from testing the internal props/state to the UI itself.

package.json
andrewn: Ah, thanks so much for spotting them! This was an experiment before I found out about Jest projects.

I've removed them!

PR #1478, opened 2020-06-28
Prototype with proposed i18n architecture 

### LINKED ISSUE #1434: Web Editor in Spanish / Consistency with Languages Selection in p5 website
**- Existing feature enhancement / Request for Comments**

**- Actual behaviour**

As part of my GSoC project, I want to add the Spanish translation to the web editor.

I want to add functionality in the web editor to choose the language of the editor (Spanish version as the goal).
This will be based on:
https://github.com/processing/p5.js-web-editor/issues/595

Now, I want to be consistent with the p5 website, which currently uses a toogle button to choose between 3 languages.

There is  an issue for the website where they deal with a dropdown approach instead of the toogle button:
https://github.com/processing/p5.js-website/issues/569
https://github.com/processing/p5.js-website/pull/583

So I think that I'd follow that approach in the web editor.
Now, I want to ask the community where they think that it is better to put that dropdown in the web editor?

**-Screenshot**

At this moment I'm using a new tab in the preferences section

screen1

and as a first approach I used the toogle button to choose between languages,  like in the website, but given the potential expansion of languages that approach doesn't look good. 

screen2

So, what do you think about this particular location in the web editor which , or do you think that somehow a more exposed position would be better? Like the one in the pull request and locate the dropdown in the first level of the GUI.

 **Maybe for newcomers it is easier just to add it in the top right section of the GUI, maybe positioned left to the current gear icon.** 

**- Would you like to work on the issue?**

Yes, please assign me.

### LINKED ISSUE #1447: i18n library
**Nature of issue?**
New feature request / Request For Comments

**New feature details:**
As a new feature request, the Internationalization of the web editor and in particular, its localization in Latin american Spanish (see #1434) depends in choosing a i18n library.
Currently, there is no library included in the web editor, and while p5 website is using one, maybe it is better to move to another library given the current stack.
I propose the following libraries.

**Possible choices**
Extending the discussion about two possible libraries mentioned in #595 the possible choices are:

- [i18njs](https://i18njs.com/) (Same as in p5 website)
- [React-i18next](https://github.com/i18next/react-i18next) 
- [React-intl](https://formatjs.io/docs/react-intl/) (Same as in MERN example)
- [React-intl-Redux](https://github.com/ratson/react-intl-redux#readme)

**Considerations**
- Work with React
- Display "live" versions of the components so that changes to them are immediately visible
- First case to test Internationalization/Location is Latinamerican Spanish(see #1434)
- Depending in the amount of labels, async loading of labels would be nice
- Easy to add new languages
- Facilitating compliance with Unicode 
- Minimizing the use of concatenated strings 
- Avoiding hard-coded text 
- Designing for independence from cultural conventions (e. g., date and time displays),

oruburos:
Fixes #1434 , Fixes #1447 
- [x] has no linting errors (`npm run lint`)
- [x] is descriptively named and links to an issue number, i.e. `Fixes #123`


client/components/Nav.jsx
```diff
@@ -244,87 +257,89 @@ class Nav extends React.PureComponent {
...
-          <ul className="nav__dropdown">
-            <li className="nav__dropdown-item">
-              <button
-                onClick={this.handleNew}
-                onFocus={this.handleFocusForFile}
-                onBlur={this.handleBlur}
-              >
-                New
-              </button>
+      <Translation ns="Menu">
```
andrewn, line 260: What do you think about using the [`withTranslation` HoC](https://react.i18next.com/latest/withtranslation-hoc) instead of this `Translation` component?

I think it would be clearer to see what's actually being changed in these files as we'd wrap Nav in `withTranslation('Menu')(Nav)` at the bottom of the file and then use `this.props.t('key')`.

This would avoid indenting all the files and would be less changes in the PR diff views.

client/i18n.js
```diff
@@ -0,0 +1,37 @@
...
+    debug: true,
+    backend: options,
+    getAsync: false,
+    initImmediate: false,
+    useSuspense: true,
+    whitelist: availableLanguages,
+    interpolation: {
+      escapeValue: false
+    },
+    saveMissing: true,
```
andrewn, line 34: `Menu.Open` is missing, but I'm not sure what `saveMissing` actually does. Would be great if it logged it somewhere!

Screenshot 2020-06-29 18 07 53

client/index.jsx
```diff
@@ -25,6 +26,8 @@ const App = () => (
 const HotApp = hot(App);
 
 render(
-  <HotApp />,
+  <Suspense fallback={(<div>Loading translations</div>)}>
```
andrewn, line 29: Let's use the [`<Loader />`](https://github.com/processing/p5.js-web-editor/blob/master/client/modules/App/components/loader.jsx) here :-)

catarak:
Amazing work! This is so exciting ✨

client/i18n.js
```diff
@@ -0,0 +1,37 @@
+import i18n from 'i18next';
+import { initReactI18next } from 'react-i18next';
+// import detector from 'i18next-browser-languagedetector';//
+// import Fetch from 'i18next-fetch-backend';
+// import Fetch from 'i18next-fetch-backend';
```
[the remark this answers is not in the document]
oruburos: it has been addressed in the new commit

client/i18n.js
```diff
@@ -0,0 +1,37 @@
...
+  allowMultiLoading: false, // set loadPath: '/locales/resources.json?lng={{lng}}&ns={{ns}}' to adapt to multiLoading
+};
+
+i18n
+  .use(initReactI18next) // pass the i18n instance to react-i18next.
+  .use(Backend).init({
+    lng: 'en',
+    defaultNS: 'menu',
+    fallbackLng, // if user computer language is not on the list of available languages, than we will be using the fallback language specified earlier
+    debug: true,
```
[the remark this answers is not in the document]
oruburos: it has been addressed in the new commit

client/modules/IDE/components/Toast.jsx
```diff
@@ -2,15 +2,17 @@ import PropTypes from 'prop-types';
...
 
 import ExitIcon from '../../../images/exit.svg';
 
 function Toast(props) {
+  const { t } = useTranslation('common');
   return (
     <section className="toast">
       <p>
-        {props.text}
+        {t(props.text)}
```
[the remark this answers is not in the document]
andrewn: After talking to @oruburos, `props.text` will be passed through and displayed. This is fine in the short-term since the Editor UI will continue to work whilst we translate the ap in follow-up PRs.

server/server.js
```diff
@@ -79,6 +79,7 @@ app.options('*', corsMiddleware);
 app.use(Express.static(path.resolve(__dirname, '../dist/static'), {
   maxAge: process.env.STATIC_MAX_AGE || (process.env.NODE_ENV === 'production' ? '1d' : '0')
 }));
+app.use('/translations', Express.static('translations/locales/'));
```
catarak, line 82: Rather than serve the translations statically, would it instead be better to bundle them in with the webpack build? Unfortunately, serving the files this way breaks the application in production, because they are not copied in the [Docker](https://github.com/processing/p5.js-web-editor/blob/develop/Dockerfile) build. On the staging server, fetching the translations is returning a 404 error.

cc @andrewn

catarak: Or, the Dockerfile needs to be updated to copy the translations.

catarak: Or or, the [webpack prod configuration](https://github.com/processing/p5.js-web-editor/blob/develop/webpack/config.prod.js) needs to be updated so that it copies the files into `dist/static`

andrewn: Ah, I didn’t realise that, sorry.

The translations shouldn’t be in the bundle as we don’t want users to have to download languages they won’t use. But I guess copying the translations into `dist/static` would be simplest.

catarak: No worries at all! Let me know if y'all need any help with this.

PR #1491, opened 2020-07-10
Webpack config in dev and prod include copywebpack for static transla…

oruburos:
This PR copies the translations in dist/static as suggested in 
https://github.com/processing/p5.js-web-editor/pull/1478#discussion_r450391729

Fixes #1478

* [x] is descriptively named and links to an issue number, i.e. `Fixes #1478`

### DIFF
--- a/client/i18n.js
+++ b/client/i18n.js
@@ -7,7 +7,7 @@ const fallbackLng = ['en-US'];
 const availableLanguages = ['en-US', 'es-419'];
 
 const options = {
-  loadPath: '/translations/{{lng}}/translations.json',
+  loadPath: 'locales/{{lng}}/translations.json',
   requestOptions: { // used for fetch, can also be a function (payload) => ({ method: 'GET' })
     mode: 'no-cors'
   },
--- a/server/server.js
+++ b/server/server.js
@@ -75,11 +75,11 @@ app.use(corsMiddleware);
 app.options('*', corsMiddleware);
 
 // Body parser, cookie parser, sessions, serve public assets
-
+app.use('/translations', Express.static('translations/locales/'));
 app.use(Express.static(path.resolve(__dirname, '../dist/static'), {
   maxAge: process.env.STATIC_MAX_AGE || (process.env.NODE_ENV === 'production' ? '1d' : '0')
 }));
-app.use('/translations', Express.static('translations/locales/'));
+
 app.use(bodyParser.urlencoded({ limit: '50mb', extended: true }));
 app.use(bodyParser.json({ limit: '50mb' }));
 app.use(cookieParser());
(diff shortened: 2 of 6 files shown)


client/i18n.js
```diff
@@ -7,7 +7,7 @@ const fallbackLng = ['en-US'];
 const availableLanguages = ['en-US', 'es-419'];
 
 const options = {
-  loadPath: '/translations/{{lng}}/translations.json',
```
andrewn, line 10: I think this should stay as `'/translations/{{lng}}/translations.json'`, but the static file config in `server/server.js` needs to change...

server/server.js
```diff
@@ -75,11 +75,11 @@ app.use(corsMiddleware);
 app.options('*', corsMiddleware);
 
 // Body parser, cookie parser, sessions, serve public assets
-
+app.use('/translations', Express.static('translations/locales/'));
```
andrewn, line 78: The path needs to point to the location of the *copied* translations...

`app.use('/translations', Express.static(path.resolve(__dirname, '../dist/static/translations'));`

catarak: @andrewn given your suggested fix here, I think this line may be unnecessary, given [L79](https://github.com/processing/p5.js-web-editor/pull/1491/files/78f87b6ec17c7546ecb841c870bccab59c70ca28#diff-ea75b530d74f94670cea053feb8b961bR79), which serves everything in `/dist/static` to `/`, which I think would include subfolders. Definitely double check and test this though!

andrewn: @catarak, you're right that L79 will serve `/dist/static/translations` but it will sent with a max-age of 1 day. Everything bundled by WebPack uses the file's content hash in the filename (e.g. `app.c41b13c86f9f5577f9cc.js`). The browser can cache this for a long time (in principle, forever) since it'll never change. If the bundle's updated, it has a new filename and the browser fetches the file at the new URL.

For now, we're not versioning the translations assets using a content hash in the filename so we should serve them from `/dist/static/translations` **but** with a different cache strategy. Using an E-Tag with no `max-age` means that the browser can make a quick conditional GET request to check if the file's been updated. If it hasn't been updated, no data is received but if it has, it will be sent the latest file.

In future we should probably use the same content-hash-in-filename strategy for these files too, but it's a bit more complex so this is a fix in the short-term that gets staging working again.

@oruburos this line (L78) needs to be:

```js
app.use(
  '/locales',
  Express.static(
    path.resolve(__dirname, '../dist/static/locales'),
    {
      // Browsers must revalidate for changes to the locale files
      // It doesn't actually mean "don't cache this file"
      // See: https://jakearchibald.com/2016/caching-best-practices/
      setHeaders: res => res.setHeader('Cache-Control', 'no-cache')
    }
  )
);
```

catarak:
This looks good to me! Thanks for working on this ✨

Issue #1492, opened 2020-07-13
Translations missing in Nav


catarak:
#### Nature of issue?

- Found a bug

#### Details about the bug: 

- Web browser and version:  Chrome
- Operating System:  Mac OSX
- Steps to reproduce this bug:
As an authenticated user, the options in the right-most dropdown aren't rendered correctly. 

<img width="279" alt="Screen Shot 2020-07-13 at 4 58 56 PM" src="https://user-images.githubusercontent.com/6063380/87353169-4bd88200-c52a-11ea-9e77-740d1daa4829.png">

It seems like they're missing from the [en-US translations](https://github.com/processing/p5.js-web-editor/blob/develop/translations/locales/en-US/translations.json). 

RE: @oruburos && @andrewn

PR #1510, opened 2020-07-22
Minimal Viable Navigation Menu

oruburos:
- I18Next configuration leaning on default separator and namespace
- use of withTranslation HOC is simplified by avoiding the use of parameters

These two aforementioned changes solve a double fetching that it is happening in the web editor.

But affects the keys that are present at the moment, so another change is present in this PR ( Related to #1492 `)

- Translations keys use a nested naming structure. 
Components like Nav, Preferences, About, KeyboardShortcut and Toast (among others) are encapsulated in a prefix.

* [x] is descriptively named `


catarak:
Yay!!!! @andrewn I'll let you review this one?

client/modules/IDE/components/About.jsx
```diff
@@ -26,15 +26,15 @@ function About(props) {
         </p>  */}
       </div>
       <div className="about__content-column">
-        <h3 className="about__content-column-title">{t('NewP5')}</h3>
+        <h3 className="about__content-column-title">{t('About.About_Modal.NewP5')}</h3>
```
andrewn, line 29: I wonder if the `About_Modal` is needed as there's a bit of duplication. Could it just be `About.Modal.NewP5`?

I'm not sure though, what do you think?

client/modules/IDE/components/About.jsx
```diff
@@ -67,7 +67,7 @@ function About(props) {
             rel="noopener noreferrer"
           >
             <AsteriskIcon className="about__content-column-asterisk" aria-hidden="true" focusable="false" />
-            {t('Reference')}
+            {t('Nav.Help_DropDown.Reference')}
```
andrewn, line 70: Again, I wonder if these could be shorter?

`Nav.Help_DropDown.Reference` vs `Nav.Help.Reference`?

I'm not sure though, what do you think?

client/modules/IDE/components/KeyboardShortcutModal.jsx
```diff
@@ -58,31 +58,31 @@ function KeyboardShortcutModal() {
           <span className="keyboard-shortcut__command">
             {metaKeyName} + S
           </span>
-          <span>{t('Save')}</span>
+          <span>{t('Nav.File.Save')}</span>
```
andrewn, line 61: My feeling is that this should be prefixed `KeyboardShortcuts` like everything else? What was your thinking behind this?

client/modules/IDE/components/Preferences/index.jsx
```diff
@@ -117,7 +117,7 @@ class Preferences extends React.Component {
                   value="light"
```
andrewn, line 117: The `aria-label`s need to be translated in these files too as they're used by screen readers.

andrewn:
Also, can you remove the commented out code [in the About box](https://github.com/processing/p5.js-web-editor/pull/1510/files#diff-6cf56092f024968eef1416838ec53a9fR17-R26). I checked with @catarak and she said it's ok.

I think it's good not to have untranslated stuff as you convert everything!

Thanks!

client/components/Nav.jsx
```diff
@@ -100,11 +100,11 @@ class Nav extends React.PureComponent {
     const { unsavedChanges, warnIfUnsavedChanges } = this.props;
     if (!unsavedChanges) {
       this.props.showToast(1500);
-      this.props.setToastText('Opened new sketch.');
+      this.props.setToastText('Toast.OpenedNewSketch');
```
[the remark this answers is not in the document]
oruburos: I don’t know, they are used as a part of the Toast, even I'm calling it from Nav. For me is clearer where this translation will be showed. What do you think?

client/modules/IDE/components/About.jsx
oruburos: Yes, looks better and less letters = less chances for typos

client/modules/IDE/components/About.jsx
oruburos: In this case the issue is that when I retrieve Nav.Help will get me an object and i18n will complain

I can switch to this convention, with Title as the label that will go in the DropDown 

"Help": {
  "Title": "Help",
  "KeyboardShortcuts": "Keyboard Shortcuts",
  "Reference": "Reference",
  "About": "About"
},
Notice that this convention will be required in the other dropdowns in Nav menu
what do you think?

client/modules/IDE/components/KeyboardShortcutModal.jsx
oruburos: I have two ideas at the same time:
First, in this way there is only one ‘Save’ mapping across the translation file, so every occurrence of the Save text would be consistent.
If we introduce ‘KeyboardShortcuts.Save’ now we have to check that both translations are consistent.

But also, I suppose that there are parts of the editor that will need to split the same original text  in N different translations, but I don’t know if this entry is one of these cases

client/components/Nav.jsx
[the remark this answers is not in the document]
andrewn: Ok, that makes sense to me. Let’s leave it.

client/modules/IDE/components/About.jsx
andrewn: Yes, that’s much clearer I think. Let’s make that change everywhere if possible.

client/modules/IDE/components/KeyboardShortcutModal.jsx
andrewn: Is this a good candidate for a top-level prefix called “Common”?

So this could be “Common.Save” and we replace it in both “Nav” and “KeyboardShortcuts”

oruburos:
Hi @andrewn  , I'm updating the PR.

andrewn:
`StopSketch` doesn't look translated in English...?

Screenshot 2020-07-30 at 15 13 21

client/modules/IDE/components/About.jsx
```diff
@@ -10,31 +10,21 @@ function About(props) {
   return (
     <div className="about__content">
       <Helmet>
-        <title>p5.js Web Editor | About </title>
+        <title>p5.js Web Editor | {t('About.TitleHelmet')} </title>
```
andrewn, line 13: In the browser's title bar this shows:

> p5.js Web Editor | p5.js Web Editor | About

andrewn:
I think this key needs to be updated...

Screenshot 2020-07-30 at 15 19 05

andrewn:
@oruburos A couple of small things to fix then I think we're good to go!

catarak:
🎉 🎉 🎉

PR #1518, opened 2020-07-27
Bugfix/issue#1512

### LINKED ISSUE #1512: The search list of sketches & collections do not reset back to original state (does not show all the sketches & collections) after search field becomes empty after a search.
#### Nature of issue?

- Found a bug

#### Details about the bug: 

- Web browser and version: On Chrome 84.0.4147.89 (Official Build) (64-bit)
- Operating System: Windows 10 OS Version 1909 (Build 18363.959)
- The search list of sketches & collections do not reset back to original state (does not show all the sketches & collections) after the search field becomes empty after a search. They reset to original state only upon clicking the clear button. To reproduce this bug see the GIF below:

bug

I would like to fix this, if this indeed is a bug.

SundeepChand:
Fixes #1512 

* [x] is descriptively named and links to an issue number, i.e. `Fixes #1512 `

### DIFF
--- a/client/modules/IDE/components/Searchbar/Searchbar.jsx
+++ b/client/modules/IDE/components/Searchbar/Searchbar.jsx
@@ -35,6 +35,11 @@ class Searchbar extends React.Component {
   };
 
   handleSearchChange = (e) => {
+    if (e.target.value === '') {
+      // Manually reset the searchValue if the search field becomes empty.
+      this.handleResetSearch();
+      return;
+    }
     this.setState({ searchValue: e.target.value }, () => {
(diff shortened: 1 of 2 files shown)


SundeepChand:
@catarak Yeah, this approach also works. I had also thought of doing the same but was a bit apprehensive about removing some pre-written code.

SundeepChand:
On examining a little further, I found the function **handleSearchEnter()** to be redundant.  So I made this commit.

catarak:
@SundeepChand makes sense! Originally when I had worked on this with @raclim we left it in because it didn't seem harmful, but maybe it's good to remove it for the sake of future confusion.

PR #1527, opened 2020-07-30
[#1526] Add ellipsis to sidebar file names

### LINKED ISSUE #1526: File names overflow wrap in Sidebar
#### Nature of issue?

- Found a bug

#### Details about the bug: 

- Web browser and version:  Chrome
- Operating System:  Mac OSX
Screen Shot 2020-07-30 at 4 31 10 PM

If you create a file with a name longer than the width of the file Sidebar, it overflows in a way that doesn't look good. The file name should be cut off with an ellipsis.

catarak:
Fixes #1526

I went a little overboard and decided to copy Glitch's design on this one—would love some feedback!
Screen Shot 2020-07-30 at 7 02 23 PM


client/modules/IDE/components/FileNode.jsx
```diff
@@ -10,6 +10,47 @@ import FolderRightIcon from '../../../images/triangle-arrow-right.svg';
 import FolderDownIcon from '../../../images/triangle-arrow-down.svg';
 import FileIcon from '../../../images/file.svg';
 
+function FileName({ name }) {
+  const nameArray = name.split('.');
```
ghalestrilo, line 14: It's a good idea to catch `if (!name) return;` at the beginning of the function, otherwise `.split` can explode

client/modules/IDE/components/FileNode.jsx
```diff
@@ -10,6 +10,47 @@ import FolderRightIcon from '../../../images/triangle-arrow-right.svg';
...
+  if (nameArray.length > 1) {
+    const extension = `.${nameArray[nameArray.length - 1]}`;
+    const fileName = nameArray.slice(0, -1).join('');
+    const firstLetter = fileName[0];
+    const lastLetter = fileName[fileName.length - 1];
```
ghalestrilo, line 19: ditto

client/modules/IDE/components/FileNode.jsx
```diff
@@ -10,6 +10,47 @@ import FolderRightIcon from '../../../images/triangle-arrow-right.svg';
 import FolderDownIcon from '../../../images/triangle-arrow-down.svg';
 import FileIcon from '../../../images/file.svg';
 
+function FileName({ name }) {
+  const nameArray = name.split('.');
+  if (nameArray.length > 1) {
+    const extension = `.${nameArray[nameArray.length - 1]}`;
```
ghalestrilo, line 16: You can use nameArray.slice(-1) here, if that's more readable

client/modules/IDE/components/FileNode.jsx
```diff
@@ -10,6 +10,47 @@ import FolderRightIcon from '../../../images/triangle-arrow-right.svg';
...
+        <span>{firstLetter}</span>
+        {fileName.length > 2 &&
+          <span className="sidebar__file-item-name--ellipsis">{middleText}</span>
+        }
+        {fileName.length > 1 &&
+          <span>{lastLetter}</span>
+        }
+        <span>{extension}</span>
+      </span>
+    );
```
ghalestrilo, line 32: If possible, I think `FileName` should be just this block, and the name logic should go on another function, which returns a string. This will avoid having to duplicate this code below.

client/modules/IDE/components/FileNode.jsx
```diff
@@ -10,6 +10,47 @@ import FolderRightIcon from '../../../images/triangle-arrow-right.svg';
...
+  }
+  const firstLetter = name[0];
+  const lastLetter = name[name.length - 1];
```
ghalestrilo, line 35: ditto

client/modules/IDE/components/FileNode.jsx
```diff
@@ -10,6 +10,47 @@ import FolderRightIcon from '../../../images/triangle-arrow-right.svg';
...
+    <span className="sidebar__file-item-name-text">
+      <span>{firstLetter}</span>
+      {name.length > 2 &&
+        <span className="sidebar__file-item-name--ellipsis">{middleText}</span>
+      }
+      {name.length > 1 &&
+        <span>{lastLetter}</span>
+      }
+    </span>
+  );
```
ghalestrilo, line 47: ditto

client/modules/IDE/components/FileNode.jsx
catarak: `name` is a required string proptype, which means that the component itself will complain if it's used without this prop, so is this necessary?

client/modules/IDE/components/FileNode.jsx
catarak: I always forget that I can use `slice` for this 😄

catarak: I think I'm actually going to change this to `nameArray.charAt(nameArray.length - 1)` as I think that's the most readable.

catarak: Wow, so apparently `charAt` can work in some really wacky ways (see this [StackOverflow answer](https://stackoverflow.com/a/30742383)), so......... I'm just going to stick with bracket notation.

client/modules/IDE/components/FileNode.jsx
catarak: Makes sense!

client/modules/IDE/components/FileNode.jsx
ghalestrilo: True, in this case it's kind of pointless

client/modules/IDE/components/FileNode.jsx
ghalestrilo: Make sure to check/test what happens if you pass an empty string to it!
Javascript doesn't like `array[-1]`

ghalestrilo: Nevermind, you're catching this on line 15. My bad!

catarak:
Other things I still need to consider for this:
- How does this sound with a screen reader? Do I need to add an `aria-label`?
- How does this for for keyboard-only users? Is there a way to expand the text?

ghalestrilo:
Maybe, if there's already a concrete improvement delivered by this PR, you can address these concerns later in another issue/PR.

catarak:
>     * How does this sound with a screen reader? Do I need to add an `aria-label`?
I'm pretty sure this is broken by this PR, so I want to fix it.

>     * How does this for for keyboard-only users? Is there a way to expand the text?
This, however, is something I think could be addressed in a different PR!

catarak:
Tested this with VoiceOver, file labels are working 😄

PR #1513, opened 2020-07-24
Add Dropdown Menu to the mobile IDE View

ghalestrilo:
Part of the Mobile UI Project. This PR adds an `<OverlayManager />` component, and a general-purpose `<Dropdown />` component. It creates a navigation dialog for the `<MobileIDEView />`.

I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] is from a uniquely-named feature branch and has been rebased on top of the latest master. (If I was asked to make more changes, I have made sure to rebase onto master then too)
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #1528, opened 2020-07-31
Feature/mobile examples

ghalestrilo:
Requires #1513 

Creates mobile **My Stuff** / **Examples** views, through the `<MobileDashboardView />`


ghalestrilo:
This PR implements a bug where the navigation from `<IDEView />` to **other views** stops working.

- Popups and modals work fine.
- Simulating clicks from the console also do not work (aka `$("a[href="/login"]).click()`)
- It doesn't seem to be related to the custom `useAsModal` / `useModalBehavior` hooks, because:
  1. It doesn't seem to happen on `feature/mobile-header-dropdown-menu`, which implements them
  2. Commenting / dummying them out does not fix the bug
- It seems to involve `routes.jsx` / `server.routes.js`, no idea how, though.

server/routes/server.routes.js
```diff
@@ -156,4 +144,55 @@ router.get('/:username/collections', (req, res) => {
   ));
 });
 
+
+// Mobile Routes
+if (process.env.MOBILE_ENABLED) {
+  router.get('/mobile', (req, res) => res.send(renderIndex()));
```
andrewn, line 150: If these routes are the same as the desktop routes above, one way to avoid duplication would be to mount them again under the `/mobile` namespace in server.js:

```
// this is supposed to be TEMPORARY -- until i figure out
// isomorphic rendering
app.use('/', serverRoutes);

if (process.env.MOBILE_ENABLED) {
  app.use('/mobile’, serverRoutes);
}
```

ghalestrilo: Wow this is amazing, I didn't know this was possible, thanks andrew!

catarak:
IMG_2954
This is how it looks right now on my phone. Some thoughts:
- The "Sketch" heading and sketch titles don't align. I think the sketch heading needs less padding.
- The table itself is overflowing off the right side of the screen
- I think the table text overall could be smaller, the same font size as the headings
- If it still feels cramped, you could possibly put the dates on a different line from the sketch title (see CodePen dashboard for reference:
IMG_2955
- It could make sense to remove the times from the "created" and "updated" dates, for more room.

ghalestrilo:
The navigation bug has been fixed. The navigation was being prevented because `#warnIfUnsavedChanges` (IDEView.jsx:54) was returning `false` by default. It has been updated to return `true`

catarak:
image

I was just playing around with the styling a bit and was thinking something more like this for the columns? What I did here was:
1. Made the table 
```css
table {
  table-layout: fixed;
  white-space: nowrap;
}
```
2. Changed all of the font sizes in the table to `1rem`
3. Removed the times from the table
4. Made the width of the first column `65%`, the second one `23%`, then centered the button in the third column using `text-align: center`. 
5. I think to the first column we could also add `text-overflow: ellipsis`.

catarak:
It may even make sense to remove the third column on mobile since you can't really download a sketch on mobile?

ghalestrilo:
I'll try the changes you suggested! Also, I've actually been able to download the sketch. Not sure what you'd do with it, but it's possible

ghalestrilo:
I made some adjustments, and despite the table header looking comical, the table is looking **okay** on resolutions >= 320px, both on the Sketch and Collections panes.

Screenshot from 2020-08-04 17-05-25
Screenshot from 2020-08-04 17-05-35

I want to make some visual improvements in the future, like moving that "New" button. Maybe creating a `<Fab />`, or similar component.

ghalestrilo:
OK now I really think it's done
Sorry for commiting after requesting the review, but I found some stuff to improve which would better be fixed now

catarak:
> OK now I really think it's done
> Sorry for commiting after requesting the review, but I found some stuff to improve which would better be fixed now

No worries! I will review this today 🌈

catarak:
This is so great! I think some improvements could be made to the UI, as you have suggested:
- RE: New Sketch Button. One option would be to have an icon at the top with dropdown menu containing common actions from the "My Stuff" view (maybe like, "New Sketch", "New Collection". This is the pattern that CodePen uses:
IMG_2969
Glitch.com just has a "New Project" button, and then the Project list page is pretty ugly:
IMG_2970
IMG_2971
CodeSandbox has an upper navigation that has a "New Sandbox" button, which is nice. We could think about having a Nav bar that stays mostly consistent throughout the app (like with the dropdown icon that appears now, but stuff to the left of the icon could change depending on the page. 
IMG_2972
- Next, the table rows could be improved. The wrapping text is not ideal, and it seems like tables with more than two columns on mobile don't look great. Looking at some other tools:
CodePen has the ability to switch between list/grid view, and rather than putting things into columns, puts the dates below, and separates out the column sort into dropdowns:
IMG_2982
IMG_2983
Glitch does a similar thing (you can see above).
- Lastly, I don't like how the tabs look with the striped table. In your design, it's nice when the tab matches the background color of the table, but it's confusing when sometimes it matches and sometimes it clashes with the table row. One place to start would be the desktop design that uses the pink dotted underline.

ghalestrilo:
I really like the Dropdown approach, I think it can be useful for other features in the future. I'll see what I come up with.
About the table layout, I agree. For mobile, Card Lists >>> Tables in terms of usability.

For now we're using the same component as the default view (to reuse some data fetching logic), and this limits what we can do with it. Is it the case of implementing a new component?

I'll give the tabs a go, too. the pink will probably look a lot better

ghalestrilo:
It took a bit of extra css but now it looks adorable!
also, @andrewn can you give an opinion on the style/theming code structure? (`theme,js` and `MobileDashboardView.jsx` styled component definitions)

image
image

catarak:
Minor design suggestions I have:
- Center the loader horizontally and vertically, rather than aligned at the top as it is now.
Screen Shot 2020-08-11 at 1 32 58 PM
- For me, there's a tiny bit of spacing between the search bar and the navigation bar that I think should be taken away.
- The caret next to the sorting headers should be a few more pixels to the right (add more space between the text and caret)
- Rather than align the "Created" and "Updated" dates with the headers, I think it would look better to space out the sorting headers evenly horizontally, and then for the dates that are in the list items, to prefix the dates with "Created: " and "Updated: "

ghalestrilo:
Nice! I made the changes you suggested:

Screenshot from 2020-08-11 16-58-35
Screenshot from 2020-08-11 16-58-50

the button to the far-left can't be much thinner without practically merging with the left cell (visually), because there isn't enough content, so it looks a bit weird and needlessly right-heavy

catarak:
Awesome 😄 

I was able to get just the header to center all of the items by selecting the `<tr>` and `<td>` just in the `<thead>`:
```css
.<generatedclass> thead tr {
      grid-template-columns: 33% 33% 33% 0fr;
}

.<generated class> thead td {
    display: flex;
    justify-content: center;
}
```
Which looks like this:
Screen Shot 2020-08-11 at 4 38 15 PM

catarak:
Then, I would also remove all of the `padding-left` on the `tbody td` selector, which is causing the dates to not be aligned with the sketch titles.

catarak:
The other changes you made are perfect 🌈

ghalestrilo:
Great! I managed to get the header the way you suggested by fiddling with the grid layout. About the 12px left-padding: I think it's not what caused the misalignment - I think it was the cell's `justify-self`. In fact, removing it moves the text too far to the left 🤦🏻 Right now it looks like this:

Screenshot from 2020-08-11 18-33-31

If the problem persists on your device, can I ask you to take a look into it? I tested a couple different resolutions and browsers here but cannot reproduce

catarak:
This is super close to being done! Just a couple more things. I made a couple of CSS adjustments, and then also:
- The CollectionList is missing a default sort order
- The AssetList items need labels for the size and sketch they belong to. Because there's no sorting/filtering on this list at the moment, maybe we could even remove the headers because they don't make any sense?

ghalestrilo:
I hid the header on the assets list, but I don't know what to do about the collections list: it has 4 fields, and fitting them all on the screen ends up looking crammed. Sorry about not catching that earlier!

Screenshot from 2020-08-13 11-02-42

ghalestrilo:
I made the adjustments you suggested to better fit text on the header. I think this is ready to merge :) 
Give me your thoughts

catarak:
I don't want to leave this PR open any longer so I think this is ready to merge! A couple of small things that I think could go in another PR:
Screen Shot 2020-08-17 at 2 58 33 PM
- In the Asset List, I think the "Size" and "Sketch" could be labelled (i.e. prefixed with "Size: " and "Sketch: ")
- I also noticed that the table rows are alternating colors in this list. Since they have the border/shadows I think maybe this should be removed and match the Sketch List/Collection List. What do you think?

PR #1537, opened 2020-08-07
Aria Labels in Toolbar

oruburos:

This pull request
introduces translations in Aria labels in toolbar component.
It also introduces the i18n test file to deal with the toolbar.test


client/modules/IDE/components/Toolbar.test.jsx
```diff
@@ -1,9 +1,8 @@
 import React from 'react';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import lodash from 'lodash';
-
-
 import { ToolbarComponent } from './Toolbar';
+import { i18ntest } from '../../../i18n-test';
```
andrewn, line 5: If you're not using the `i18ntest` variable, then there's no need to assign it here. You can just do this:

`import '../../../i18n-test';`

client/modules/IDE/components/Toolbar.jsx
```diff
@@ -86,23 +87,23 @@ class Toolbar extends React.Component {
             this.props.setTextOutput(true);
             this.props.setGridOutput(true);
           }}
-          aria-label="Play sketch"
+          aria-label={i18n.t('Toolbar.PlaySketchARIA')}
```
andrewn, line 90: I think we should use `withTranslation()` here. The [react-i18next](https://react.i18next.com/latest/using-with-hooks#translate-your-content) documentation doesn't mention the method you're using here at all which makes me concerned about using it.

All of these can be `this.props.t('...')`

oruburos:
@andrewn I made the changes requested, but the tests are failing in toolbar.test, do you have an idea what could fix this? Can you take a look please?

PR #1568, opened 2020-08-22
Use translations for all tests

andrewn:
This refine's @oruburos' PR #1537 to ensure that all Jest tests using React testing-library can access the `en-US` translations.

This is good because we want to be testing against how user's see the app e.g. "click the button with label Play sketch".

It also uses translated ARIA Labels in the Toolbar (which causes the tests to fail in the first place).

* [x] is descriptively named


andrewn:
@oruburos Can you review this please

oruburos (approved):

PR #1562, opened 2020-08-19
Add prototype of language selector design

catarak:
✨ Requesting feedback ✨ from @andrewn and @oruburos 

Yesterday, @ghalestrilo and I talked about integrating the translations with the mobile design (#1559). In trying to come up a design for how to select languages on mobile, I was messing around with the desktop design to see if that would generate ideas and I came up with this:

Screen Shot 2020-08-19 at 2 58 17 PM
Screen Shot 2020-08-19 at 2 58 12 PM
Screen Shot 2020-08-19 at 3 12 48 PM

I don't want to step on your toes so feel free to ignore this, change this, etc.—do what you think is best! I figured I would just put it out there 😄 

I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] is from a uniquely-named feature branch and has been rebased on top of the latest master. (If I was asked to make more changes, I have made sure to rebase onto master then too)
* [ ] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #1582, opened 2020-08-26
Language UI Dropdown new location

oruburos:
Built on #1562

* [x] is descriptively named and links to an issue number

This PR locates the Language Selector in a new position, and sets a language in defaultprops to pass the test.

### DIFF
--- a/client/components/Nav.jsx
+++ b/client/components/Nav.jsx
@@ -806,7 +806,7 @@ Nav.propTypes = {
   }),
   t: PropTypes.func.isRequired,
   setLanguage: PropTypes.func.isRequired,
-  language: PropTypes.string,
+  language: PropTypes.string.isRequired,
 };
 
 Nav.defaultProps = {
@@ -819,8 +819,7 @@ Nav.defaultProps = {
   warnIfUnsavedChanges: undefined,
   params: {
     username: undefined
-  },
-  language: 'en-US'
+  }
 };
 
 function mapStateToProps(state) {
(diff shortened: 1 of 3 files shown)


oruburos:
@andrewn Can you have a look at this, please?

client/components/Nav.jsx
```diff
@@ -820,7 +819,8 @@ Nav.defaultProps = {
   warnIfUnsavedChanges: undefined,
   params: {
     username: undefined
-  }
+  },
+  language: 'en-US'
```
andrewn: I don't think we should hard-code `en-US` here as it introduces duplication throughout the codebase. Let's make language required as the Redux store will be initialised with a default language.

andrewn (changes requested):
This is looking really good. Can you remove the `en-US` default prop and we're good to go!

client/components/Nav.jsx
oruburos: Ok, I did it to remove a warning that is received when you run
 ``` npm run test ```
 with result : 
``` Warning: Failed prop type: The prop `language` is marked as required in `Nav`, but its value is `undefined`.``` 
but the tests pass anyway

oruburos:
@andrewn I got ride of the language entry in default prop, can you have a look, please?

client/components/Nav.jsx
andrewn: If there's a warning in the tests then that needs to be fixed.

It's happening because `language` is required and it's not being being supplied. Try adding a default language (`en-US`) to [the `props` object](https://github.com/processing/p5.js-web-editor/blob/develop/client/components/__test__/Nav.test.jsx#L48-L50).

oruburos: Correct, your suggestion worked. Thanks.

oruburos:
@andrewn can you have a look at this?

andrewn (approved):

PR #1597, opened 2020-09-03
:sparkles: add toast to MobileIDEView

### LINKED ISSUE #1596: Save Sketch Toast on Mobile
#### Nature of issue?

Currently, the mobile editor can save, but it's missing the toast notification. This is because the `<Toast />` component is missing on `<MobileIDEView />`

ghalestrilo:
Fixes #1596

### DIFF
--- a/client/components/mobile/MobileScreen.jsx
+++ b/client/components/mobile/MobileScreen.jsx
@@ -1,19 +1,33 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { remSize } from '../../theme';
 
-const Screen = ({ children, fullscreen }) => (
-  <div className={fullscreen && 'fullscreen-preview'}>
+const ScreenWrapper = styled.div`
+  .toast {
+    margin-top: ${props => remSize(props.slimheader ? 49 : 68)};
+    font-size: ${remSize(12)};
+    z-index: 0.5;
+    width: 100%;
+    padding: ${remSize(8)};
+  }
+`;
+
+const Screen = ({ children, fullscreen, slimheader }) => (
+  <ScreenWrapper className={fullscreen && 'fullscreen-preview'} slimheader={slimheader}>
     {children}
-  </div>
+  </ScreenWrapper>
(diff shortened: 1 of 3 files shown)


catarak:
Looking great 😄 What would you think about placing the toast under the header? I don't know how I feel about it blocking the title. Maybe the text could also be smaller on mobile?

ghalestrilo:
Sure! I did it, and it does look better, but I wanna hear your thoughts on this:

1. It still covers the first two lines of code
2. Mobile apps usually have a pop-up banner above the header, to indicate lack of connectivity, or similar infrequent info. I was thinking of this as an alternative?

catarak:
> Sure! I did it, and it does look better.

I agree it does look better! Also agree that it is annoying that it covers the first two lines of code.

> 2. Mobile apps usually have a pop-up banner above the header, to indicate lack of connectivity, or similar infrequent info. I was thinking of this as an alternative?

Do you have some examples we could use as a visual reference?

ghalestrilo:
I thought of doing something like [this](https://css-tricks.com/pop-from-top-notification/), except pushing everything down a bit. In hindsight, not a brilliant idea :^)
I've seen a couple of apps placing the toast to the bottom, it's an alternative. Tell me which option feels best, and we'll go from there
image

edit.: changed the example - it's best not to hide the bottom bar with the toast

catarak:
I like how it looks at the bottom! It feels less intrusive, I like that it's closer to reach with your thumb so that you can get rid of it if you want to, but it also doesn't feel like it's in a location that it won't be noticed.

ghalestrilo:
I think it looks nicer now, check it out!

catarak:
For me, this is what the toast looks like:
Screen Shot 2020-09-11 at 6 06 19 PM

ghalestrilo:
lord, I forgot a css rule. definitely not a design choice!

catarak (approved):
Made one last change so that the toast doesn't sit right up against the edges, which looked strange with the border radius. Now this is ready to be merged in!

PR #1601, opened 2020-09-04
Mobile Floating Sketch Preview

ghalestrilo:
Adds a floating sketch preview to the bottom left of the `<MobileIDEView />`, which renders the project (autorefreshing). It's triggered by pressing and holding the `Play` header button. This required a new hook (`useLongPress`).


catarak:
I'm so excited you're working on this! Some thoughts I have looking at this now (which obviously you can address in your own time, or we can discuss, etc.):
- I think the floating canvas maybe needs a border or drop shadow or something to show it's floating
- How do you let users know that the play button has two actions, a short press and a long press, and that there is a difference between them?
- When the floating canvas is visible, should the play button turn into a stop button? What if the user wants to reload the sketch? Should it be like desktop in which a play and stop button are both visible?
- I have a feeling you're on this already, but being able drag the floating canvas around feels crucial. 
- I think it would be interesting to play around with some different canvas sizes! Also, how to handle different canvas sizes, especially ones with different aspect ratios?

ghalestrilo:
Thanks a lot for the feedback!

- Yes! it's on the map :) also, I want to slightly round its corners and maybe add an 1px outline.
- Absolutely no idea right now
- Great point. Right now, it defaults to autorefreshing, for compactness (don't know where to place this information on screen without feeling clunky)
- Forgot about it! Will try it out!
- Setting a max height-width (and resizing the sketch to fit seems) like the simplest path (basically, `object-fit: contain` behavior). Open to suggestions!

ghalestrilo:
Thinking of using [this](https://stackblitz.com/edit/usedraggable?file=DraggableBox.js) for the draggable behavior of the sketch preview

catarak:
> Thinking of using [this](https://stackblitz.com/edit/usedraggable?file=DraggableBox.js) for the draggable behavior of the sketch preview

seems to work well and seems really simple/not overly complicated!

ghalestrilo:
Hi! Now is a good time to get feedback on this feature :)

Good:

- Press+Hold triggers `<FloatingPreview />`
- You can drag it around
- `<FloatingPreview />` has a nice shadow
- Press Stop to hide/stop it.
- Works on mobile (touch events)
- Adjusts sketch dimensions to fit preview

Bad:

- Size adjustment is a bit inconsistent (doesn't support height > width for instance). In the future, it's worth investigating the `object-fit` css prop for a more robust solution (hopefully)
- Dragging supports *either* touch or mouse events atm, not both simultaneously.
- There is a bug on mobile preventing the sketch from rendering. It's on develop rn, so this branch is not the cause. On desktop it seems to work fine, though, but a fix will be necessary to use this properly

I don't want this branch to get too large / outdated, so maybe these issues can be addressed in new ones. Tell me what you think makes sense! @catarak

ghalestrilo:
The bug seems related to p5.js sound, I think it's still the same one we've been seeing, but no idea how to fix it

Screenshot from 2020-10-12 11-14-35

catarak:
> The bug seems related to p5.js sound, I think it's still the same one we've been seeing, but no idea how to fix it
> 
> Screenshot from 2020-10-12 11-14-35

If you're able to use https when running the project locally, you can fix this error. What I do is use [ngrok](https://ngrok.com/), which gives you an https url.

ghalestrilo:
Running via ngrok solved it! Auto-resizing is still an issue, though. This is what it looks like:

Screenshot_2020-10-14-09-38-43-312_com brave browser_1
Screenshot_2020-10-14-09-41-36-418_com brave browser_1
Screenshot_2020-10-14-09-41-47-073_com brave browser_1

raclim:

PR #1624, opened 2020-09-29
Revert "Make console interactive"

catarak:
Reverts processing/p5.js-web-editor#667


catarak:
turns out #667 is not working in production :(

PR #1636, opened 2020-10-16
Prevent browser default unload behavior

### LINKED ISSUE #1633: Unsaved changes handler is excessive
#### Nature of issue?
- Found a bug

#### Details about the bug: 
- Web browser and version: Google Chrome 86.0.4240.75
- Operating System: Windows

Open a sketch, press Play, then close the tab.

Even though I made no changes, the annoying `beforeunload` handler prevents me from closing the tab.

Some ideas:

a) Remove the handler entirely, it's more frustrating than it helps.
b) Automatically save drafts to prevent data loss. LocalStorage is a great place for that.
c) Make installation of `beforeunload` handler configurable.
d) Properly detect actual changes and don't prevent user's from closing a tab just because they played the sketch.

oliversalzburg:
When returning a non-null and non-undefined value from an onbeforeunload handler, causes the browser to show a generic warning dialog.

Before this change, the handle would determine that there are no unsaved changes if there are none, but it would still return a value that signals the browser to show the default dialog.

This change achieves that no dialog is shown if there are no unsaved changes or if we're in a view state where this behavior is undesirable.

Fixes #1633

### DIFF
--- a/client/modules/IDE/pages/IDEView.jsx
+++ b/client/modules/IDE/pages/IDEView.jsx
@@ -66,7 +66,6 @@ function warnIfUnsavedChanges(props) {
       return false;
     }
     props.setUnsavedChanges(false);
-    return true;
   }
   return true;
 }
@@ -247,7 +246,13 @@ class IDEView extends React.Component {
     }
   }
 
-  handleUnsavedChanges = () => warnIfUnsavedChanges(this.props);
+  handleUnsavedChanges = (event) => {
+    if (!warnIfUnsavedChanges(this.props)) {
+      // See https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload
+      event.preventDefault();
+      event.returnValue = undefined;
+    }
+  }
 
   render() {
     return (


oliversalzburg:
It should be good now, based on my current understanding of the subject.

catarak:
I think this is too aggressive—now if there are unsaved changes, you can navigate away from the editor.

oliversalzburg:
@catarak Hmm, I interpreted the code to invoke `window.prompt` if there are unsaved changes and that that would halt the unloading until the dialog is confirmed.

I believe I understand what you mean though. I will try to get this running locally to properly verify.

oliversalzburg:
Okay, the `window.confirm` doesn't actually work at all. It's not valid in an `onbeforeunload` handler:

> Blocked confirm('Are you sure you want to leave this page? You have unsaved changes.') during beforeunload.

oliversalzburg:
Browser behavior is definitely more varied than I initially assumed. And the `warnIfUnsavedChanges` function is used in other scenarios than just the unload handler.

I now believe the correct approach is to handle the event according to the spec and leave the original function as-is.

oliversalzburg:
Some note on why this wasn't good enough or that a different PR is replacing it would have been nice. Seems kinda rude.

Congrats that you got a solution though.

catarak:
@oliversalzburg it was not my intention to be rude, I interpreted your last comment as not knowing how to proceed on this PR, which is why I made a new one. Opening this PR was really helpful and helped me figure out the solution!

oliversalzburg:
Okay. I still feel like this could have been communicated better. I had pushed a new commit with a different approach, that my comment was in reference to. Sorry if it was misleading.

I'm glad a solution for everyone's benefit was achieved in the end.

PR #1683, opened 2020-11-17
Fix Duplicate Sketch button

### LINKED ISSUE #1677: "Duplicate Sketch" button broken
#### Nature of issue?

- Found a bug

#### Details about the bug: 

Using "duplicate sketch" from the dropdown in "my sketches" doesn't duplicate the sketch; instead an uncaught error in promise appears. (Maybe I'm doing something wrong?)

Screen Shot 2020-11-14 at 11 55 10 PM

- Web browser and version:  Tested on Chrome 86.0.4240.198 (Official Build) (x86_64) and Firefox 82.0.3 (64-bit)
- Operating System:  macOS Catalina 10.15.7
- Steps to reproduce this bug: go to "my sketches" and try to duplicate a sketch

Screen Shot 2020-11-14 at 11 56 13 PM

Not sure this matters, but this is the sketch I'm trying to duplicate: https://editor.p5js.org/bandaloo/sketches/zXMOZkuYU

Swapnil-2001:
Fixes #1677 

@catarak kindly let me know if the changes are alright.

### DIFF
--- a/client/modules/IDE/actions/project.js
+++ b/client/modules/IDE/actions/project.js
@@ -265,12 +265,12 @@ export function cloneProject(id) {
       if (!id) {
         resolve(getState());
       } else {
-        apiClient.get(`/projects/${id}`)
-          .then(res => res.json())
-          .then(data => resolve({
-            files: data.files,
+        const currentUser = getState().user.username;
+        apiClient.get(`/${currentUser}/projects/${id}`)
+          .then(res => resolve({
+            files: res.data.files,
             project: {
-              name: data.name
+              name: res.data.name
             }
           }));
       }
--- a/client/modules/IDE/actions/project.js
+++ b/client/modules/IDE/actions/project.js
@@ -265,8 +265,7 @@ export function cloneProject(id) {
       if (!id) {
         resolve(getState());
       } else {
-        const currentUser = getState().user.username;
-        apiClient.get(`/${currentUser}/projects/${id}`)
+        apiClient.get(`/projects/${id}`)
           .then(res => resolve({
             files: res.data.files,
             project: {
--- a/server/controllers/project.controller.js
+++ b/server/controllers/project.controller.js
@@ -81,6 +81,17 @@ export function getProject(req, res) {
   });
 }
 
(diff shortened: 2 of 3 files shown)


client/modules/IDE/actions/project.js
```diff
@@ -265,12 +265,12 @@ export function cloneProject(id) {
       if (!id) {
         resolve(getState());
       } else {
-        apiClient.get(`/projects/${id}`)
-          .then(res => res.json())
-          .then(data => resolve({
-            files: data.files,
+        const currentUser = getState().user.username;
```
catarak: I don't think this is the right solution—what if a user is duplicating a sketch that doesn't belong to them?

catarak (changes requested):

client/modules/IDE/actions/project.js
Swapnil-2001: I see...yes I hadn't thought of that, sorry!

I went through the `GET` routes specified on the server side and the appropriate one appeared to be `router.get('/:username/projects/:project_id', ...)`, so I thought of implementing it this way.

catarak: Sketches/projects (you can use the terms interchangeably) are unique by id so you shouldn't need to include the username. I don't think this is the reason it's failing—I would look at how it's different when calling `cloneProject` from `SketchList.jsx` versus `Nav.jsx`.

Swapnil-2001: I agree the route should remain `'/projects/:project_id'`, however, I couldn't find an implementation that would allow fetching of projects only by their ID. So I took the liberty of adding another controller and `GET` route for the same. 

Could you please take a look?

catarak:
Thanks for working on this! In reviewing your code, I realized it didn't make any sense to make the extra API call when frontend already has the sketch, so I made a new PR: https://github.com/processing/p5.js-web-editor/pull/1686. Want to take a look at that one?

Swapnil-2001:
> I realized it didn't make any sense to make the extra API call when frontend already has the sketch

Ah of course. The new PR looks good!

PR #1694, opened 2020-11-25
Refactor examples.js to use async/await

### LINKED ISSUE #1693: Refactor `fetch-examples` script to only use async/await
#### Nature of issue?

The file [`examples.js`](https://github.com/processing/p5.js-web-editor/blob/develop/server/scripts/examples.js) has been edited by many people over the years and could use a refactor. It uses callbacks, Promises, and async/await, and I think that it should be refactored to just use async/await.

Swapnil-2001:
Fixes #1693 

@catarak please let me know if the changes are okay.

### DIFF
--- a/server/scripts/examples.js
+++ b/server/scripts/examples.js
@@ -3,7 +3,6 @@ import Q from 'q';
 import mongoose from 'mongoose';
 import objectID from 'bson-objectid';
 import shortid from 'shortid';
-import eachSeries from 'async/eachSeries';
 import User from '../models/user';
 import Project from '../models/project';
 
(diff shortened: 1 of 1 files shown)


catarak:
Thanks for working on this! Your changes look great. One thing I would also suggest changing is the reliance on the library `async` (and the function `eachSeries`) which is used to make using callback functions easier. I believe you could get rid of it and just use the library `q`!

Swapnil-2001:
Okay, I'll be changing that!

Swapnil-2001:
Could you please go through the changes? This is my first time working with the library `q` so I'm not 100% sure!

server/scripts/examples.js
```diff
@@ -236,125 +234,123 @@ function createProjectsInP5user(projectsInAllCategories) {
...
-            projectCallback();
-          });
-        }, (categoryErr) => {
-          categoryCallback();
+        newProject.save((saveErr, savedProject) => {
+          if (saveErr) throw saveErr;
+          console.log(`Created a new project in p5 user: ${savedProject.name}`);
         });
-      }, (examplesErr) => {
-        process.exit();
```
catarak, line 347: You removed this `process.exit`, so now when you run the `fetch-examples` script, it hangs when it's done and never closes. Could you add this back?

catarak (changes requested):

server/scripts/examples.js
Swapnil-2001: I've added it back. Please let me know if there is anything else that needs to be changed.

catarak:
Looking at this again, I also noticed that the `mongoose` queries (e.g. `User.findOne`) use callbacks—could these use `async/await` as well?

Swapnil-2001:
Of course!

Also, I replaced `remove` with `deleteOne`, since `remove` is deprecated.

PR #1708, opened 2020-12-15
Removes setTimeout for setting cursor's position.

### LINKED ISSUE #1705: Lag in bringing the cursor back to it's original postion after tidy operation.
#### Nature of issue?

Found a bug

#### Details about the bug: 

There is a small lag in moving the cursor to its original position after the tidy operation. This I believe is because we wrap the function in a setTimeout which delays it.

```javascript
setTimeout(() => {
      this._cm.focus();
      this._cm.doc.setCursor({ line: currentPosition.line, ch: currentPosition.ch + INDENTATION_AMOUNT });
}, 0);
```
Why do we need to use setTimeout here as the position to which we're bringing back the cursor is simply the old line number and character number.

- Web browser and version: Chrome Version 87.0.4280.88 
- Operating System:  MacOS Big Sur
- Steps to reproduce this bug: Execute the tidy operation using either the keyboard shortcut or in the menu.

namanarora00:
Fixes #1705

* [x] is descriptively named and links to an issue number, i.e. `Fixes #1705`

I did encounter some warnings on running `npm run lint` but I believe these are being taken care of in another PR. The changes I've made do not create any new warnings.

### DIFF
--- a/client/modules/IDE/components/Editor.jsx
+++ b/client/modules/IDE/components/Editor.jsx
@@ -304,10 +304,8 @@ class Editor extends React.Component {
     } else if (mode === 'htmlmixed') {
       this._cm.doc.setValue(beautifyHTML(this._cm.doc.getValue(), beautifyOptions));
     }
-    setTimeout(() => {
-      this._cm.focus();
-      this._cm.doc.setCursor({ line: currentPosition.line, ch: currentPosition.ch + INDENTATION_AMOUNT });
-    }, 0);
+    this._cm.focus();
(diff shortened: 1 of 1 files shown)


catarak (approved):
This is working great for me! I just noticed that the Sublime keymap is unindenting the code when you use the keyboard shortcut to run tidy code so I removed it too.

namanarora00:
Thanks for dealing with this, @catarak. While this is fine do you think instead of deleting the keymap we could change the keyboard shortcut for tidy itself since Shift+Tab is a commonly used shortcut for unindenting?

catarak:
@namanarora00 yes! I'm gonna open a new issue for that :)

PR #1732, opened 2021-01-08
Add Cmd+Shift+F as tidy code shortcut and prevent browser default behaviour

### LINKED ISSUE #1711: Change keyboard shortcut for tidy code
@namanarora00: "Instead of deleting the keymap for Shift+tab, could we change the keyboard shortcut for tidy itself since Shift+Tab is a commonly used shortcut for unindenting?"

_Originally posted by @namanarora00 in https://github.com/processing/p5.js-web-editor/issues/1708#issuecomment-745754560_

vulongphan:
Fixes #1711


vulongphan:
@catarak what I did was to detect key short cut of <kbd>Cmd</kbd>+<kbd>Shift</kbd>+<kbd>F</kbd> to call `tidyCode()` and to prevent browser default behaviour (Full Screen Mode in Chrome on Mac). Could you please have a look and tell me if I have to change anything?

client/modules/IDE/components/Editor.jsx
```diff
@@ -155,14 +160,21 @@ class Editor extends React.Component {
...
+    this._cm.on('keyup', (_cm, e) => {
       const temp = this.props.t('Editor.KeyUpLineNumber', { lineNumber: parseInt((this._cm.getCursor().line) + 1, 10) });
       document.getElementById('current-line').innerHTML = temp;
+      this.handleKey(this.map, e);
     });
 
     this._cm.on('keydown', (_cm, e) => {
-      // 9 === Tab
-      if (e.keyCode === 9 && e.shiftKey) {
+      this.handleKey(this.map, e);
```
catarak: I'm not sure how using this function `this.handleKey` is helping in this scenario? Just from reading this it seems like it's adding some intermediate code that is unnecessary, but please explain if I'm wrong!

client/modules/IDE/components/Editor.jsx
```diff
@@ -155,14 +160,21 @@ class Editor extends React.Component {
...
+      // 91 === Cmd
```
catarak: There's a variable that's imported to this file called `metaKey` which will be set to `'Cmd'` if the user is on Mac, and `'Ctrl'` if the user is on Windows/Linux. I think you should use this variable rather than just `'Cmd'` since `'Cmd'` doesn't exist on Windows/Linux!

client/modules/IDE/components/Editor.jsx
```diff
@@ -62,7 +62,11 @@ const beautifyHTML = beautifyJS.html;
 window.JSHINT = JSHINT;
 window.CSSLint = CSSLint;
 window.HTMLHint = HTMLHint;
-delete CodeMirror.keyMap.sublime['Shift-Tab'];
+
+
+// delete CodeMirror.keyMap.sublime['Shift-Cmd-F'];
+
+console.log(CodeMirror.keyMap.sublime);
```
catarak: Could you delete this `console.log` and the commented out code?

catarak (changes requested):

client/modules/IDE/components/Editor.jsx
vulongphan: The reason why I use this function is because to detect shortcut <kbd>Cmd</kbd>+<kbd>Shift</kbd>+<kbd>F</kbd> the following conditional statement would not work
```
if (e.keyCode === 91 && e.keyCode === 16 && e.keyCode === 70){ }
```
I guess the reason is because the browser cannot detect multiple key press. To solve this, I used an object called `this.map` to store the `keydown` state of the keyboard. For example, if the shortcut <kbd>Cmd</kbd>+<kbd>Shift</kbd>+<kbd>F</kbd> is pressed, `this.map` will be
```
{91: true, 16: true, 70: true}
```
 and when we release the shortcut, `this.map` will be
```
{91: false, 16: false, 70: false}
```
I wonder if there is a better way to get around this.

client/modules/IDE/components/Editor.jsx
vulongphan: Yes, I will incorporate the use of `metaKey` in my changes!

client/modules/IDE/components/Editor.jsx
vulongphan: Yes!

vulongphan:
I have made the changes accordingly, could you please have a look?

Thank you!

client/modules/IDE/components/Editor.jsx
catarak: there is! if you look at the documentation for a [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent), you'll see that it contains `metaKey`, which will be true if <kbd>Command</kbd> is pressed, `ctrlKey`, which will be true if <kbd>Control</kbd> is pressed, and `shiftKey`, which will be true if <kbd>Shift</kbd> is pressed.

vulongphan:
@catarak I am sorry but does this mean that this PR should be closed now? Please correct me if I missed something here

catarak:
@vulongphan I just added a comment about using `KeyboardEvent.metaKey`, `KeyboardEvent.shiftKey`, and `KeyboardEvent.ctrlKey`—could you use these rather than `handleKey`?

vulongphan:
@catarak It was my bad I did not see your comment earlier on. Also please ignore the first two commits ([e53f956](https://github.com/processing/p5.js-web-editor/pull/1732/commits/e53f9561ec43d76d2b65a0c1c0cd77cc030c0142) and [0be0c85](https://github.com/processing/p5.js-web-editor/pull/1732/commits/0be0c854501869bfa1660fc66ba1ed4f1c35dd41)) and check [059b171](https://github.com/processing/p5.js-web-editor/pull/1732/commits/059b1710c74058bf5801df4d778cc9215bd5a87c).
I am really sorry for the inconvenience.

catarak:
> @catarak It was my bad I did not see your comment earlier on. Also please ignore the first two commits ([e53f956](https://github.com/processing/p5.js-web-editor/pull/1732/commits/e53f9561ec43d76d2b65a0c1c0cd77cc030c0142) and [0be0c85](https://github.com/processing/p5.js-web-editor/pull/1732/commits/0be0c854501869bfa1660fc66ba1ed4f1c35dd41)) and check [059b171](https://github.com/processing/p5.js-web-editor/pull/1732/commits/059b1710c74058bf5801df4d778cc9215bd5a87c).
> I am really sorry for the inconvenience.

No need to apologize!!! Contributing is about learning rather than being perfect 😄

client/modules/IDE/components/Editor.jsx
```diff
@@ -168,8 +167,9 @@ class Editor extends React.Component {
     });
 
     this._cm.on('keydown', (_cm, e) => {
-      // 9 === Tab
-      if (e.keyCode === 9 && e.shiftKey) {
+      // 70 === f
+      if ((e.metaKey || e.ctrlKey) && e.shiftKey && e.keyCode === 70) {
```
catarak: i think it would also make sense to check the value of [`metaKey`](https://github.com/processing/p5.js-web-editor/blob/059b1710c74058bf5801df4d778cc9215bd5a87c/client/modules/IDE/components/Editor.jsx#L39), so that Ctrl + Shift + F doesn't work on Mac.

catarak (changes requested):

client/modules/IDE/components/Editor.jsx
vulongphan: Yes, I have added the check accordingly!

catarak:
Working great for me! Would you mind updating the [KeyboardShortcutModal](https://github.com/processing/p5.js-web-editor/blob/develop/client/modules/IDE/components/KeyboardShortcutModal.jsx) component with the new shortcut? Thank you!

vulongphan:
Hi @catarak, I have updated the file but for some reason I cannot see the change in my local development. Could you please check for me? Attached is the local development after I updated the file
Screen Shot 2021-02-18 at 5 13 10 PM

catarak:
I'm seeing the same thing! I think the issue is that you need to update the file [Nav.jsx](https://github.com/processing/p5.js-web-editor/blob/develop/client/components/Nav.jsx#L425) as well.

vulongphan:
@catarak yes I really do agree that we have to update `Nav.jsx` as well. I have updated the file and now see the changes in my local development.
However, `npm run test` has failed which does not seem to be caused by the changes I have made. Could you please give me some suggestion?

Thank you!

catarak:
Because you're changing the appearance of `Nav.jsx`, you need to update the snapshot as well: `npm run test -- --updateSnapshot`. I'm going to go ahead and just do this so I can merge this fix in.

catarak (approved):

PR #1788, opened 2021-03-04
Switching to prettier from beautify js

### LINKED ISSUE #1223: Keep cursor position when tidying code
#### Nature of issue?

The tidying code feature resets the cursor back to the beginning of the file that was tidied. For long files, this can be irksome, because you then have to scroll back to your position. It would be helpful to keep the cursor position from before the tidying.

satyasaibhushan:
Fixes #1223 

* [*] has no linting errors (`npm run lint`)
* [*] is from a uniquely-named feature branch and has been rebased on top of the latest `develop` branch. (If I was asked to make more changes, I have made sure to rebase onto `develop` then too)
* [*] is descriptively named and links to an issue number, i.e. `Fixes #123`

I've attached a screenshot of unwanted behaviour in the issue 
Here I'm attaching a desired behaviour one

https://user-images.githubusercontent.com/40578313/110018902-d1c16b80-7d4d-11eb-8851-44efb5e06945.mov


client/modules/IDE/components/Editor.jsx
```diff
@@ -346,25 +346,23 @@ class Editor extends React.Component {
         }
       );
       this._cm.doc.setValue(formatted);
-      return cursorOffset;
+      if (cursorOffset) {
```
satyasaibhushan: Hey @catarak ,
please also add `this._cm.focus()` , before setting the cursor.

client/modules/IDE/components/Editor.jsx
```diff
@@ -346,25 +346,23 @@ class Editor extends React.Component {
...
+        this._cm.doc.setCursor(this._cm.doc.posFromIndex(cursorOffset));
+      }
     } catch (error) {
-      return null;
+      console.error(error);
```
satyasaibhushan, line 351: coming to the error part, I thought of replicating the original functionality. As the syntax error is only shown when we save but not when we tidy the code.

client/modules/IDE/components/Editor.jsx
satyasaibhushan: You can also remove the `if` statement now as the `cursorOffset` is always non zero
(CurorsorOffset is the latest cursor position but not the change in cursor position).
If we want to be specific we can do it by `if(cursorOffset==this._cm.doc.indexFromPos(this._cm.doc.getCursor()))`

satyasaibhushan:
Hey @catarak, 
If you agree with me with above comments, I can add another commit to this

client/modules/IDE/components/Editor.jsx
catarak: In my testing I found this wasn't necessary? Could be Mac only. I'll try it on Windows too.

satyasaibhushan: This is necessary if we chose to use the menu for tidying the code.

catarak: of course, thank you!

client/modules/IDE/components/Editor.jsx
catarak: I think it's okay to show syntax errors when tidying the code! I think it makes more sense, anyway.

catarak:
Didn't notice your comment until now, just pushed the changes you mentioned. Look good to you?

satyasaibhushan:
Perfect!

catarak (approved):

PR #1912, opened 2021-08-05
Add the zh-TW language translation.

codemee:
Add the zh-TW (Traditional Chinese) language translation for P5 Web Editor.

On branch develop
Your branch is up to date with 'origin/develop'.

Changes to be committed:
	modified:   client/components/Nav.jsx
	modified:   client/i18n.js
	modified:   package-lock.json
	new file:   translations/locales/zh-TW/translations.json

Fixes #issue-number

* [ ] is from a uniquely-named feature branch and has been rebased on top of the latest `develop` branch. (If I was asked to make more changes, I have made sure to rebase onto `develop` then too)


catarak:
Hi @codemee! Thank you for working on this translation. Lots of folks are adding new translations, which is super exciting, but I'm also concerned with the maintenance of these translations—for example, if there are UI changes that add new copy, this copy needs to be added across **all** of the translations as well. Since there is already a Simplified Chinese translation, I'm curious about 

1. What are the differences between Simplified and Traditional Chinese, especially as it relates to the web, coding and art?
2. Are there folks that can read Traditional Chinese but not Simplified Chinese?
3. Do you think this translation [increases access](https://github.com/processing/p5.js/blob/main/contributor_docs/access.md) to p5.js and the p5.js Editor?

Thank you for your help! @limzykenneth @Qianqianye it would also be helpful if y'all weighed in 😄

codemee:
Hi

1. The defferrences between Simplified and Traditional Chinese is that Simplified Chinese reduce the number of characters to a small set and replacing some characters with simplified shapes. This forces some character with multiple meanings in the smaller Simplified Chinese character set. It also make some characters look different in Simplified and Traditional Chinese. As a result, people used to read Simplified or Traditional Chinese are not familiar with the other one.
2. Besides, Simplified Chinese is used in mainland China, Traditional Chinese is used mainly in Taiwan, the phrases of the same meaning used in different territories are not always the same. For example, the Simplied Chinese phrase for "defualt" is "默认" but "預設" in Traditional Chinese.  It's not only different in the shapes of characters, but also in the mind thinking.
3. With localized version of P5.js Editor, it would be simpler to intriduce P5.js to people who use Traditional Chinese. Most people fear of learning new things, it would be better for them to focus on learning the new technology without worrying about the language use in the UI of  any tools.

In fact, here in [Flag Tech](https://www.flag.com.tw/maker/pd), we are designing a studying kit for teaching ML5.js using P5.js Editor for interacting with micro-controlled board. So we are look foward to Traditional Chinese version of P5.js Editor.

Thanks.

limzykenneth:
I do very much look forward to having Traditional Chinese translation in p5.js (for the p5.js website when I was working on Simplified Chinese) but just never got around to it, partly because I can't find the time, partly because as @codemee mentioned, terms used in Simplified Chinese and Traditional Chinese are often different so it couldn't be a simple direct transform from the Simplified script to the Traditional script.

I think @codemee's answers are pretty good, I'm just going to expand a bit from my end.

I'm kind of a weird case with my background and I'm able to understand both Simplified and Traditional Chinese so I can't say much about whether other Chinese users are able to do the same as well. As pointed out above, the difference in terms used can become a source of confusion for some users. I started learning coding by reading materials in both scripts and I got used to the differences but I can imagine it being a roadblock if the materials you have on hand contradicts what p5.js uses.

It is an interesting problem as the two usually is managed as two separate languages while being for the most part the same language, from the maintenance perspective though I think it is better to see them as two separate languages.

catarak:
thank you so much @codemee and @limzykenneth for explaining the differences! this is really, really helpful—seems like it's absolutely the right option to support both translations. 

one last question—I know @almchung has been working on translations for the p5.js friendly errors, and we'd like to change the language of the friendly errors when the language in the p5.js editor is set. if the friendly errors only have a Simplified Chinese translation, would it make sense to display that translation if Traditional Chinese is selected, or would that be super confusing and it would instead make sense to display the default English instead? 

> In fact, here in [Flag Tech](https://www.flag.com.tw/maker/pd), we are designing a studying kit for teaching ML5.js using P5.js Editor for interacting with micro-controlled board. So we are look foward to Traditional Chinese version of P5.js Editor.

this is so awesome! please share with the p5.js team whenever you would like: [hello@p5js.org](mailto:hello@p5js.org)

limzykenneth:
> one last question—I know @almchung has been working on translations for the p5.js friendly errors, and we'd like to change the language of the friendly errors when the language in the p5.js editor is set. if the friendly errors only have a Simplified Chinese translation, would it make sense to display that translation if Traditional Chinese is selected, or would that be super confusing and it would instead make sense to display the default English instead?

I would suggest displaying default English in this case. Essentially like I said above, treating them as two different languages works most of the time.

almchung:
It is exciting to see a new translation added to p5! @catarak @limzykenneth i18next, which FES translations are based on, provides the fallback option and even allows us to keep the fallbacks different per region ([LINK](https://www.i18next.com/principles/fallback#fallback-to-different-languages)) so this should cause no problem.

codemee:
How can I help to also translate the errors into Traditional Chinese?
Thanks.

Cassie Tarakajian ***@***.***> 於 2021年9月15日 週三 01:49 寫道：

almchung:
@codemee, you can find the doc on the FES internationalization here: ([LINK](https://github.com/processing/p5.js/blob/main/contributor_docs/internationalization.md)).

I'm in the process of linking the Web Editor's internationalization feature with the FES's. If you want to work when this feature is available, please check in again in a month or two. Thank you! 😄

codemee:
@almchung 
OK, thanks.

catarak (approved):

PR #1916, opened 2021-08-19
Resolved Issue no #934

### LINKED ISSUE #934: Merge together the exact same console events

brijsiyag:
Fixes #934

* [x] is descriptively named and links to an issue number, i.e. `Fixes #934`


catarak:
Thanks for working on this issue! I was testing this out, and it was working great. However, after looking at the code, I realized that I didn't agree with the approach—I thought that the code should be removed from the React component and into the [console reducer](https://github.com/processing/p5.js-web-editor/blob/develop/client/modules/IDE/reducers/console.js), as redux is responsible for managing data changes.

After making this change locally, I couldn't get the styling quite write—as the `times` count got really bug, it obscured the console message. I dug around into the `console-feed` library, and it turns out this functionality is now built in! So yeah, that's the reasoning behind all of these changes I made.

catarak:
Create Release Environment

catarak:
Create Release Environment

catarak (approved):

PR #1921, opened 2021-08-27
Fixed #934

### LINKED ISSUE #934: Merge together the exact same console events

shubhamkapoor01:
Fixed #934

logic to solve the issue:
made changes in `client/modules/IDE/components/Console.jsx`
we compare each consoleEvent with previous one and if it's exact same we don't make a new event instead use the same one, ie merge them.

* ✅ has no linting errors (`npm run lint`)
* ✅ is from a uniquely-named feature branch and has been rebased on top of the latest `develop` branch. (If I was asked to make more changes, I have made sure to rebase onto `develop` then too)
* ✅ is descriptively named and links to an issue number, i.e. `Fixes #123`


catarak:
Thanks for working on this! Unfortunately this isn't working for me. You can see this with the following sketch:
```js
function setup() {
  createCanvas(400, 400);
}

function draw() {
  background(220);
  print('hello');
}
```

What I notice happening in the console is that the number will change sometimes, but it's not increasing like it should. If you look in the developer tools console, the number steadily increases.

shubhamkapoor01:
sure i'll look into this today!

catarak:
Fixed by #1916

PR #2022, opened 2022-04-28
Chore/react deprecation warnings

### LINKED ISSUE #2021: Remove React warnings
### How would this new feature help [increase access](https://github.com/processing/p5.js/blob/main/contributor_docs/access.md) to the p5.js Editor?

Improves stability of the application.

When running the p5.js Editor locally, React outputs to the console some deprecation warnings, due to updating the React version. I would like to fix these warnings in an attempt to keep React up to date.

catarak:
Fixes #2021

Changes:
Removes many of the React deprecation warnings. The rest of the warnings can be removed when upgrading to React 18.


client/modules/User/components/CookieConsent.jsx
```diff
@@ -175,10 +175,13 @@ function CookieConsent({ hide }) {
                   kind={Button.kinds.secondary}
                   onClick={acceptAllCookies}
                 >
-                  {t('Cookies.AllowAll')}
+                  <>{t('Cookies.AllowAll')}</>
```
lindapaiste: In my opinion it would be better to to handle this in the `Button` by allowing `string` as a valid `children` type.  This can be done with `children: PropTypes.node`.

raclim:
I upgraded `react-helmet` to its latest version, `v6.1.0`, and `react-tabs` to `3.2.3`, which is [the latest one compatible with our current version of React](https://www.npmjs.com/package/react-tabs). I think https://github.com/processing/p5.js-web-editor/pull/2218 can address the remaining warnings for now.

client/modules/App/App.jsx
lindapaiste, line 1: I have an open PR that converts `App` to a function component but this looks right.

client/modules/IDE/actions/projects.js
```diff
@@ -4,30 +4,33 @@ import { startLoader, stopLoader } from './loader';
...
-      url = '/projects';
-    }
-    return apiClient
-      .get(url)
-      .then((response) => {
-        dispatch({
-          type: ActionTypes.SET_PROJECTS,
-          projects: response.data
+  return (dispatch) =>
+    new Promise((resolve) => {
```
lindapaiste, line 8: I don't think this accomplishes anything?  When we `return apiClient.get().then().catch()` that is already returning a `Promise`.

IMO we should stay away from explicit `new Promise((resolve) => ` syntax *unless* it's for the purpose of converting a callback-based result to a `Promise`.  ie. `someFunction(arg, (err, res) => { if (err) { reject(err); } else { resolve(res); }})`.

You could keep it as-is.  You could also convert it to `async`/`await` (`return async (dispatch) => {`) as `async` functions **always** return a `Promise`.

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -97,26 +97,14 @@ class IDEView extends React.Component {
...
-  }
-
-  componentWillUpdate(nextProps) {
-    if (nextProps.params.project_id && !this.props.params.project_id) {
-      if (nextProps.params.project_id !== nextProps.project.id) {
-        this.props.getProject(nextProps.params.project_id);
+  componentDidUpdate(prevProps) {
+    // if the url has changed to open a new sketch
+    if (this.props.params.project_id && !prevProps.params.project_id) {
+      if (this.props.params.project_id !== this.props.project.id) {
```
lindapaiste, line 103: This is the same condition that was there before...but does it make sense?  It's saying we should load the project if we have an id AND we did not have an id previously AND we have not already loaded the project for that id.  I don't think we want the `&& !prevProps.params.project_id` part.  What if `this.props.params.project_id` changes from one valid id to another valid id?

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -298,7 +286,9 @@ class IDEView extends React.Component {
         <main className="editor-preview-container">
           <SplitPane
             split="vertical"
-            size={this.state.sidebarSize}
+            size={
+              this.props.ide.sidebarIsExpanded ? this.state.sidebarSize : 20
```
lindapaiste, line 290: This was actually something that I discussed with @dewanshDT in his rewrite of this component (he has it converted to a function component).  We came to the same conclusion you did here -- that it's better for the state to just store the expanded size.

client/modules/IDE/actions/projects.js
```diff
@@ -4,30 +4,33 @@ import { startLoader, stopLoader } from './loader';
...
+      dispatch(startLoader());
+      let url;
+      if (username) {
+        url = `/${username}/projects`;
+      } else {
+        url = '/projects';
+      }
+      return apiClient
```
lindapaiste, line 16: There's an eslint rule against using `return` within `new Promise()`.  But maybe that one only showed up in the branch where I updated eslint. https://eslint.org/docs/latest/rules/no-promise-executor-return

With the previous structure, the `return` statement on this line causes it to return a `Promise`.  With the current/proposed structure, the `return` doesn't do anything and should be deleted.

lindapaiste (changes requested):
These are mostly positive changes, though it will create conflicts with some of the work that @dewanshDT has been doing and with some of my existing PRs.

I just looked at the code and I haven't loaded it up to make sure that the tabs still work as expected.

lindapaiste:
> I upgraded `react-helmet` to its latest version, `v6.1.0`

Did that actually fix the warnings?  My understanding is that `react-helmet` is dead and we need to switch to [`react-helmet-async`](https://www.npmjs.com/package/react-helmet-async).

raclim:
Some of these changes have fallen out of date, so I'm going to close this for now!

Issue #2042, opened 2022-07-18
Refactoring Redux code: hooks, thunks, and redux-toolkit


lindapaiste:
Most of the Redux code is this package was written years ago and has not been touched since.  In that time, Redux has come a long way!  The bad news is that this codebase goes against a lot of the [Redux Best Practices](https://redux.js.org/style-guide/).  The good news is that the newer ways of doing things are much easier and involve a lot less boilerplate.

Some things which can be improved/changed:
- Can use [`createSlice`](https://redux-toolkit.js.org/api/createSlice) which automatically creates the action creators for a reducer so that you don't need to [define a bajillion constants](https://github.com/processing/p5.js-web-editor/blob/develop/client/constants.js) or define basic action creator functions.  This can be tackled one "slice" at a time, as Redux Toolkit is fully compatible with old/vanilla Redux.
- Can use [`createAsyncThunk`](https://redux-toolkit.js.org/api/createAsyncThunk) to automatically dispatch `pending`, `error` and `fulfilled` actions for each asynchronous action.
- Can use `useSelector` and `useDispatch` hooks instead of the `connect` HOC.

I can do a lot of this myself, but I wanted to post it as a discussion first because I don't want to waste time rewriting code if it's not wanted.

kjhollen:
thank you @lindapaiste! the editor is currently in maintenance mode as we transition to new leadership (see #2010). hope to have an update next month!

mrkirthi-24:
Is this issue open?

lindapaiste:
> Is this issue open?

Yes! It's a broad topic and a lot of work has been done already but it's not complete.

As far as using hooks instead of `connect`, I think we've got that pretty much everywhere...if you include the PRs that haven't been merged.

We finally merged in a PR that enables Redux Toolkit but most of the reducers haven't been converted.  I've only done the `toast`, `assets`, and `ide` so far.

We have not done anything yet with `createAsyncThunk` (or RTK query).

Maybe I need to make a checklist.  If there's a particular piece that you want to work on you should search the open PRs to see if it's mentioned.

https://github.com/processing/p5.js-web-editor/pulls?q=is%3Apr+is%3Aopen+redux

Ashish2271:
@lindapaiste  i have gone through all the pr's and your work on this conversion  i would like to work on rest of reducers and implentation of RTK query

ASCE-D:
[...]

Issue #1458, opened 2020-06-15
Refactor IDEView


catarak:
#### Nature of issue?

- Existing feature enhancement

#### Feature enhancement details:

In order to work on making the web editor mobile friendly, @ghalestrilo suggested refactoring and breaking apart `IDEView.jsx`. I, personally, have wanted to refactor this component for a while, so I think now is a great time to work on this.

catarak:
Suggestions from @ghalestrilo:

### Behavior

- move method getTitle to pure ext. function receiving props obj
- move mehotd isUserOwner to pure ext. function receiving props obj
- move warnIfUnsavedChanges to pure ext. function which:
  - receives props obj
  - returns a callback (which receives route) (because of how it will be passed to the child component)
  - note: this function does too much: it should be further split

- move method handleGlobalKeydown to:
  - ext. function receiving props obj
  - ext. function receiving callbacks obj (derived from props)
  - redux action

### Rendering

- Lines 210 to 246: move to functional component `<Layout />`

- Lines 335 to 376: create functional component `<PreviewWrapper />` and wrap `<PreviewFrame />` inside it:

```jsx
<PreviewWrapper>
  <IDEView />
</PreviewWrapper>
```

- Lines 382 to 462: depending on how many prop passes are required
  - move to functional component `<IDEOverlays />` 
  - move to connected component `<IDEOverlays />`
  - Create provider-like wrapper `<WithOverlays />` and wrap the `<IDEView />` inside it:

```jsx
<WithOverlays>
  <IDEView />
</WithOverlays>
```

### Lifecycle

- (optional) move code from lifecycle methods to ext functions (ensuring they're correctly bound to this)
- refactor component into hooks

"I broke it down into 3 parts to be a bit more manageable. And just getting the Behavior section will already make the migration a lot easier, cause it contains most of the core functionality. The rest is good but less important."

ghalestrilo:
Analyzing/grouping the props received by `<IDEView />` seems helpful to understand what responsibilities are being coupled in this view. I hope this can bring some insight for the refactor

```js
// Modal Props: Could be provided through an overlay component
const modalProps = {
  toast: PropTypes.shape({
    isVisible: PropTypes.bool.isRequired
  }).isRequired,
  
  showErrorModal: PropTypes.func.isRequired,
  showEditorOptions: PropTypes.func.isRequired,
  showKeyboardShortcutModal: PropTypes.func.isRequired,
[...]

PR #2043, opened 2022-07-19
use react-redux hooks in Toast component

lindapaiste:
Related to issues #1458 & #2042 

Changes:
* Refactor the `Toast` component to use `react-redux` hooks instead of the `connect` HOC, in line with [Redux best practices](https://redux.js.org/style-guide/#use-the-react-redux-hooks-api).
* Handle the `isVisible` check inside of the `Toast` component itself instead of in the parent.  This means that we only need one `if` instead of three, and the parents no longer need to access any toast state from the store.


3ru (approved):
LGTM:+1:

raclim (approved):
Looks great :), thanks so much for your patience!

PR #1602, opened 2020-09-04
Add new file/folder bugfix

neelesh7singh:
Fixes #1003 
I'm adding an event listener whenever newFile or newFolder model is opened which close the models if clicked outside the models.
I have also changed the ide reducer a little bit so that both models can't be opened at the same time in anyways.

I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] is from a uniquely-named feature branch and has been rebased on top of the latest master. (If I was asked to make more changes, I have made sure to rebase onto master then too)
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #2049, opened 2022-08-06
Combined Modal component for New File, New Folder, and Upload.

lindapaiste:
Changes:
The components for `NewFileModal`, `NewFolderModal` and `UploadFileModal` already use the same CSS class names for styling.  This PR moves the HTML structure of the modal and the click-outside logic (added in #1602) into a common shared `Modal` component.  Any differences between the three are handled by passing props to the `Modal`, such as `title` and `closeAriaLabel`.


client/modules/IDE/components/Modal.jsx
```diff
@@ -0,0 +1,64 @@
...
+  onClose,
+  closeAriaLabel,
+  contentClassName,
+  children
+}) => {
+  const modalRef = useRef(null);
+
+  const handleOutsideClick = (e) => {
+    // ignore clicks on the component itself
+    if (e.path.includes(modalRef.current)) return;
```
raclim: I think this line may need some type of additional check, since it seems to return an error when `modalRef.current` is undefined. Otherwise, I feel like everything else looks great! :)

lindapaiste:
Two notes re: close on outside click.

1. I did not change the logic for handling outside clicks, but I realize now I should. I am checking the same way as before with `e.path.includes(modalRef.current))`.  I recall that it was working for me at the time I submitted this PR, but I must have updated Chrome since then as `e.path` was [removed in v109](https://chromestatus.com/feature/5726124632965120) and it no longer works for me.

> Modal.jsx?cf57:19 Uncaught TypeError: Cannot read properties of undefined (reading 'includes')
    at HTMLDocument.handleOutsideClick (Modal.jsx?cf57:19:1)

We can change it to `e.composedPath()` as [explained here](https://stackoverflow.com/a/39245638).

There is another approach `ref.current.contains(e.target)` which is used [here](https://github.com/processing/p5.js-web-editor/blob/269ca2292886b76a474278b87f35376ad1c0bd62/client/modules/User/components/Collection.jsx#L37) and [here](https://github.com/processing/p5.js-web-editor/blob/dd433b768da0a08dfda547ffc37a5624c37de4f3/client/modules/IDE/hooks/custom-hooks.js#L35).

2.  Currently the `New File` and `New Folder` modals close on outside click but the `Upload` does not.  Is that what we want?  In this PR I applied the closing behavior to all three, but I can remove it from `Upload` by adding an additional prop like `disableOutsideClick` or something.

lindapaiste:
@raclim I fixed note 1 with the `ref.current.contains(e.target)` approach.  This should be good to go unless you have any thoughts about note 2.

raclim (approved):
Hi @lindapaiste, thanks so much for working on this—sorry it took a while to revisit it! These updates makes sense to me, and I also think it looks good to go! 

Since we just created a new release I'd like to wait for a moment before addressing more PRs as a precaution, so this will probably be merged in by the end of the week!

lindapaiste:
> Currently the New File and New Folder modals close on outside click but the Upload does not. Is that what we want? In this PR I applied the closing behavior to all three, but I can remove it from Upload by adding an additional prop like disableOutsideClick or something.

@raclim I realized why we did not have the click-outside behavior on the upload modal previously.  It's because clicking the "Drop files here or click to use the file browser" section is considered an outside click and therefore closes the modal.  The `e.target` of that click is the hidden `<input type="file" multiple="multiple" class="dz-hidden-input" ... />` element which is a direct child of the `body`.  Tbh I don't understand how that becomes the `e.target` rather than the `<div id="uploader" class="uploader dropzone dz-clickable">` but it is.

I can fix this by making it so that the hidden input is inside of the modal `div` instead of in the `body`.  There is a `hiddenInputContainer` [option](https://docs.dropzone.dev/configuration/basics/configuration-options) for this.

Please do not push this production until we merge in a fix.

raclim:
Sounds good, thanks for catching this—I was just about to push the release 😂 I'll hold off until then

PR #2052, opened 2022-08-07
Move keydown handling out of IDEView

lindapaiste:
Progress on #1458 and #824

Should be merged after #2049 as I moved the `Escape` key handling into the new `Modal` component.

Changes:
* Create new utility hook `useKeyDownHandlers` for adding one or more keydown handlers to the `document`, and component `DocumentKeyDown` which wraps the hook for use in class components.
* Apply hook/component to `Nav`, `Overlay`, and `Modal`
* Move the keydown handling from `IDEView` into new hook/component `useIDEKeyHandlers`/`IDEKeyHandlers`, which accesses redux props directly and allows a few props to be removed from `IDEView`.
* Use the same keydown handling in `MobileIDEView` instead of repeating it.


client/modules/IDE/hooks/useKeyDownHandlers.js
```diff
@@ -0,0 +1,55 @@
...
+  }, [keyHandlers]);
+
+  /**
+   * Will call all matching handlers, starting with the most specific: 'ctrl-shift-f' => 'ctrl-f' => 'f'.
+   * Can use e.stopPropagation() to prevent subsequent handlers.
+   * @type {(function(KeyboardEvent): void)}
+   */
+  const handleEvent = useCallback((e) => {
+    const isMac = navigator.userAgent.toLowerCase().indexOf('mac') !== -1;
+    const isCtrl = isMac ? e.metaKey && this.isMac : e.ctrlKey;
```
lindapaiste: I’m looking back at this and realized that the `&& this.isMac` is something from the old code that I forgot to delete.  Will fix.

raclim:
This looks great for me so far! I think the only issue I've noticed is that the sketch will still run despite using the Keyboard shortcut `⌘ + ⇧ + Enter` to stop it.

raclim (approved):
I think I found the source of my issue with stopping the sketch, which was in lines 34-38 in the `useKeyDownHandlers.js` file. Within those lines, I think the order in which the handlers were being called might not be consistent across different browsers/devices, which is why my sketch would restart again. I updated the check to use an `else if` statement, and also removed a few linting warnings and errors. 

I'm sorry again that this took a while to review and thank you for making these really awesome changes!

lindapaiste:
@raclim I need to put in a quick bug fix because apparently `e.key` is not always defined 🤦 .  It [might just be Chrome](https://stackoverflow.com/questions/59534586/google-chrome-fires-keydown-event-when-form-autocomplete) but the autocomplete form filling sends a `keydown` event with no `e.key` which causes an error.

image
image

raclim:
@lindapaiste, sounds good, thanks for catching this and giving a heads up!

PR #2054, opened 2022-08-08
Fix prop-drilling in `IDEView`

### LINKED ISSUE #824: Prop Drilling vs. Redux

### LINKED ISSUE #2053: Increase font size button can concatenate `20` => `202` instead of adding
The "Increase" and "Decrease" font size buttons can have strange behavior if they are clicked after typing in the input.

To reproduce, click on the input and type `20`.  Then, _without hitting Enter_, click the "Increase" button.  You should see this:
image

If the user hits Enter to submit the form then the "Increase" and "Decrease" buttons will work as expected.

lindapaiste:
This is a draft because the unit tests for `<Preferences/>` need to be updated.

Fixes #824
Fixes #2053 
Continues work on #1458 
Expands on some of my previous PRs.

Changes:
### Props Drilling
I handled the following components:
- `Preferences`
- `Sidebar`
- `ShareModal`

The `IDEView` no longer passes down any props to these components.  They now take no props and access everything via react-redux hooks.  I converted all three from class to function components.

### Code Breakup
I moved all of the overlays from `IDEView` to a new component `IDEOverlays`, as suggested [here](https://github.com/processing/p5.js-web-editor/issues/1458#issuecomment-644277911).  This new component is connected to the redux store which means that more prop drilling goes away, as the `IDEView` no longer needs to deal with the `isVisible` and `close` logic for the modals and overlays.


dewanshDT:
hey is someone currently working on this?

lindapaiste:
😭 I did a rebase and fixed the conflicts but now it's showing 81 commits instead of 9.  It should be https://github.com/processing/p5.js-web-editor/compare/develop...lindapaiste:p5.js-web-editor:refactor/props-drilling

lindapaiste:
Just a quick check-in: I did end up refactoring the tests to work with new propsless/connected design.

There are still two tests which fail with regards to handling non-numeric inputs on the text size input.  I changed this from an input with type “text” to type “number” which means that there is some automatic validation by the DOM.  So I don’t think there is any actual problem, as you cannot type letters into the input.  But it does cause the tests to fail 🤷🏻‍♀️

raclim:
Thanks for your update on this! Hmm, that makes sense to me! I'll also think on this on my end and will update you with any thoughts!

lindapaiste:
I'm going to close this as I have re-submitted most of the changes in individual PRs.  

Reminder to myself -- I still need to revisit changing the font size input to `type="number"`, which is the part that broke the tests.   And also implement these clean ups of the JS.
```
 const validateFontSize = (number) => {
    if (Number.isNaN(number)) return 16;
    return clamp(number, 8, 36);
  };

  // The current contents of the input, which may be invalid. String or number.
  const [fontSizeString, setFontSizeString] = useState(fontSize);
  const fontSizeNumber = parseInt(fontSizeString, 10);

  // Immediately submit any valid fontSize input
  useEffect(() => {
    if (validateFontSize(fontSizeNumber) === fontSizeNumber) {
      dispatch(setFontSize(fontSizeNumber));
    }
  }, [fontSizeNumber]);

  // Handler for increase and decrease.
  const updateFontSize = (value) => {
    const valid = validateFontSize(value);
    setFontSizeString(valid);
    dispatch(setFontSize(valid));
  };

  const onFontInputSubmit = (event) => {
    event.preventDefault();
    const valid = validateFontSize(fontSizeNumber);
    dispatch(setFontSize(valid));
  };

  const fontSizeInputRef = useRef();
  ```

PR #2118, opened 2023-02-17
added web editor and p5 version info to the about section

### LINKED ISSUE #2070: Display editor and p5.js version numbers in 'about' menu
Would be super useful in tracking issues to easily know what versions one is using...

dewanshDT:
Fixes #2070

Changes:
I've made changes to the `About.jsx` to show the version info of the editor and the p5.js, please check the commits to learn more.

p5about


client/modules/IDE/components/About.jsx
```diff
@@ -1,13 +1,21 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { Helmet } from 'react-helmet';
+import axios from 'axios';
 import { useTranslation } from 'react-i18next';
 import { Link } from 'react-router';
 import SquareLogoIcon from '../../../images/p5js-square-logo.svg';
 // import PlayIcon from '../../../images/play.svg';
 import AsteriskIcon from '../../../images/p5-asterisk.svg';
+import packageData from '../../../../package.json';
```
lindapaiste, line 9: This is great!

client/modules/IDE/components/About.jsx
```diff
@@ -1,13 +1,21 @@
...
 
 function About(props) {
   const { t } = useTranslation();
+  const [p5version, setP5version] = useState(0);
+  useEffect(() => {
+    axios
+      .get('https://cdn.jsdelivr.net/npm/p5/package.json')
```
lindapaiste: This part has the potential to be incorrect because it will always fetch the latest available version number of p5, but the editor could be a bit behind with the version that it's using.

I'm thinking as I type so bear with me...

The version number in the default files is [hard-coded here](https://github.com/processing/p5.js-web-editor/blob/050b44741da24228954a3774e238c02e938e3803/server/domain-objects/createDefaultFiles.js#L12-L13).  We could export a `versionNumber` constant that we use in both places and that would be pretty good.

There's still a chance that someone is editing a previously saved sketch that they created with an even older version of p5.  If we want to ensure that the modal displays the version number that is used in their sketch, the safest way to do that would be to look at the actual content of the file.  It's possible that their `index.html` loads the library from a different CDN or with an `@latest` URL so this would not be guaranteed to find a match.  If no match is found, then we could display the version from the default files.  Or we could display "unknown" or don't show anything if it's not found.

Finding it might look like:

```
const version = useSelector((state) => {
    const index = state.files.find(file => file.name === 'index.html');
    return index?.content.match(/\/p5\.js\/([\d.]+)\//)?.[1];
});
```
That's getting the `index.html` file from the Redux state and looking for a text that matches `/p5.js/1.5.0/` where the version number can be any series of numbers and periods (aka `[\d.]+`).

dewanshDT: I agree on this, It is important to display the correct version number especially if they are using the older version of p5js.
I'll try this 😃

dewanshDT:
@lindapaiste I updated as you suggested, and It works flawlessly 🔥, take a look

dewanshDT:
I think this works great please check this out and there's nothing more to work on this PR unless any UI changes are required.

raclim (approved):

Issue #2134, opened 2023-02-26
Improper setup of obsolete package `react-hot-loader`


lindapaiste:
While working on other features I have encountered a bunch of errors, warnings, and problems relating to the `react-hot-loader` package.  This is a hindrance to the future development of this project and provides unnecessary obstacles to development.

### Package is obsolete
**The best way to fix all of these issues is to not use this package.**  The package itself recommends that you don't use it anymore:
> React-Hot-Loader has been your friendly neighbour, living outside of React. But it has been limiting its powers and causing not the greatest experience. It's time to make a next step.

> 
> React-Hot-Loader is expected to be replaced by [React Fast Refresh](https://github.com/facebook/react/issues/16604). Please remove React-Hot-Loader if Fast Refresh is currently supported on your environment.
[source](https://github.com/facebook/react/issues/16604)

### Doesn't work most of the time
![image](https://user-images.githubusercontent.com/28965286/221435158-d3b0fed8-ee01-4df6-bd45-c0ab1a65b22d.png)
I usually see the "hot update was not successful" message and need to manually refresh the page.

### Improper setup
![image](https://user-images.githubusercontent.com/28965286/221434809-199a7960-36d2-4390-a798-9fa99f64a0b8.png)
I have seen an error related to using `hot(App)` and `ReactDOM.render()` in the same file.  We are definitely doing that!

https://github.com/processing/p5.js-web-editor/blob/9073487232b5f53756c3525981e992cee0fcd0fc/client/index.jsx#L31-L33

### Missing peer dependency
![image](https://user-images.githubusercontent.com/28965286/221434964-9143dbb5-8925-4c1c-a85e-da03821d0f6e.png)
[Here's the discussion on that issue](https://github.com/gaearon/react-hot-loader/issues/1227)

### Can create other problems
![image](https://user-images.githubusercontent.com/28965286/221435226-677bb49d-0164-456a-af64-f0a5d5bc10dd.png)
Here's an issue that I encountered while working on the `react-router` code.  The `react-hot-loader` package converts all components to a proxied version which breaks the router ([discussion](https://stackoverflow.com/a/70243956/10431574)).  I am able to work around this by writing the router code differently, but it's annoying.

PR #2135, opened 2023-02-26
[WIP] Upgrade `react-router` from "^3.2.6" to "^6.8.1"

lindapaiste:
Fixes #875

---

My work on this is PR paused because I've gotten far enough to see that updating `react-router` breaks the `redux-auth-wrapper` package.  I need to refactor out that package before continuing.

I also need to find a way to implement the legacy `setRouteLeaveHook` which is used by `IDEView` and fix some tests which are missing the router provider.

---

Changes:

Updates `react-router`/`react-router-dom` from version 3.2.6 to current version 6.8.1.  This is up three major versions so there are a lot of changes.  It's very easy to update function components but a real pain to update class components.

Setup/Configuration:
* `router-compatibility.jsx`
   * Create utility `ElementFromComponent` for backwards compatibility with components that are created inline and therefore difficult to instantiate via JSX.  Can now convert from `component={userIsAuthenticated(AccountView)}` to `element: <ElementFromComponent component={userIsAuthenticated(AccountView)} />`.
   * Create HOC `withRouter` similar to the one that used to be exported from the `react-router` package.  This is needed to access `params`, `location`, and `navigate` in legacy class components which cannot use hooks.
* `./routes.jsx`
   * The `routes` variable no longer depends on the `store`, as individual routes can access Redux state via context.
   * Created a wrapper `Main` around the `App` in order to handle legacy `onChangeRoute` and `onEnter` function.  Renders the child content using new `<Outlet/>` component.
   * Routes now use `element` instead of `component`.
   * I had to convert from `<Route/>` elements to route objects.  It *should* have been possible to keep the `<Route/>` syntax and convert it using the `createRoutesFromElements` utility, however this did not work due to the hot reloader package that we are using.  [Relevant discussion](https://stackoverflow.com/a/70243956/10431574)  See also #2134 
* `./router.js`
   * Create the router object with `createBrowserRouter`.
   * I wanted this is in a separate file in order to export the `navigate` function with no dependencies on other files, but it depends on the `./routes` so...not the best.

Components:
* Change import paths for `Link` from `react-router` to `react-router-dom`.
* Use the hooks `useParams`, `useNavigate`, and `useLocation` in function components.
   * Can remove some properties from `props` and `propTypes`.
* Replace all `browserHistory.push()` calls with `navigate()`.
* Deleted two navigations methods from `DashboardView` which are not actually used.

Etc:
* In redux actions, access the `navigate` function as a property of the global router object.
* Refactor the `createRedirectWithUsername` function.

---


raclim:
Thank you so much for all of your work on this so far! In the spirit of organizing some PRs, unfortunately I think I'm going to close this one for now due to the immensity of the changes, and also that I'd usually prefer to merge these types of updates incrementally!

lindapaiste:
@raclim It's going to be hard to break this up into incremental changes because we cannot use v6 APIs without upgrading the package, and we cannot upgrade the package version without using the correct APIs.  The only way to make it more incremental would be to upgrade one version at a time (3 to 4, 4 to 5, then 5 to 6).

raclim:
@lindapaiste Yes I was thinking along those lines, or some other way to break this down into a smaller set of changes, even if it's just two larger updates. Sorry I wasn't specific about what I meant by it! Please feel free to rework this PR and reopen it, or use it to create new ones if you'd still like to tackle this!

Issue #2140, opened 2023-02-28
No favicon


parteekcoder:
@raclim 
Reproduce

1. open sketch any other user like https://editor.p5js.org/greenStone83/sketches/IQMmFIkYv
2. you will find there is no favicon

Reason: it is making request to https://editor.p5js.org/greenStone83/sketches/favicon.ico for favicon
instead  should make to https://raw.githubusercontent.com/processing/p5.js-website-OLD/master/favicon.ico

parteekcoder:
![image](https://user-images.githubusercontent.com/100110395/221856254-9987256c-79a4-436f-85f2-9ed6a33eaaca.png)

![image](https://user-images.githubusercontent.com/100110395/221856285-f061b71a-9ed1-4c3e-abd6-bdf7b39452d9.png)

Noothan-am:
hi @parteekcoder ,I would like to fix this issue can you please assign it to me..?

parteekcoder:
thanks @Noothan-am , but I already created a PR for this see #2141

Skar3Krow:
Hey @parteekcoder , I would like to work on this issue if it's still not fixed.

PR #2141, opened 2023-02-28
favicon issue resolved

parteekcoder:
issue #2140 solved

@raclim please review this PR and suggestions of your will be admired

### DIFF
--- a/server/views/index.js
+++ b/server/views/index.js
@@ -12,7 +12,7 @@ export function renderIndex() {
       ${process.env.NODE_ENV === 'production' ? `<link rel='stylesheet' href='${assetsManifest['/app.css']}' />` : ''}
       <link href='https://fonts.googleapis.com/css?family=Inconsolata' rel='stylesheet' type='text/css'>
       <link href='https://fonts.googleapis.com/css?family=Montserrat:400,700' rel='stylesheet' type='text/css'>
-      <link rel='shortcut icon' href='favicon.ico' type='image/x-icon' / >
+      <link rel='shortcut icon' href='https://raw.githubusercontent.com/processing/p5.js-website-OLD/master/favicon.ico' type='image/x-icon' / >
       <script>
         if (!window.process) {
           window.process = {};


lindapaiste (changes requested):
Great catch on the malformed URL!  Once again I think there's a better solution though, because we actually *do* have the file included in the website so we don't need to load it from GitHub.

The current url path `href='favicon.ico'` is *relative* so it looks for the file in the current directory.  That's why it resolves to https://editor.p5js.org/greenStone83/sketches/favicon.ico which is not found.

We can change it to an *absolute* path by starting with a `/`.  The path `href='/favicon.ico'` will look for the file in the root directory and this URL https://editor.p5js.org/favicon.ico does load.

parteekcoder:
yeah @lindapaiste that is also a good suggestion , but let @raclim @catarak decide as here also is same thing at line https://github.com/processing/p5.js-web-editor/blob/develop/server/views/404Page.js#L48

But thanks @lindapaiste I am always open to new suggestions

parteekcoder:
@raclim @catarak @davepagurek  plz review this PR , your suggestions are admired by me

raclim:
Thanks for working on this! I'm down to try out and follow through with @lindapaiste's suggestion if it sounds good to you!

parteekcoder:
> Thanks for working on this! I'm down to try out and follow through with @lindapaiste's suggestion if it sounds good to you!

is @lindapaiste also maintainer of this repo

parteekcoder:
> Thanks for working on this! I'm down to try out and follow through with @lindapaiste's suggestion if it sounds good to you!

I want to contribute to p5web-editor , I am understanding the codebase and solving the issues to have a better understanding

parteekcoder:
> Thanks for working on this! I'm down to try out and follow through with @lindapaiste's suggestion if it sounds good to you!

yeah , I am also always open to new suggestions

lindapaiste:
> > Thanks for working on this! I'm down to try out and follow through with @lindapaiste's suggestion if it sounds good to you!
> 
> is @lindapaiste also maintainer of this repo

I'm not a maintainer, I just have a lot of opinions 😜.  I think @raclim is the only active maintainer at the moment.

raclim:
I'm going to close this PR for now, but please feel free to revisit this or create a new one! :)

Issue #2071, opened 2022-10-25
Download Feature not working for projects containing folders


Andriy-Kulak:
#### Details about the bug:

- p5.js version: 1.4.1
- Web browser and version:  Brave | 1.44.112 Chromium: 106.0.5249.119 (Official Build)
- Operating System: MacOs
- Steps to reproduce this:
 
When I go to file and download. it opens a new tab but doesn't download anything... Please assist asap! Thanks

kjhollen:
Hmm, it appears to be working for me. How long did you wait with the blank tab open? I had to wait ~10 seconds before the download started and the tab was blank, then closed automatically when the download started.

tjvezina:
I think I'm seeing the same issue after the fix for #2038, which seems to be related to folders in a project - I'm able to download simple projects, but any that include a folder, even an empty one, fails to download and shows a 502 error after 60 seconds.

hol-lie:
I believe I am seeing a similar issue which seems to be related to the use of asset folders in a project. Another thing to be noted is that I am using the p5.play library. When I try to download this project, I believe there is an issue with the .zip extension, giving me an empty file. It seems like maybe the .zip file extension is instead using a / file seperator see the attached image below.
<img width="530" alt="Screen Shot 2022-12-13 at 4 38 53 PM" src="https://user-images.githubusercontent.com/120525177/207459985-0b377a62-7dbc-4bec-bcf7-8ecfc9b3f021.png">

arcanosam:
> I think I'm seeing the same issue after the fix for #2038, which seems to be related to folders in a project - I'm able to download simple projects, but any that include a folder, even an empty one, fails to download and shows a 502 error after 60 seconds.

I'm having the same trouble as you can see below.

Indeed my project have folders too

Web browser and version: Firefox 109.0 and 109.0.1

Operating System: Windows 11

![2022_02_01_download_bug](https://user-images.githubusercontent.com/1272609/216024005-23edc3f5-7f67-4321-863c-aafa4154ae94.jpg)

kjhollen:
Thanks all for the additional info. Just updating the title to reflect how the bug shows up.

parteekcoder:
@raclim this issue is solved in the PR #2153

PR #2153, opened 2023-03-08
downloadzip-solved

parteekcoder:
fixes issue #2071 

hey @raclim @catarak please review this PR , your suggestions are admired

Problem:
When there is empty folder in projects , it don't enter this line https://github.com/processing/p5.js-web-editor/blob/develop/server/controllers/project.controller.js#L226 hence it doesnot encounters `resolve()` statement

Solution:
As empty folder don't have childrens we can directly resolve that


raclim (approved):

parteekcoder:
thanks @raclim for reviewing this PR

PR #2168, opened 2023-03-20
Generate the 404 page using React on the frontend.

### LINKED ISSUE #2159: Discussion: Why is the 404 page rendered by the server?
Right now the [404 not found page](https://github.com/processing/p5.js-web-editor/blob/develop/server/views/404Page.js) is generated as string of HTML and CSS and is returned directly by the server, no React involved.

The most obvious downside of this approach is that the text on this page will never get translated to the user's current language, which is a barrier to accessibility.

I'm researching this a bit and realizing that we *do* need to serve something different from the server in order to get the correct HTTP status code on the page (404 rather than 200).  But it seems like the server should return more of a skeleton and let React inject the content.  In fact, I think it could still return the same [index page](https://github.com/processing/p5.js-web-editor/blob/develop/server/views/index.js)?  react-router can be set up to display the "page not found" content when there is no matched route.

### LINKED ISSUE #2256: Improper loading of static files on 404 pages
### p5.js version

_No response_

None

### Web browser and version

_No response_

### Actual Behavior

image

This one is not easily reproducible because the 404 pages choose a sketch at random, so I can't link to a particular URL where it happens.  However I noticed it on a 404 page which attempted to render this sketch: https://editor.p5js.org/p5/sketches/Advanced_Data:_Load_Saved_Table

This sketch loads a `.csv` file:
```
function preload() {
  table = loadTable("assets/bubbles.csv", "header");
}
```

The `/server/views/404Page.js` file re-writes that path:
https://github.com/processing/p5.js-web-editor/blob/1c05241e9ead4bc6f622188ba205c3329268fc72/server/views/404Page.js#L123-L131

And we get a URL https://rawgit.com/processing/p5.js-website/main/dist/assets/examples/assets//bubbles.csv which does not exist.  The rawgit.com service has actually been deprecated since 2018 and it should not be used.  But it does still work if we use the correct URL, which is https://rawgit.com/processing/p5.js-website/main/src/data/examples/assets/bubbles.csv as the file is located at https://github.com/processing/p5.js-website/blob/main/src/data/examples/assets/bubbles.csv.  The fatal problem is that we are creating improper paths.

### Expected Behavior

When the sketch is run by the `previewServer`, the output code is:
```
function preload() {
  table = loadTable("https://cdn.jsdelivr.net/gh/processing/p5.js-website@main/src/data/examples/assets/bubbles.csv", "header");
}
```

This works, and it matches what is included in the `.files` property of the sketch object:
```
{
    "name": "bubbles.csv",
    "content": "",
    "children": [],
    "fileType": "file",
    "_id": "6459341065978bfc7dc6c7b4",
    "url": "https://cdn.jsdelivr.net/gh/processing/p5.js-website@main/src/data/examples/assets/bubbles.csv",
    "createdAt": "2023-05-08T17:40:32.645Z",
    "updatedAt": "2023-05-08T17:40:32.645Z",
    "id": "6459341065978bfc7dc6c7b4"
}
```

### Steps to reproduce

It happens at random on any 404 URL, depending on which sketch is chosen.

lindapaiste:
Fixes #2159
Fixes #2256 

Screenshot 2023-03-19 190246

Done:
* [x] Copy the HTML and CSS from the server 404 page to the frontend.
* [x] Support translating the page texts (needs the translations).
* [x] Fetch a random sketch from user p5.
* [x] Display sketch using existing `EmbedFrame` component.
* [x] Disable the portions of  the `EmbedFrame` code for sending errors to the IDE console from when it is used on the 404 page (via a new `sendMessages` prop).
* [x] Adjust the canvas size to fill the area below the message only.
* [x] Add a link to the sketch.
* [x] Show the toolbar at the top of the 404 page.
* [x] Remove malformed assets URL replacement. (#2256)
* [x] Add a catch-all `react-router` route to show the 404 page when no route matches.
* [x] Change the catch-all server route to return the `renderIndex()` content with a 404 status.
* [x] ~~Cleanup server "object exists" checks with `async`/`await` syntax.~~
* [x] Show the frontend 404 page when a matching user/collection/sketch object is not found by passing a variable `window.__SERVER_404__` from server to client.

To Do:
* [ ] Figure out the correct settings for the `textOutput` `gridOutput` and `basePath` props.
* [ ] (maybe?) Create an API endpoint to return a single random sketch, instead of fetching all.


lindapaiste:
@raclim I've got this to a point where it's working and I feel that it's a major improvement.  Let me now if you have any questions about the code or any other feedback.

We will eventually want to add translations of the 404 page text for all languages.

There are some minor TODOs remaining.  ~~The missing env/config vars `"PREVIEW_SCRIPTS_URL"` and `"EDITOR_URL"` cause warnings in the console but no other problems.  From what I can tell these vars are used for dispatching messages from the  embedded sketch to the editor so that they can be displayed in the console.  That behavior is not needed here.  Ideally I would figure out how to avoid executing those parts of the code.~~ done.

lindapaiste:
@raclim I removed the refactoring of `userExists` etc. to make this clearer.  I will submit those changes in a separate PR.

lindapaiste:
ugh I need to fix the import of the `Nav` now that we refactored it and moved things around.

raclim:
Thanks so much for keeping up and working on all of these different components and changes so far!

PiyushChandra17 (approved):
Nicely Done 👍

PR #2169, opened 2023-03-20
Fix legacy document

3ru:
### Changes:

Fixed to refer to commit

- Fix broken links
- Fix errors on markdown

For more information, please read the comments on this PR.

---


lindapaiste:
This raises an interesting question about how we ensure that the links you've just updated will stay up to date in the future.

I'm not sure if it makes more sense to use the permalink version of the URL, `https://github.com/processing/p5.js-web-editor/blob/edae248eede21d7ad7702945929efbcdfeb4d9ea/client/modules/IDE/components/Sidebar.jsx#L88`, which points to line 88 in a specific commit.
**Pros**: we know that this link will **always** contain `tabIndex="0"`, even if the component code changes and that line isn't on line 88 any more.
**Cons**: the code in this link might not match the current component code.

3ru:
That's right.
This referencing method does not keep up with the development of the code.
Referencing commits is certainly a bit more robust than the status quo, but it may also delay our noticing the problem. (I noticed the problem with this document this time because it didn't actually make sense where I clicked.)


Having considered both sides, this time, I will modify it to refer to commits. This is because there is currently a PR in this repository that makes a very valuable but destructive change, and the reference could be broken immediately after this PR.

Let us at least be aware of and remember this issue so that we can address it again in the future when the documentation and codebase diverge.

raclim (approved):

PR #2176, opened 2023-03-22
Dashboard back btn fix

### LINKED ISSUE #2175: "Back to Editor" should return to your Sketch with the unique Sketch URL
I have noticed that if you are currently in a sketch that you are working on, have had it saved, and have the sketches unique URL in the Browsers URL bar... if you go to the "Sketches" or "Collections" menu item and then select "Back To Editor"... it returns you to the editor, it returns you to the source code you were last working on, but the URL is no longer the unique URL, it is https://editor.p5js.org/

Why is this an issue? For people who are using the P5 editor for a class, it is common to share the URL with their classmates. I currently teach Code! 2 at NYU and many of my students have been having trouble getting the unique URL back in the URL bar of the browser.. If a beginner programmer does the steps that I have described above, it is actually tricky to get the Unique URL back, it does not come back by playing or saving... you need to click back on the "Sketches" menu, select your sketch and re-open it, even though it is already opened.

My suggestion is that you should keep track of the URL of the current project, and the "Back To Editor" link should be a URL that brings you back to the project. (I assume the javascript "back" function may be used instead, which is causing the issue)

dewanshDT:
Fixes #2175 

Changes:
removed hardcoded link `'/'` in `<Link>` in `nav__back-link` and added use of `browserHIstory` which fixes the issue.

### DIFF
--- a/client/components/Nav.jsx
+++ b/client/components/Nav.jsx
@@ -246,7 +246,6 @@ class Nav extends React.PureComponent {
           />
         </li>
         <li className="nav__item nav__item--no-icon">
-          {/* <Link to="/" className="nav__back-link"> */}
           <button
             className="nav__back-link"
             onClick={(e) => {
@@ -263,7 +262,6 @@ class Nav extends React.PureComponent {
               {this.props.t('Nav.BackEditor')}
             </span>
           </button>
-          {/* </Link> */}
         </li>
       </ul>
     );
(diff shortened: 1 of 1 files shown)


lindapaiste:
Going back by 1 in the `browserHistory` will work in some cases but not all.

For example, we show the "Back to Editor" link even if you have not viewed the editor at all.  Navigate directly to https://editor.p5js.org/p5/sketches and you'll see it.

I think we can get the last-viewed sketch id, if one exists, from the Redux `project` state.  I'll try that.

PR #2180, opened 2023-03-23
2084 failing test in toolbarunittestjsx

### LINKED ISSUE #2084: Failing Test in Toolbar.unit.test.jsx
#### Details about the bug:

When tests are run using the command "npm run test:ci", regardless of the branch the user is in or the changes that have been made, one test regarding the Toolbar.unit.test.jsx file will fail.

raclim:
Fixes #issue-number

Changes:

### DIFF
--- a/client/modules/IDE/components/Toolbar.jsx
+++ b/client/modules/IDE/components/Toolbar.jsx
@@ -18,6 +18,7 @@ class Toolbar extends React.Component {
     super(props);
     this.handleKeyPress = this.handleKeyPress.bind(this);
     this.handleProjectNameChange = this.handleProjectNameChange.bind(this);
+    this.handleProjectNameClick = this.handleProjectNameClick.bind(this);
     this.handleProjectNameSave = this.handleProjectNameSave.bind(this);
 
     this.state = {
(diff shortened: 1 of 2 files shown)


client/modules/IDE/components/Toolbar.jsx
```diff
@@ -36,6 +37,17 @@ class Toolbar extends React.Component {
     this.setState({ projectNameInputValue: event.target.value });
   }
 
+  handleProjectNameClick() {
+    if (this.canEditProjectName) {
+      this.props.showEditProjectName();
+      setTimeout(() => {
+        if (this.projectNameInput) {
```
lindapaiste, line 44: FYI, there is a newish feature in JavaScript which is a lifesaver when dealing with possibly undefined variables. It's called the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) - `?.`.  You can write `this.projectNameInput?.focus()` instead of an `if` statement.  When `this.projectNameInput` is `undefined` it will not call `focus()`.  I use `?.` all the time and I don't know how I lived without it!

raclim: Thanks for the suggestion, I'll try updating it!

Issue #2184, opened 2023-03-26
Problems with existing mobile components


lindapaiste:
I'm making this issue to document some of the problems with the existing `MobileIDEView` (and other mobile components) and keep track of them as I fix them.  There are a lot of features that haven't been implemented, but I'm focusing more on bugs and mistakes.

Note that none of the mobile UI is available to end users at the moment due to the `MOBILE_ENABLED` flag, which is `false` in production.

* [x] 1. Theme not applied to Preferences:
Fixed by #2185 

![image](https://user-images.githubusercontent.com/28965286/227730929-54341349-8520-4aa7-a57e-6e486165c2be.png)

* [ ] 2. 404 Not Found error when accessing `/preferences` directly:
These URLs need to be added to the `server` in addition to the `client`.  Which is simple enough, but also annoying and feels fragile.

![image](https://user-images.githubusercontent.com/28965286/227731059-ac5f9862-6cfa-4cc1-81ff-6dccf4e7a384.png)

* [x] 3. App crashes when selecting Theme due to invalid value `"Dark"` instead of `"dark"`.
This gets saved to the user's preferences in the API and cannot be fixed by the user.
Fixed by #2185 

![image](https://user-images.githubusercontent.com/28965286/227731397-2181619e-0fa8-47c2-b673-6476855d74b6.png)

* [x] 4. Prop-types warnings:
Fixed by #2185 

![image](https://user-images.githubusercontent.com/28965286/227730972-f8119430-5095-4137-92bc-8f0c2ed25cc2.png)
![image](https://user-images.githubusercontent.com/28965286/227734094-e39ca09e-9563-495b-8934-712baa126f2d.png)

* [x] 5. Props passed to `PreviewFrame` which don't do anything, because that component only accepts one prop (`fullView`):
Fixed by #2185 

![image](https://user-images.githubusercontent.com/28965286/227734150-920391fe-2e18-4901-994b-b920ce99332f.png)

* [ ] 6. Preview doesn't do anything, due to these `postMessage` errors:
I have not figured out why this works on the desktop version but doesn't work here. 
![image](https://user-images.githubusercontent.com/28965286/227734719-bf6eb921-572c-4944-b67d-36720fbcb4ef.png)

dewanshDT:
hey @lindapaiste I'm already working on these issues on this branch [dewansh-mobile](https://github.com/dewanshDT/p5.js-web-editor/tree/dewansh-mobile), but these issues are not well documented

[...]

PR #2185, opened 2023-03-26
Fix problems in `MobileIDEView` (#2184)

lindapaiste:
Builds upon PR #2054 
Progress on issue #2184

---

Changes:

### Preferences:
* `MobilePreferences` is now a wrapper around the main `Preferences` content instead of recreating it.
* Accept a `className` prop on `Preferences` so that it can be restyled with `styled-components`.
* Make the mobile `Screen` be an extension of the desktop `RootPage`, which applies global text colors based on the current theme. So now those text colors are applied to the `Preferences`.
* Delete the `PreferencePicker` which is no longer needed.
* Remove `"MobilePreferences"` key from translations, since this data already exists in `"Preferences"`.

### Navigation:
* Convert function `getNavOptions` into a component `NavMenu` because it's super weird to call hooks inside a plain function.
* Combine the two lists of `NavMenu` items (logged out and logged in) rather than repeating the shared items twice.
* Fix propTypes error in `ActionStrip`.
* Delete `FloatingNav` which is not used.

### MobileIDEView:
* Use `react-redux` hooks `useSelector` and `useDispatch` instead of `connect`.
* Removed a lot of the existing `connect` props by connecting components deeper in the tree instead of props drilling.
* Combined the `withChangeDot` in the `MobileIDEView` and the original lines in the `Editor` into a reusable `UnsavedChangesIndicator` component so that it doesn't need to be repeated in both versions.
* Converted a bunch of hardcoded string labels into translations.
* Copied the autosave logic from `IDEView` into a new component `AutosaveHandler` which can be used by `MobileIDEView`.  I'm considering this a temporary solution because I did not remove the code from `IDEView` so now we have it two places.

#### Etc:
* Remove all of the props passed to `PreviewFrame` as they don't do anything.

---

---

image
image


raclim:
Hey, I noticed you closed #2184, so I'm going to go ahead with closing this as well!
Please re-open this or let me know if that isn't the case and is still in-progress, thanks!

Issue #2186, opened 2023-03-26
Mutation in `files` Redux reducer


lindapaiste:
I am working on setting up the Redux store using the [`@reduxjs/toolkit` `configureStore` function](https://redux.js.org/usage/migrating-to-modern-redux#store-setup-with-configurestore), which automatically includes a lot of functionality such as `thunk`, dev tools, and a check for state mutations in reducers.  That check immediately found one!

![image](https://user-images.githubusercontent.com/28965286/227810474-519f0d6c-2aeb-4951-8c97-0dab51d5829b.png)

https://github.com/processing/p5.js-web-editor/blob/b3e49dab26417aa1f5492ef08984429a1dc64ef1/client/modules/IDE/reducers/files.js#L277-L281

The problematic code is in the `default` case of the reducer, which means that it runs every single time that an action is dispatched anywhere in the store (which is why the error showed up when handling an unrelated action).  So that's not the right place.  The sorting of children should happen only when children are added, removed, or renamed.

Alternatively, the store state could be unsorted and the sorting could be applied in a selector function.  That would require changes to components which access the `state.files` directly, rather than through some `selectFiles` function which could sort.

I could try to fix the mutation within the current reducer code but this will be an extremely satisfying reducer to rewrite using Redux Toolkit (RTK). There are many changes to deeply-nested properties, and those are such a pain to write properly without mutations!  It will be so much cleaner and nicer and easier to read if it's written using RTK, where you can mutate a draft version of the state.  [Related docs](https://redux.js.org/usage/structuring-reducers/immutable-update-patterns).

We actually have this same sort of mutation in multiple places, including the `CREATE_FILE` and `UPDATE_FILE_NAME` cases. This is a mutation because each `file` object in the mapped array is the same object instance as in the original array, and therefore cannot be modified.  But that's hard to understand and I definitely think that best long-term solution is to use RTK.

https://github.com/processing/p5.js-web-editor/blob/b3e49dab26417aa1f5492ef08984429a1dc64ef1/client/modules/IDE/reducers/files.js#L234-L239
https://github.com/processing/p5.js-web-editor/blob/b3e49dab26417aa1f5492ef08984429a1dc64ef1/client/modules/IDE/reducers/files.js#L218-L223

PR #2187, opened 2023-03-27
Install and configure Redux Toolkit

lindapaiste:
This is step 1 of issues #2042 and #2186 

Implements the recommended first step in the [Migrating to Modern Redux guide](https://redux.js.org/usage/migrating-to-modern-redux)
> Replace the existing manual Redux store setup with Redux Toolkit's `configureStore`

Changes:
- Install package `@reduxjs/toolkit`.
- Rewrite store setup using `configureStore`.
- Update `redux` and `@redux-devtools` packages to their latest versions to resolve a warning:
> Symbol.observable as defined by Redux and Redux DevTools do not match. This could cause your app to behave differently if the DevTools are not loaded. Consider polyfilling Symbol.observable before Redux is imported or avoid polyfilling Symbol.observable altogether.
- Allow RTK to handle syncing the store with the browser DevTools extension, and only use the `DockMonitor` if the browser extension is not installed.  Using both [causes](https://github.com/zalmoxisus/redux-devtools-extension/issues/188) [problems](https://github.com/nicksp/redux-webpack-es6-boilerplate/issues/5).
- Change checks of [deprecated](https://github.com/zalmoxisus/redux-devtools-extension/issues/220) `window.devToolsExtension` to current `window.__REDUX_DEVTOOLS_EXTENSION__`.
- Remove `reselect` as a top-level dependency because it is bundled with RTK.
- Change imports of `createSelector` from `reselect` to `@reduxjs/toolkit`.
- Create an instance of [`listenerMiddleware`](https://redux-toolkit.js.org/api/createListenerMiddleware) and link it to the store.  This is just a tool which can be used in future PRs and does not do anything right now.

Note: `redux-thunk` is also bundled with RTK, but it needs to stay as dependency because is used directly in some test files.

Questions:
- I kept the hot module replacement code in the store setup but I don't know if it's actually needed.  HMR has never worked for me so I cannot test it out.  This [answer from 2016](https://stackoverflow.com/a/34697765) shows the exact setup that we have, but the answers from [2019](https://stackoverflow.com/a/57924527) and [2020](https://stackoverflow.com/a/61942666) say it's not needed.
- I don't understand the use of `window.__INITIAL_STATE__` in our `index.jsx`.  It seems like it's used as a way to [pass state from the backend to the frontend](https://stackoverflow.com/questions/43739067/is-window-initial-state-still-the-preferred-way-to-pass-initial-state-to-the), but I don't see anywhere in the `server` part where we are setting this property.
- Now that we have support for the Redux DevTools browser extension, which happens automatically with `configureStore`, I'm not sure if we really need the manual setup of the `DockMonitor` on top of that?

Here's what it looks like in Edge, where I have no browser Redux DevTools, and the `DockMonitor` is shown:
image

In Chrome, the `DockMonitor` won't show up on top of the web app anymore but the browser Redux DevTools work!  Those tools show the Log but also a bunch of other stuff.
image
image
image
image


PiyushChandra17 (approved):
Nicely Done 👍

raclim:
Hey! Just wanted to ask here since this PR should definitely get in before the end of the year 😅 if you're okay with having your PR edited by others (like in the case of merge conflicts), of if you'd prefer to handle them yourself! (I'm not sure if there's an etiquette for this?)

lindapaiste:
> Hey! Just wanted to ask here since this PR should definitely get in before the end of the year 😅 if you're okay with having your PR edited by others (like in the case of merge conflicts), of if you'd prefer to handle them yourself! (I'm not sure if there's an etiquette for this?)

Hey I'm fine with you doing it if you understand the conflicts.  Like I definitely don't care if you regenerate a `package-lock.json` file.  There may be cases where it makes sense for me to do it if the conflicts are more complicated.

I'm looking at this now though and hopefully I can resolve this one.

Sometimes I have problems when I go to merge in the `develop` branch because the merge commit won't go through if it fails any tests.  My process in that case is:

- Remove `"pre-commit": "lint-staged"` from the `husky` section of `package.json` file (but don't commit this).
- Merge in `upstream/develop`.
- Fix issues so that it passes all tests and commit those changes (usually as an amend to the merge commit).
- Push.

raclim (approved):
Thanks for working through the conflicts! I'm going to quickly merge this in now then while it's good.

PR #1999, opened 2022-02-11
feat: configure webpack-dev-server for independent frontend development

hydrosquall:
## Motivation

- Enable people to contribute to `p5.js-web-editor` without needing to run a full node server + database
- Fixes #1348 
    - A redo of #1257 , given that Parcel doesn't support the range of customizations needed by the web editor. See [here](https://github.com/processing/p5.js-web-editor/pull/1257#issuecomment-1035708968) for a full breakdown of why I let go of Parcel.

## Changes

- Added `webpack-dev-server` dependency + upgraded `webpack-cli`
- Created a pair of `lodash.template` templates `index.html` and `preview.html` for the two primary routes needed for the app's functionality. Populates with the environment variables through the same `dotenv` configuration path as the server. Omits some of the variables that seemed unrelated to frontend development.
- Updated documentation for how to use the new dev server
- Manually verified screenshot: https://a.cl.ly/L1uWEq1l

## Testing

Try running this locally:

```bash
npm run start:frontend-server
```

## Checklist

* [x] has no linting errors (`npm run lint`)
* [x] is from a uniquely-named feature branch and has been rebased on top of the latest `develop` branch. (If I was asked to make more changes, I have made sure to rebase onto `develop` then too)
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #2207, opened 2023-05-07
feat: configure webpack-dev-server for independent frontend development [reboot of #1999]

### LINKED ISSUE #1348: [dev-task] Separate Webpack Dev server from API server

lindapaiste:
Fixes #1348 

**Credit for this code goes to @hydrosquall's PR #1999.**

Changes:

I merged the current `develop` branch into that branch and fixed the conflicts.

I also made some changes to how the vars are passed from the config to the `index.html` to ensure that we get `boolean` `true`/`false` values instead of string `'true'`. It's possible to inject the variables directly into the `<script>` with `window.process.env = <%= ENV_FROM_WEBPACK %>` but that is a syntax error if the file is accessed directly.  It seemed safer to place the variables into an HTML property and read from the HTML to the JS, so that's what I did.

This development environment can only load the IDE and cannot handle login, saving, or even viewing the p5 examples.  The next step will be to interact with with an API from the production site or a staging site.  I see that PR #1398 was merged in order to allow this, but I'm still getting CORS errors 🤷 

> Access to XMLHttpRequest at 'https://editor.p5js.org/editor/p5/projects' from origin 'http://localhost:8000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.


lindapaiste:
Hmm...hold off on merging this.  Turns out that it interferes with the normal `start` command as the `public/index.html` seems to take priority.

hydrosquall:
Thanks so much for looking into this and bringing this up to date with the latest code @lindapaiste !

> Turns out that it interferes with the normal start command as the public/index.html seems to take priority.

What if we put the dev server templates into a different folder, like `local-first-public/index.html` and `local-first-public/preview.html` ? We could update the two calls to `resolveAppPath`  in `config.local-dev-server.js` , leaving the main "public" folder clear for the usual start command.

Some people may want to use the existing dev server because they want to run the rest of the backend stack, but this "local-first" option should hopefully make life easier for those whose changes really don't depend on anything from the backend.

PR #2212, opened 2023-05-09
Cleanup unnecessary `catch` clauses

lindapaiste:
Changes:

- Added [rule `no-useless-catch`](https://eslint.org/docs/latest/rules/no-useless-catch) to `.eslintrc`
> A catch clause that only rethrows the original error is redundant, and has no effect on the runtime behavior of the program. These redundant clauses can be a source of confusion and code bloat, so it’s better to disallow these unnecessary catch clauses.

- Deleted lines of code that look like this:
```
try {
```
(code here unchanged, but indentation level is reduced)
```
} catch (err) {
  throw err;
}
```
- Or like this:
```
.catch((err) => {
  throw err;
});
```

Note:  All changed files are in the `/server/scripts` directory.  We don't have any redundant `catch` clauses in other parts of the codebase.


raclim (approved):
Thanks for working on this!

PR #2218, opened 2023-05-21
Enable hot reloading - Fixes #2134

### LINKED ISSUE #2134: Improper setup of obsolete package `react-hot-loader`

lindapaiste:
Fixes #2134

Changes:
- Replace deprecated package `react-hot-loader` with `react-refresh`.
- Integrate `react-refresh` with webpack using package `@pmmmwh/react-refresh-webpack-plugin`.
- Follow [this example setup](https://github.com/pmmmwh/react-refresh-webpack-plugin/tree/main/examples/webpack-hot-middleware).
- Replace `webpack-hot-middleware` with forked version `@gatsbyjs/webpack-hot-middleware`, which is up to date for webpack v5.
- Minor version bump of `react` and `react-dom` to the highest v16 version (to fit [these recommendations](https://github.com/pmmmwh/react-refresh-webpack-plugin/#getting-started)).

This PR improves the developer experience by allowing changes in the code to show up immediately without reloading the whole page.  React and redux state will stay active even when code is changed.

The only weird thing is that hot reloading messages from the embedded preview app will show up in the editor console. I don't think it's a big deal.
image


client/modules/User/pages/AccountView.jsx
```diff
@@ -48,7 +48,7 @@ function AccountView({ location }) {
   const { t } = useTranslation();
 
   const queryParams = parse(location.search);
-  const showError = !!queryParams.error;
+  const showError = !!queryParams.eryror;
```
raclim: I think this is a typo! 

I also recently had a conversation about how using double NOT operators can be confusing for new contributors because it obscures the data type and might make what's going on unclear. I know it's been implemented already in a few places within the codebase, but I was wondering if you had any thoughts on that!

raclim (commented):
Thanks so much for your work and sorry it took so long to get to this, this really does improve the developer experience! This looks great to me so far besides the portion that I commented on!

client/modules/User/pages/AccountView.jsx
lindapaiste: Omg thank you for catching this.  Yes it's a typo and there shouldn't be any change to this file at all.

I like `!!` because it's concise but I do see how it can be confusing.  `Boolean(queryParams.error)` is probably clearer.  I kinda think that the clearest would be to use `if (queryParams.error)` directly in an `if` or ternary statement and get rid of the derived boolean variable.

raclim: Hmm yeah, I agree that using an `if` or ternary might be the best alternative, but I do also like the brevity of `!!`. This might be something to think on or potentially flesh out in the developer docs down the line!

raclim (approved):
Looks great to me! I might be wrong, but I think the hot reloading messages disappear once some of the deprecation warnings are addressed.

PR #2220, opened 2023-05-22
Fix conflicts in `eslint` package versions (#2219)

### LINKED ISSUE #2219: Cannot run `npm install` due to dependency conflicts in eslint packages
### Increasing Access

Currently you cannot run `npm install` to set up the development environment.  You must run `npm install --legacy-peer-deps`.  This could be confusing to new developers and make it harder to contribute.

image

I'm creating this issue so that I can fix it.

`npm install` does not work in this repo right now due to a dependency conflict on the `eslint` package.  We are using `eslint` version `7.31.0`.  However we are using an outdated version of `eslint-config-airbnb` which expects a much older version of `eslint`.

We have `eslint-config-airbnb@^16.1.0` which expects a peer dependency of `eslint@^4.9.0`.  We can fix the conflict by updating the `eslint-config-airbnb` package to its latest version which is in the `19.x` range -- up 3 major versions.  This has not been done previously because updating that package [introduces lots of new rules](https://github.com/processing/p5.js-web-editor/issues/2217) and causes lots of code to fail linting checks.  I opened #2217 to discuss the specifics of what new rules we like and don't like.  In order to upgrade the package and fix the `npm install` conflict without changing lots of code, I recommend modifying the `.eslintrc` config to mimic the current rule conditions.  We can then tackle rules one-by-one.

lindapaiste:
Fixes #2219

Changes:
* Update `eslint-config-airbnb` to the latest version.
* Bump `eslint` from `7.31.0` to `7.32.0`.
* Turn off (set to `0`) rules which I think that we don't want to keep.
* Downgrade to a warning (set to `1`) rules which I think that we should keep.
* Fixed a few rule violations that we only had in one place, so I didn't think that they warranted separate PRs:
   * Deleted unreachable `else` branch in `client/modules/IDE/pages/MobileIDEView.jsx` (violated `no-dupe-else-if`)
   * Don't assign properties on imported object `FileActions` in `client/modules/IDE/components/FileNode.jsx` (violated `no-import-assign`)
   * Changed `export { ConnectedFileNode as default };` to `export default ConnectedFileNode;` in `client/modules/IDE/components/FileNode.jsx` (violated `no-restricted-exports`)
   * Added some curly braces in `client/modules/User/actions.js` so that the inner function of the `Promise` doesn't return anything (violated `no-promise-executor-return`)
   * ~~Changed `if (module.hot)` to `if (import.meta.webpackHot)` in `client/store.js` [as recommended by webpack](https://webpack.js.org/api/hot-module-replacement/) for strict ESM compliance. (violated `import/no-import-module-exports`)~~ I had to roll this back due to [issues with Jest](https://stackoverflow.com/questions/64961387/how-to-use-import-meta-when-testing-with-jest).
  * Added missing prop types to `RedirectToUser`. (violated `react/prop-types`)

There are no linting errors but there are 64 warnings.


lindapaiste:
@raclim if we want to fix the conflicts on this so that it can be merged just let me know.  The import part is the `.eslintrc` and the two version bumps in the `package.json`.  There may be conflicts (currently and in the future) regarding the rule violations which I chose to fix rather than adding exceptions for, as a lot of those mistakes have been cleaned up in other PRs.

raclim:
Thanks for raising this, that makes sense to me! I think I might push the next release date to Jan 16th to get a few more PRs in, so I'm going to add this to the current milestone.

PR #2226, opened 2023-06-03
Remove titles from SVG files - Fixes #2225

### LINKED ISSUE #2225: Unhelpful tooltip "Triangle"
### Increasing Access

We want tooltips only when they are descriptive and helpful.

tooltip-triangle

When hovering the dropdown icon next to a file name, you see a tooltip that says "Triangle".  This comes from a `<title>` attribute inside the .svg file.

https://github.com/processing/p5.js-web-editor/blob/7212f76f6261ad52a33c9c87e17e5d90be903b3c/client/images/down-filled-triangle.svg#L4

-----

In other cases, we have `<title>` attributes in our SVG files which _are_ helpful...in English.  They will not get translated with the rest of the page.  We should remove all `<title>` attributes inside the SVG image file and instead use `title=` properties in React components, which can be properly translated with i18n.
tooltip-stop

lindapaiste:
Fixes #2225 

Changes:

- Primary purpose of this PR:
  - Removed `<title>` tags from all svg files.
- Changes to components:
  - Copied the translated `aria-label` text to a `title` attribute for a few icon buttons: play, stop, preferences, folder open/close.  This will show up in a tooltip when the button is hovered. I'm not sure if we actually want this?
- Additional cleanup of SVG files:
  - Removed software tags like `<desc>Created with Sketch.</desc>` and `<!-- Generator: Sketch 47 (45396) - http://www.bohemiancoding.com/sketch -->`.
  - Removed empty `<defs></defs>`.
  - Collapsed multiple group (`<g>`) tags into one for some files.
  - For triangles, I removed all of the nested `transform`s and wrote the three points relative to to root viewbox.
  - In some cases I removed the hard-coded color and replaced it with `currentColor` since we override the color via CSS anyway.  I can roll this back if we want to keep the intrinsic color.  It only matters for files which we load using the `?byUrl` approach, as those will have the color that is specified in the .svg file.

You'll understand the differences in svgs better if you click the <> button to "Display the source diff".


raclim (approved):
Thanks for your work on these changes!

PR #2210, opened 2023-05-09
Cleanup language menu code in `Nav` component

lindapaiste:
Summary:
We don't need to manually specify all of the languages to include in the `Nav` menu because we already have all of the languages and their names in the `i18n.js` file!  This PR removes over 100 lines of code and makes it easier to add new translations.

Changes:
- Generate the list items in the `Nav` component by looping over the `availableLanguages` array.
- Remove the portion of the `translations.md` doc about editing `Nav.jsx` as this is no longer necessary.
- Added some info to the `translations.md` doc about choosing language codes and tweaked the section about editing the `i18n.js` file.

I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] has no test errors (`npm run test`)
* [x] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* [ ] is descriptively named and links to an issue number, i.e. `Fixes #123`

Issue #2150, opened 2023-03-06
Hover on navbar Links shows "main-navigation", "user-menu" and the drop down arrow shows "triangle"


Aryan1982:
![m5](https://user-images.githubusercontent.com/62349184/223084140-af0cf910-9a5d-4b97-aa9c-daaaa85467ff.png)
![m1](https://user-images.githubusercontent.com/62349184/223084147-3457da2d-bf13-47bd-aee7-7c9e43356b47.png)
![m2](https://user-images.githubusercontent.com/62349184/223084150-21aecf72-0674-4124-b23d-09c2a2d09b4f.png)
![m3](https://user-images.githubusercontent.com/62349184/223084153-7d037174-0a92-49b8-9edf-e08f67c4ad34.png)
![m4](https://user-images.githubusercontent.com/62349184/223084160-8e623a27-bdf9-49ed-80f0-ad623b18f686.png)

Also hovering on the P5.js logo at the top corner shows "main-navigation"
If its a valid issue can I work on it?

Issue #1266, opened 2020-02-17
Project menu tooltip


shakti2379:
#### Nature of issue?
- Feature Enhancement

#### Details about the bug: 

- Web browser and version: <!-- On Chrome 80
- Operating System: Linux

#### Feature enhancement details:
click on the project menu, hover on any item it will show project menu as tooltip, which is redundant, i think it should either show some more details about the item in tooltip or no need of tooltip there

shakti2379:
if it looks fine, would like to work on it:)

ashu8912:
@shakti2379 I have already made a pr regarding this that pr will fix every svg tooltip realated problem.

ashu8912:
#1259

catarak:
thanks for reporting! you are right that this `title` should be removed. See https://developer.paciellogroup.com/blog/2012/01/html5-accessibility-chops-title-attribute-use-and-abuse/!

davidbmx:
Hi @catarak i create a pr for this issue ref #1382

PR #2227, opened 2023-06-04
Refactor Nav pt. 2 - create sub-components

### LINKED ISSUE #2224: Refactor/breakup `Nav` component
### Increasing Access

- It will help @dewanshDT in his GSoC project...
- ... which will help make the editor mobile responsive and more accessible to everyone!
- It will also help future contributors who work on this component.

The [`Nav` component](https://github.com/processing/p5.js-web-editor/blob/7212f76f6261ad52a33c9c87e17e5d90be903b3c/client/components/Nav.jsx) is over 1,000 lines long and has a lot going on.  I want @dewanshDT to start working on making this UI mobile-friendly, but the component is very unwieldy.  I would like to refactor some of the logic first and then he will have an easier time working on the styling.

### Breakup
I suggest that we create individual components for various parts of the UI, such as `NavBar`, `DropdownMenu`, `MenuItem`, etc.

### Defining Lists
We can define the contents of the menu as a tree of JSX elements, like we are doing now. That would look something like this:
```
<NavBar>
  <DropdownMenu key="file" title="File">
    <MenuItem onClick={handleNew}>New</MenuItem>
    <MenuItem href="/p5/sketches">Examples</MenuItem>
  </DropdownMenu>
</NavBar>
```
We could also use a structure of JavaScript objects and arrays.  This was [suggested back in 2019](https://github.com/processing/p5.js-web-editor/issues/879#issuecomment-472156978) and is the way that it's [handled in the unused `MobileIDEView`](https://github.com/processing/p5.js-web-editor/blob/7212f76f6261ad52a33c9c87e17e5d90be903b3c/client/modules/IDE/pages/MobileIDEView.jsx#L80-L136).  That would look like:
```
const navLinks = [
  {
    key: 'file',
    title: 'File',
    subMenu: [
      {
        key: 'new',
        title: 'New',
        onClick: handleNew
      },
      {
        key: 'examples',
        title: 'Examples',
        href: '/p5/sketches'
      }
    ]
  }
];
```
My personal opinion is that the JSX is better as it's more contributor-friendly.  Right now the drawback of the JSX is that we are passing a lot of the same props to each element, but that can be cleaned up.

### Event Handling
We are passing down `onBlur`, `onFocus`, and `onClick` handlers to each item in the dropdown menus.  My goal is to get rid of these props but be still be able to trigger changes in menu state from child components.

I'm thinking that React Contexts is probably the best way, though injecting props with [`React.Children.map()` and `React.cloneElement()`](https://github.com/search?q=repo%3Amui%2Fmaterial-ui%20react.children.map&type=code) is also an option.

A `MenuItem` would be aware of which `DropdownMenu` is its parent. It would have access to handler functions from some sort of context provided by the parent `NavBar`.  That way the `MenuItem` can deal with the event handling internally and the component itself doesn't need to take a lot of props.

lindapaiste:
Fixes #2224 

Extends PR #2210 

Changes:

This PR should not change any UI or behavior, with one tiny exception: I removed a few `title=` tags which were causing unhelpful tooltips (see also #2150, #1266).

I'm changing how the UI is generated by breaking it up into various building blocks that can be put together to form the current `Nav`.  I'm cutting pieces of code out of `Nav` and pasting them into new files.  In the best cases, I'm cutting code that's repeated in multiple spots and pasting it only to one usage. 

#### `NavBar`

The `NavBar` is the parent that wraps all of the elements in the navigation.
- Controls the state of which dropdown is currently open.
- Contains all existing timeout logic from `Nav`.
- Registers event listeners for close on `Esc` and close on outside click behavior.
- Generates event handlers for top-level dropdowns and individual menu items, and makes them available via context.
- Provides a context of which menu is open.
- Renders the `header` and `nav` elements.

#### `NavDropdownMenu`

The `NavDropdownMenu` is a top-level item in the nav bar which contains a submenu (File, Edit, etc.).
- Gets its event handlers via context.
- Knows if it is open via context.
- Provides its own `id` to its children via context.
- Renders the UI for the button including the title and the triangle icon, and a dropdown list containing its children.

#### `NavMenuItem`

The `NavMenuItem` is for each item a dropdown menu.
- Gets its event handlers via context.
- Wraps in an `li` with the correct `className` (likely to be changed to a styled component in the future).
- Has a prop `hideIf` for cleaner conditional rendering.
- Basically a wrapper around `ButtonOrLink`.

#### `ButtonOrLink`

I put this in the `common` folder because it can be used in other places.  `ButtonOrLink` is a helper that allows buttons and links to be used interchangeably.
- Renders a `button` when there is no `href` prop.
- If the `href` is an external link, it renders an `a` and automatically adds `target="_blank" rel="noopener noreferrer"`.
- If the `href` is an internal link, it renders a `react-router` `Link`.

## Changes to `Nav`

- For each dropdown menu, delete a block of code that looks like this:
```
<li className={navDropdownState.file}>
  <button
    onClick={this.toggleDropdownForFile}
    onBlur={this.handleBlur}
    onFocus={this.clearHideTimeout}
    title="File"
    onMouseOver={() => {
      if (this.state.dropdownOpen !== 'none') {
        this.setDropdown('file');
      }
    }}
  >
    <span className="nav__item-header">
      {this.props.t('Nav.File.Title')}
    </span>
    <TriangleIcon
      className="nav__item-header-triangle"
      focusable="false"
      aria-hidden="true"
    />
  </button>
  <ul className="nav__dropdown">
```
and replace with:
```
<NavDropdownMenu id="file" title={this.props.t('Nav.File.Title')}>
```

- For each menu item, delete a block of code that looks like this:
```
{this.props.project.id && (
  <li className="nav__dropdown-item">
    <button
      onClick={this.handleShare}
      onFocus={this.handleFocusForFile}
      onBlur={this.handleBlur}
    >
      {this.props.t('Nav.File.Share')}
    </button>
  </li>
)}
```
and replace with:
```
<NavMenuItem
  hideIf={!this.props.project.id}
  onClick={this.handleShare}
>
  {this.props.t('Nav.File.Share')}
</NavMenuItem>
```

- Move the click-outside and keydown handling into `NavBar`.
- Move the `dropdownOpen` state into `NavBar`, along with associated logic from `toggleDropdown`, `handleFocus`, `handleBlur`, `clearHideTimeout`, etc.
- Delete the `this.setDropdown('none');` at the end of each `handle` function because this is now handled by the `NavMenuItem` (on the `onMouseUp` event, so that it doesn't conflict with `onClick` props).
- For functions which are now trivial, delete the handler and move the function inline. ie. `onClick={this.props.stopSketch}` instead of `onClick={this.handleStop}`.
- Delete so many function bindings which aren't needed anymore.

The `Nav` component can continue to be broken up and cleaned up.  It's easy to convert it to a function component now because all of the logic has been moved elsewhere.  But I wanted to put in this PR while the general structure of `Nav` is the same so that it's easier to see and review the changes.


raclim (approved):
Thanks so much for breaking this down and explaining your updates so thoroughly! I think splitting it into those proposed sub-components makes sense and is a great update to the `Nav` component. I also wasn't too familiar with React Contexts until this PR, and think it's pretty neat!

PR #2232, opened 2023-06-04
Feature: Preview uploaded files in the editor

### LINKED ISSUE #57: Different view for media files in editor
Right now, for S3 hosted files, the editor just silently fails. It would be cool if info could be displayed about the files, but for now, just an error that these files can't be edited.

lindapaiste:
Fixes #57

Changes:
- Creates an `AssetPreview` component which is rendered inside the editor container when the user clicks on a file with a `url` (that is, an asset uploaded to S3).
- Gets the mime type from the file name using package `mime` (which we use on the server).
- Handles the following mime types:
  - `image/*` => `` element
  - `video/*` => `<video>` element
  - `audio/*` => `<audio>` element

I might need to tweak the CSS for images which are taller than they are wide.  I'm using ` max-width: 100%;  height: auto;` for both image and video.

image
image
image


raclim (approved):
This is really neat, thanks so much for working on this!

PR #1873, opened 2021-05-23
Autocomplete and Hinter

peilingjiang:
Resolves #99

js-hinter

I was looking for WIPs and issues regarding autocompletion and found #99. Although the issue has been idle for a while, I think this feature can greatly help beginners to start and help reduce typos and increase efficiency in general. It turns out that CodeMirror provides a great base for implementing it so I tried to use its hinter addon and build this PR.

In addition to autocompletion, the hinters also have links to the corresponding documentation page. Not exactly the same as requested in #99 but provides a similar reference capability.

Some more details of this implementation:
- The style has been configured to work in all three themes.
- **TBD** It is triggered every time an alphabet is typed, unlike the native CodeMirror hinter that will only be triggered by `Ctrl-Space`.
- **TBD** It only works in `css` and `javascript` modes, as CodeMirror's HTML hinter works weirdly.
- **TBD** For JavaScript, it can only hint and autocomplete p5 keywords, unlike fancy IDEs that can also complete JS and customized keywords.
- I added a lightweight new dependency, [fuse.js](https://fusejs.io/), to tolerate some misspellings, e.g., you can still get an option of `circle` even when you type `cic` as a start.
- It depends on a new utils file - p5-hinter.js - that can be updated by update-p5-hinter.js (just like update-syntax-highlighting.js). It fetches the same data but organizes it in a different way.

I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] is from a uniquely-named feature branch and has been rebased on top of the latest `develop` branch. (If I was asked to make more changes, I have made sure to rebase onto `develop` then too)
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #2236, opened 2023-06-08
Autocomplete++

### LINKED ISSUE #99: p5.js library Autocomplete and/or link to specific parts of reference
Students requested a link to the p5.js reference in the editor.   Ideally we would also eventually have an option where you could highlight a function name like `fill`, right-click, and have a "find in reference" option.

peilingjiang:
Fixes #99 and extends #1873

Changes: Extending the current dropdown hinter solution by adding a Copilot-style inline completion preview, which includes argument previews for the functions.

Screenshot 2023-06-08 at 22 42 52


peilingjiang:
Another update that I hope would boost utility - reserved keywords and namespace objects in JavaScript -

Screenshot 2023-06-08 at 23 59 54

raclim (approved):
Thanks for these updates, I think the altogether the autocomplete, argument previews, and link to references make this really helpful and amazing!

Atcold:
Wow, impressive, @peilingjiang!
I \*really\* like the computational literacy idea behind P5.js, but the lack of autocompletion and hints in the online editor made the whole thing still rather hard for newcomers.
I truly believe your contribution is a major step forward towards easing the getting-started process!
Thank you \*very\* much!

Side question (perhaps for @raclim).
How do I enable these functionalities?
I see the PR has been merged into the `develop` branch.
How does a user check it out for the online editor?
Thanks!

raclim:
@Atcold The PR has been merged, but we haven't pushed it out live to production yet due to some bug fixes and additional testing we'd like to perform before distributing it to everyone, which we're planning to complete by the end of June. I'm hoping to provide a more consistent way to communicate when these releases will happen, but for now I'll definitely send an update when it's out live!

raclim:
Hi, just wanted to update that this is out live on production! :) 

I made one minor change, where I set the initial preference for the Autocomplete and Hinter to 'Off' instead of 'On', since I felt this feature might be a little confusing when a new user is encountering the editor for the very first time. However, please let me know if you feel like this should be changed or have a different suggestion, thanks again for all of your work/input on this!

asukaminato0721:
> or have a different suggestion

I have some thoughts:

for those who have an account and some works, they may be familiar with the online editor. Then they may not be so confusing. So making this option on for them is a boost.

For newbies: since now there are a lot of web editors based on https://github.com/microsoft/monaco-editor, and it has auto-complete out of the box. (See https://swc.rs/playground for example). I guess even a newbie can handle it.

Atcold:
I agree with @wuyudi.

New users would benefit from having suggestions out of the box.
They may not know such a utility exists and that it can be activated.
So, setting it initially on would help feature discoverability.

On the other side, if one dislikes the functionality (for whatever reason), one can look for ways to stop it.

In short, one doesn't know one doesn't know.
Hence, awareness should be prioritised.

PR #1164, opened 2019-09-20
Sketch collections

andrewn:
Implementing features in #1115.

## Editor API for Collections

|                Route                 |               Purpose               |            Restrictions            |
| ------------------------------------ | ----------------------------------- | ---------------------------------- |
| `GET /collections`                   | Fetch current user's collections    | Logged in user                     |
| `GET /:username/collections`         | Fetch a specific user's collections |                                    |
| `POST /collections`                  | Create a new collection             | Logged in user                     |
| `PATCH /collections/:id`             | Update collection metadata          | Logged in user must own collection |
| `DELETE /collections/:id`            | Delete collection                   | Logged in user must own collection |
| `POST /collections/:id/:projectId`   | Add a project to a collection       | Logged in user must own collection |
| `DELETE /collections/:id/:projectId` | Remove a project from a collection  | Logged in user must own collection |

## Features

- [x] Editor API to CRUD collections
- [x] Lists of your own and other user's collections (`/:username/collections`)
- [x] View any user's collection (`/:username/collections/:collectionId`)
- [x] Create your own collections (`/:username/collections/create`)
- [x] Edit your collection's name and description (Click name and description on collection page)
- [x] Easily get a shareable URL for the collection ("Share" link on collection page)
- [ ] See and edit the collection slug
- [x] Delete your own collections (via dropdown on collection list)
- [x] Add any sketch to one of your collections (via File → Add to collection)
- [x] Add your own sketches to one of your collections (`/:username/collections/:collectionId` →
"Add more sketches")

## Checks

I have verified that this pull request:

* [ ] has no linting errors (`npm run lint`)
* [x] is from a uniquely-named feature branch and has been rebased on top of the latest master. (If I was asked to make more changes, I have made sure to rebase onto master then too)
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #2238, opened 2023-06-10
Use `EditableInput` component in `Toolbar`

### LINKED ISSUE #2237: Project name looks clickable when viewing others' projects
### Increasing Access

It makes the UI confusing when elements have a hover effect but don't do anything when clicked.

Steps:
1. Go to https://editor.p5js.org/p5/sketches/3D:_basic_shader
2. Hover over the title "3D: basic shader" in the toolbar

The title appears pink when hovered and shows the `cursor: pointer`👆 instead of the arrow.  It looks like you can click it -- but clicking it doesn't do anything.  We use the same hover effect on the user's own project, where clicking the title turns it into an editable input.  The title is not editable on other people's projects, so the title should not show any interaction when hovered.

image

We can fix it by adding `pointer-events: none` to the CSS.  We can probably apply that sitewide to all disabled buttons?  I'm always nervous to change anything globally which could have unintended consequences.  So it may be better to add that CSS just to this one button.
```
button:disabled {
    pointer-events: none;
}
```

lindapaiste:
Fixes #2237 
(I can put in a PR with a 3-line CSS change if we want to fix *just* that issue and not all this other stuff)

Background:
We have a nice `EditableInput` UI component which was created by @andrewn for the collection details page in #1164.  This component displays a text with a pencil icon and converts it to an input when clicked.  We use this same behavior when displaying the project name in the editor toolbar but we aren't using this component for it.  We should use it!

Changes:
- `EditableInput` component:
  - Can override the `aria-label` via props (so we can still use `t('Toolbar.EditSketchARIA')`).
  - Can pass a prop `disabled` to render as plain text with no pencil icon and no hover effects (issue #2237)
  - Added support for canceling by pressing the `Esc` key.  This will discard any text typed into the input and revert to the initial value, without submitting or saving anything.
  - Sets an `aria-hidden` property on the label and the input when they are hidden via CSS `display: none`
- `Toolbar` component:
  - Use the `EditableInput` component.
  - Delete a bunch of code which duplicates what's already handed by the `EditableInput`.
  - Delete most of the CSS for the `.toolbar__project-name`, but leave behind a few overrides which apply to the `EditableInput` when used inside the `.toolbar` to keep the style the same as the current (thinner border, uses `'secondary-text-color'` instead of `'inactive-text-color'`). 
- `isEditing` state management:
  - `isEditing` is now a local component state of the `EditableInput` and does not need to be in Redux.
  - Delete Redux actions `showEditProjectName` and `hideEditProjectName` and remove `isEditingName` from state.
  - The `Toolbar.unit.test.jsx` file can no longer see if these actions have been dispatched, so instead I'm looking at the `disabled` and `focused` states of the `<input>` element to determine whether or not it is editing.


dewanshDT (approved):

lindapaiste:
@dewanshDT I merged your `Toolbar` changes into this.  Now see if you can merge this branch into your `MobileNav` and use the `ProjectName` component when displaying the name in the nav bar so that we have the ability to edit project names on mobile.

dewanshDT:
Okay

dewanshDT:
Hey @lindapaiste, I think we do not want to directly use the ProjectName component in the MobileNav because it doesn't have a very mobile feel to it, instead, we can open a modal to edit the name and use the ProjectName component that modal.
Let me know what you think

Issue #2174, opened 2023-03-21
Avoid direct modification of HTML


lindapaiste:
Wherever possible, the content of the page should be controlled through React using `props` and `state`.  Modifying the content using pure JS should be avoided.

The example that I found which is easy to fix is in the `Editor` component:
https://github.com/processing/p5.js-web-editor/blob/e01efe85ec566ace6ee0980be3ea573a08406904/client/modules/IDE/components/Editor.jsx#L180-L185

The line number should be stored as a state and then passed down as a prop to the `EditorAccessibility` component, which contains the `'current-line'` element.

It also seems like there is some wonky stuff going on [throughout the `uploader`](https://github.com/processing/p5.js-web-editor/blob/dd433b768da0a08dfda547ffc37a5624c37de4f3/client/modules/IDE/actions/uploader.js), but I need to dig in there to see what the code is doing.

https://github.com/processing/p5.js-web-editor/blob/dd433b768da0a08dfda547ffc37a5624c37de4f3/client/modules/IDE/actions/uploader.js#L111

https://github.com/processing/p5.js-web-editor/blob/dd433b768da0a08dfda547ffc37a5624c37de4f3/client/modules/IDE/actions/uploader.js#L121-L122

https://github.com/processing/p5.js-web-editor/blob/dd433b768da0a08dfda547ffc37a5624c37de4f3/client/modules/IDE/actions/uploader.js#L53-L56

https://github.com/processing/p5.js-web-editor/blob/dd433b768da0a08dfda547ffc37a5624c37de4f3/client/modules/IDE/actions/uploader.js#L76

lindapaiste:
I found another one in the `CopyableInput`.  We are mutating the DOM to add and remove class names on an element which is actually rendered by React. The component should have a `state` (like `tooltipActive` or whatever) and render the element with different class names depending on that state.

https://github.com/processing/p5.js-web-editor/blob/7212f76f6261ad52a33c9c87e17e5d90be903b3c/client/modules/IDE/components/CopyableInput.jsx#L15-L33

PR #2241, opened 2023-06-11
Fix #2040 - Inaccurate rendering during file upload

### LINKED ISSUE #2040: Inaccurate rendering during file upload
Animation is not rendering properly when uploading multiple files of three or more.

#### Details about the bug:

- p5.js version: 1.4.0
- Web browser and version: Google Chrome 103.0.5060.114
- Operating System: Windows
- Steps to reproduce this:
1. uploading multiple files
2. only the top two upload completion animations work
(All files are uploaded successfully)

|Successfully uploaded|Only two animations|
|:---:|:---:|
|image|image|

lindapaiste:
Fixes #2040 

The bug is caused by these lines (which I actually called out in #2174):
https://github.com/processing/p5.js-web-editor/blob/1c05241e9ead4bc6f622188ba205c3329268fc72/client/modules/IDE/actions/uploader.js#L111
https://github.com/processing/p5.js-web-editor/blob/1c05241e9ead4bc6f622188ba205c3329268fc72/client/modules/IDE/actions/uploader.js#L121-L122

Our intention is to insert an additional `input` into the form.  But the way that we are implementing it is by replacing the entire content of the uploader with new elements parsed from the modified HTML string. 

The Dropzone library stores a reference to the `previewElement` associated with each file and modifies the CSS classes on this element at various points in the code.  When our bad code gets executed for the first file all of the `file.previewElement` elements get removed from the DOM.  Any subsequent changes to these elements will not change the HTML that we actually see.

As a result, only the first file will get the "success" styles applied.  The second file gets stuck on "processing" and the rest of the files never even get to the "processing" style.

Changes:
I removed the modification of `innerHTML`.  Instead I am creating the input with `document.createElement` and inserting it with `appendChild`.  This keeps the DOM intact.

I had to dig way deep into this code in order to track down the cause of the bug and now that I understand it I do have some further improvements which I can make in future PRs.  ~~I think it would be cleaner to render these hidden inputs via React in the FileUploader component, by storing the data to component state.~~ I think we might not need the hidden inputs at all?

### DIFF
--- a/client/modules/IDE/actions/uploader.js
+++ b/client/modules/IDE/actions/uploader.js
@@ -73,7 +73,6 @@ export function dropzoneAcceptCallback(userId, file, done) {
           file.custom_status = 'ready'; // eslint-disable-line
           file.postData = response.data; // eslint-disable-line
           file.s3 = response.data.key; // eslint-disable-line
-          file.previewTemplate.className += ' uploading'; // eslint-disable-line
           done();
         })
         .catch((error) => {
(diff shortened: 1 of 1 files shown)


lindapaiste:
This is no longer needed after #2244, where we removed the hidden inputs entirely.

PR #2244, opened 2023-06-12
Cleanup of uploader code

lindapaiste:
Changes:
- Convert the `FileUploader` to a function component.
- Call the `.destroy()` method when the component unmounts.  ~~This prevents the form HTML from getting added to the page multiple times when the modal is closed and re-opened.~~ The duplicated HTML is no longer an issue after #2245.  Still seems like a good idea to cleanup.
- Move some (but not all) styling from CSS to JS.
- The `accept` and `sending` callbacks don't involve Redux at all so I think it's weird to `dispatch` them (though it does work just fine).  I'm calling them as plain functions instead.
- Turn the repeated `if` condition into a utility function `isS3Upload`.
- Delete the `localIntercept` function which uses `FileReader`.  We can use `.text()` method on the `file` object which is the [newer approach](https://developer.mozilla.org/en-US/docs/Web/API/Blob/text#usage_notes).
- Convert the `accept` function syntax from `Promise.then` to `async`/`await`.
- Remove the complicated `if` condition from the error branch with the magic of the [optional chaining `?.`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)
- Delete the hidden inputs which set the `attachments[]` field on the form.  I can't see why we would need this and it works without it.  The Dropzone already includes the binary data in the form under the `file` property. Pinging @catarak - maybe she knows why it's there.  It's possible that it used to be necessary but isn't anymore.

### DIFF
--- a/client/modules/IDE/actions/uploader.js
+++ b/client/modules/IDE/actions/uploader.js
@@ -1,126 +1,68 @@
+import { TEXT_FILE_REGEX } from '../../../../server/utils/fileUtils';
 import apiClient from '../../../utils/apiClient';
 import getConfig from '../../../utils/getConfig';
 import { handleCreateFile } from './files';
-import { TEXT_FILE_REGEX } from '../../../../server/utils/fileUtils';
 
-const s3BucketHttps =
+export const s3BucketHttps =
   getConfig('S3_BUCKET_URL_BASE') ||
   `https://s3-${getConfig('AWS_REGION')}.amazonaws.com/${getConfig(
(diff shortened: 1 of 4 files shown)


raclim (approved):

raclim:
Thanks so much for working on this! I'm not too sure what might've been happening with the hidden inputs either, but I also think it should be fine without it? 

I'll merge in #2245 right after as well, and then do the next release :)

PR #2252, opened 2023-06-17
Fixes #2250 - Don't show loader after the collection has been loaded.

### LINKED ISSUE #2250: Collection details disappears for a moment when clicking "Add Sketch"
### p5.js version

_No response_

None

### Web browser and version

_No response_

### Actual Behavior

The collection name, description, etc. disappear for a moment and then come back when the "Add Sketch" modal is opened.

### Expected Behavior

Opening the modal should not change the content behind it.

### Steps:
1. Go to the collection details page for a collection that you own.
2. Click "Add Sketch".

___

This is a result of having a single `isLoading` state for all API calls.  While we are fetching the list of sketches, the code acts as if it is fetching the collection itself -- even though that API call already completed successfully.

It can be fixed by modifying some of these conditionals:
https://github.com/processing/p5.js-web-editor/blob/1c05241e9ead4bc6f622188ba205c3329268fc72/client/modules/User/components/Collection.jsx#L216-L227

There is no possibility that `this.props.collection` would be the wrong collection (ie. stale state from a previous collection view), as the selector looks specifically for the collection matching the provided `collectionId`.
https://github.com/processing/p5.js-web-editor/blob/1c05241e9ead4bc6f622188ba205c3329268fc72/client/modules/User/components/Collection.jsx#L531

lindapaiste:
Fixes #2250 

Changes:
- Change the `defaultProps` for `collection` from an incomplete/fake object to `null`.  This way it's easy to know when the collection has been loaded.
- `<Loader/>` component won't be displayed after `this.props.collection` has been set, regardless of the value of `this.props.loading`.
- `hasCollection` only looks to see if it has a valid collection prop and does not consider `this.props.loading`.  Consequently the collection details will stay visible while the popup content is loading.
- The empty table will be displayed whenever there is a collection but it has no items.  Loading is not considered.  The change in the `defaultProps` makes this check possible.


raclim (approved):
Thanks for catching and working on this!

PR #2295, opened 2023-07-15
Upgrade react-router from v4 to v5 and begin using hooks

lindapaiste:
Related issue #857

This is step 2 and should be merged after #2294 

Upgrading `react-router` from v4 to v5 is a [non-breaking change](https://reacttraining.com/blog/react-router-v5) so all that is strictly required is changing the package versions.

[v5.1.0](https://reacttraining.com/blog/react-router-v5-1) introduces React hooks `useLocation`, `useParams`, `useHistory`, etc.  These are now the preferred way to access `location`, `params`, and `history` rather than via props (or via a global variable, in the case of `history`).

As v5.x is backwards-compatible, existing code using `this.props.location` will still work.  I have not made changes to any class components in this PR.  I have only used the hooks in existing function components where it is trivial to do so.

Note: In v6 the hooks will be the _only_ way to access the router state.  Though we can create our own backwards-compatibility for class components by creating HOCs.


raclim (approved):
Looks good to me, thanks!

PR #2302, opened 2023-07-15
Delete unused code in `CollectionView`

### LINKED ISSUE #2300: Legacy code for an unreachable collection /create page
### Increasing Access

Cleans up the codebase.

In its current state, the [`CollectionView` component](https://github.com/processing/p5.js-web-editor/blob/4161f98c2127bbead3e317c699694589add135ce/client/modules/User/pages/CollectionView.jsx) doesn't do much of anything.  There are two internal methods which are never called and a boolean check `isCreatePage` which will always be `false`.

The `CollectionView` is the route handler for the path `"/:username/collections/:collection_id"`.  It switches between displaying a single `Collection` and displaying the `CollectionCreate` UI based on whether the `collection_id` portion of the URL is the word `"create"` rather than a valid ID.

**But the `"/:username/collections/create"` URL is impossible to reach**, as [the server only checks for a valid collection ID](https://github.com/processing/p5.js-web-editor/blob/4161f98c2127bbead3e317c699694589add135ce/server/routes/server.routes.js#L124-L128) and will return a 404 page.

My plan is to delete the useless code.  However we could make the `"/:username/collections/create"` URL work if we wanted to.  Right now the UI for creating a collection is a modal that pops up over the `"/:username/collections"` page.  There is no dedicated URL for this.

image

If the server code is changed, this is what would be displayed:
image

lindapaiste:
Fixes #2300

Should be reviewed/merged after #2295 because I using the `useParams` hook from react-router v5.1+

Changes:

- Delete legacy code which doesn't do anything
- Convert from a class component to a function component
- No changes to behavior


raclim (approved):
Thanks for working on this, I think this makes sense to me so far!

PR #2307, opened 2023-07-16
Convert `APIKeyForm` to a function component and connect to Redux

lindapaiste:
Fixes a `//TODO` and also fixes a mistake that I made in the existing code for the `AccountView`.  I should be passing through arguments to the action creators on those `dispatch` calls.  As a result of that mistake, the `ApiKeyForm` doesn't work. Which is fine because the feature flag `UI_ACCESS_TOKEN_ENABLED` is `false` so we never show the form.  But we should fix it.

https://github.com/processing/p5.js-web-editor/blob/3493b3b8099d42cac7220f0181ba4fbafebddbf0/client/modules/User/pages/AccountView.jsx#L105-L111

Changes:
* Remove some API key logic from the parent `AccountView`.
* Convert `APIKeyForm` to a function component.
* `APIKeyForm` manages its own Redux state with `useSelector` and `useDispatch`.
It looks like a lot more changes when viewing the diff due to the changed indentation level.


PiyushChandra17 (approved):
Nice Done

raclim (approved):
Going to get this one in too! Thanks so much for noting and addressing this so quickly at the time!

PR #2310, opened 2023-07-17
fix(#1216): storybook config

ofhope:
Related to #1216 

It looks like Storybook has been set up previously but is currently in a broken state. 

I was attempting to explore the project via Storybook and notice the commands weren't working.

Screenshot 2023-07-17 at 9 02 24 pm

The changes in `package.json` were due to running this command `npx storybook@latest init` which took me through an install wizard. These are the packages Storybook recommends.

`addDecorator`, `addParameters` and `withKnobs` are no longer supported. Controls are created for you through prop introspection with the ability to override with `argTypes`.

I'm not familiar with the work prior to these changes so I can't offer garuntee that thats how these stories behaved prior to this upgrade.

But I can confirm all stories are now rendering.

Screenshot 2023-07-17 at 5 40 23 pm

Screenshot 2023-07-17 at 5 34 44 pm

Screenshot 2023-07-17 at 5 34 41 pm

Screenshot 2023-07-17 at 5 34 38 pm

I notice one of the initial motivations were "help reduce bugs and make contributing easier" which I absolutely agree with. It makes debugging hard to reach aspect a breeze. It also helps seperate concerns between stylistic/ux concerns and business logic.

Changes:


lindapaiste:
Thank you so much for looking into this!  I've never been able to get Storybook to work so I'm uncertain of what it looked like before.  It does seem from you screenshots that the components which use `styled-components` are rendering properly, but the ones that use `.scss` files aren't getting their styles loaded.  Do you have any ideas about loading `.scss` files?  No pressure if you don't, this is still a huge improvement.

.storybook/main.js
```diff
@@ -1,29 +1,35 @@
-const path = require('path');
-
-module.exports = {
+/** @type { import('@storybook/react-webpack5').StorybookConfig } */
+const config = {
   stories: ['../client/**/*.stories.(jsx|mdx)'],
   addons: [
-    '@storybook/addon-actions',
-    '@storybook/addon-docs',
-    '@storybook/addon-knobs',
```
lindapaiste, line 8: I noticed that you removed some packages from the `addons` array here but kept them in the `package.json`.  Do they need to be kept as dependencies or should they be removed?

.storybook/preview.js
```diff
@@ -1,31 +1,20 @@
...
-);
+const store = configureStore(initialState);
 
-addDecorator(withThemePlayground({
-  theme: themeConfigs,
-  provider: ThemeProvider
-}));
+export const decorators = [
+  (Story) => (
+      <Provider store={store}>
```
lindapaiste: If we are connecting to Redux here then maybe we want a way to modify the initial state on a per-story basis?  Just a thought.  That's something for a future PR.

package.json
```diff
@@ -244,5 +252,10 @@
     "webpack-dev-middleware": "^5.3.1",
     "webpack-hot-middleware": "^2.25.1",
     "xhr": "^2.6.0"
+  },
+  "eslintConfig": {
```
lindapaiste: We have an existing `.eslintrc` file so this needs to be put there instead.
https://github.com/processing/p5.js-web-editor/blob/3493b3b8099d42cac7220f0181ba4fbafebddbf0/.eslintrc#L2

This is why you are getting errors on the "Test and lint code base" Github action.

Edit: hmmm...  I thought that the errors would go away after moving the "extends" to the `.eslintrc` but they didn't.  I may need to look into what prettier config eslint is actually using, as `eslint-plugin-storybook` has a slightly different prettier config than ours.
https://github.com/storybookjs/eslint-plugin-storybook/blob/main/.prettierrc.js
https://github.com/processing/p5.js-web-editor/blob/develop/.prettierrc

lindapaiste (changes requested):
Looks great overall!  We do need to fix the eslint setup before merging.

.storybook/main.js
ofhope: They should be removed now. They've been deprecated https://www.npmjs.com/package/@storybook/addon-knobs

Actions are built in by default. And knobs have been replaced with `args` and `argTypes` where you can configure controls to play with. I've updated Buttons and Icons stories to use these rather than the knobs that were there.

package.json
ofhope: We could probably remove this for now and re-include it in the future. This was a suggestion by storybook when initialising

ofhope:
@lindapaiste It looks like the css is already built so it was just a matter of including the css file 🎉 
```
import '../client/styles/build/css/main.css'
```

These screen shots look better

Screenshot 2023-07-18 at 11 32 13 am

Screenshot 2023-07-18 at 11 32 15 am

package.json
ofhope: To minimise the changes in this PR I reverted this. Can always re-include it later.

```
  "eslintConfig": {
    "extends": [
      "plugin:storybook/recommended"
    ]
  }
```

.storybook/preview.js
ofhope: I'm not sure what Redux is used for but all components with stories needed it. It could be good to try and remove it as a dependancy from Storybook/Design System components 🤷🏽

lindapaiste: Yeah I actually agree. I think of Storybook as being for pure UI components which are not connected to Redux.  But the current `FileNode` and the `SocialAuthButton` components are connected and they have stories so that's why we need it for now.

ofhope:
@lindapaiste so it looks like including the storybook eslint rules in `.eslintrc` works. It took an eslint restart for it to start prompting with misconfigured stories. 

```
"extends": ["airbnb", "prettier", "plugin:storybook/recommended"],
```

As a test commenting out the icon stories and looking at the warnings we can see `eslint(storybook/default)`

> The file should have a default export.

Screenshot 2023-07-19 at 9 16 17 am

And if you fix that you get another error about having at least one story.

Screenshot 2023-07-19 at 9 23 25 am

lindapaiste:
Right now my only concern is the changes in unrelated files due to Prettier.  Ideally we want it to pass the linting checks without changing anything.  I noticed in your `package-lock.json` file that there's a minor version bump of `prettier` and this could be the cause.

Can you try the following steps?

1. `git revert 1c3748d7d349040793ef6c74c168b998f7098262` To rollback the linting changes to `FileNode`, etc. from [this commit](https://github.com/processing/p5.js-web-editor/pull/2310/commits/1c3748d7d349040793ef6c74c168b998f7098262).
2. `npm install prettier@2.2.1` to install the specific version of `prettier` that we had previously.
3. `npm run lint` -- does it pass?

I definitely want to get this merged in ASAP so that we can keep building on it.  Now that you've got the basics of Storybook working I'm seeing problems and areas for improvement in the stories and in the components themselves.  I've been reading some docs to figure out how to make the most out of Storybook and I've figured out theme switching, selecting icons from a dropdown on the `Button` story, etc.

client/common/Button.stories.jsx
```diff
@@ -1,22 +1,22 @@
...
 };
 
-export const AllFeatures = () => (
-  <Button
-    disabled={boolean('disabled', false)}
-    type="submit"
-    label={text('label', 'submit')}
-  >
-    {text('children', 'this is the button')}
+export const AllFeatures = (args) => (
```
lindapaiste, line 17: The controls on this story don't seem to do anything because we aren't passing down all of the `args`.  Looks like we can [write the story as an object](https://storybook.js.org/docs/react/writing-stories/args) with the new "Component Story Format 3".  Very cool.

```
export default {
  title: 'Common/Button',
  component: Button
}

export const AllFeatures = {
  args: {
    children: 'this is the button'
  }
};
```

ofhope: Thanks for investigating linting... I mistakenly thought these were linting issues on the main branch. I wasn't giving them much thought. I'll try your suggestion later today and see if I can revert them.

ofhope: Yeah with stories you can set some defaults for all stories. Customise default args per story. Or present them with fixed values too if you like. 

Previously this `AllFeatures` story had dynamic args as knobs and all the reset were fixed.

This is a button story from a personal project where args were overridden.
```
const Template: Story<ButtonProps> = (args) => <Button {...args} />;

export const FullWidth = Template.bind({});
FullWidth.args = {
  primary: true,
  children: 'Button',
  fullWidth: true
};
```

ofhope:
One other addition. I ran `npm uninstalled @storybook/addons-knobs` this is deprecated.

lindapaiste (dismissed):
🚀 Thank you for all of your work on this!

I'm going to go ahead and merge this.  Feel free to keep adding other improvements in additional PRs.

lindapaiste:
Ah 💩, one last thing.  Can you please commit the `package-lock.json` file after running `npm install prettier@2.2.1`?

lindapaiste (approved):
Thank you!

PR #2312, opened 2023-07-18
Account height

sdivyanshu90:
Fixes : 
When the High Contrast theme is enabled, a white space or box appears at the bottom of the page. This issue is caused by the height of the element.

Screenshot of the error:
image

Changes:
- changes height to min-height


sdivyanshu90:
Hi @raclim, could you please review my pull request? Thank you!

lindapaiste:
Thanks for finding this issue!  I see that it happens when the screen height is smaller than the height of the content.  The part that you have to scroll down to see doesn't get the dark background.

I do kinda wonder why we are putting the background color on the `.account-settings__container` div and not on a higher-up element like the `body` or `div id="root"` but that's beyond the scope of this PR.

As far as fixing the issue, your change of `width` to `min-width` is perfect!  The other changes are a bit more suspect.  I'll leave comments on specific lines in a sec.

client/styles/components/_account.scss
```diff
@@ -3,7 +3,7 @@
     color: getThemifyVariable('primary-text-color');
     background-color: getThemifyVariable('background-color');
   }
-  height: 100%;
+  min-height: 100%;
```
lindapaiste, line 6: Perfect ⭐

client/styles/components/_forms.scss
```diff
@@ -52,7 +52,7 @@
 
 .form__input {
   width: 100%;
-  min-width: #{360 / $base-font-size}rem;
+  min-width: 10%;
```
lindapaiste: If the `width` is `100%` then the `min-width` of `10%` will never be used.  The `10%` will never be greater than the `100%`.  So this doesn't make a ton of sense.  BUT....it actually is an improvement over what we have now!  What you've discovered is that the current `min-width: 360px` (`min-width: #{360 / $base-font-size}rem;`) is not good as it causes horizontal overflow when the screen is narrower than ~390px.

Instead of changing it to `min-width: 10%;`, let's delete the existing line and replace it with nothing.

client/styles/components/_nav.scss
```diff
@@ -19,6 +19,7 @@
   display: flex;
   flex-direction: row;
   justify-content: flex-end;
+  min-width: 10%;
```
lindapaiste, line 22: I don't think this line would come into play it all.  Correct me if I'm wrong.  I think you should delete this.

lindapaiste: I don't think this line would come into play it all. Correct me if I'm wrong. I think you should delete this.

client/styles/components/_account.scss
```diff
@@ -29,3 +29,22 @@
...
+
+// @media screen and (min-width: 10%) {
+//   .social_button_github,
+//   .social_button_google {
+//     display: inline-block;
+//     width: auto;
+//     margin-bottom: 0;
+//   }
+// }
+
```
lindapaiste: It's probably better to delete this than to add commented-out styles into the codebase.

lindapaiste (changes requested):

sdivyanshu90:
Thank you for your review,  @lindapaiste! I've implemented all the mentioned changes.

client/styles/components/_forms.scss
```diff
@@ -52,7 +52,6 @@
 
 .form__input {
   width: 100%;
-  min-width: #{360 / $base-font-size}rem;
```
lindapaiste, line 55: I gave you bad instructions here because I didn't realize that deleting this line will cause the inputs to be too narrow on large screens. 🤦‍♀️ 

Thinking this through...
Right now it is 360px wide always.  This is a problem on small screens where 360px is too big.  We want it to be 360px when possible but not too large to fit.  The `width: 100%` alone doesn't do this because there is no width on the parent container (it will be too small on large screens).  Doing a fixed width with `max-width: 100%` also doesn't work unless we also put a `max-width` on the container (it will overflow on small screens).

Option A - We put it back as it was before and merge this PR which fixes the background color issue.

Option B - The not perfect but very good fix.
```
.form__input {
  width: #{360 / $base-font-size}rem;
  max-width: 90vw;
```

The "perfect" fix would involve a `max-width: 100%;`, but that involves changing the CSS on a few ancestors and it becomes beyond the scope of this PR.

sdivyanshu90: Hey, thank you for the review. Right now, I will work on option B.

> The "perfect" fix would involve a max-width: 100%;, but that involves changing the CSS on a few ancestors and it becomes beyond the scope of this PR.

Sure, I will work on the perfect fix, but not immediately. I will address it later.

lindapaiste (approved):

client/styles/components/_forms.scss
lindapaiste: Re: "perfect" fix, I actually have a CSS which I think is good but this didn't seem like the right PR for it.
```
.form__input {
  width: #{360 / $base-font-size}rem;
  max-width: 100%;
```
```
.form-container {
  max-width: 100vw;
  padding: #{20 / $base-font-size}rem;
```
```
.form-container__content {
  max-width: 100%;
  & > * {
    max-width: 100%;
  }
```

PR #2313, opened 2023-07-18
fix/exportProjectAsZip

### LINKED ISSUE #2311: Error: Actions must be plain objects when downloading sketches as zip
### p5.js version

_No response_

Windows

### Web browser and version

Chrome version :114.0.5735.198

### Actual Behavior

The error occurs while downloading the sketch zip file. Although the zip file gets downloaded, but it shows the Error: "Actions must be plain objects. Use custom middleware for async actions."
I think the error occurs due to the attempt to dispatch an asynchronous action directly to the Redux store without using any middleware.

### Expected Behavior

Error should not occur.

### Steps:
1. Go to File tab.
2. Examples then download any sketch by using the down arrow button.

adadgarg:
Fixes #2311 

Changes:
Removed the exportProjectAsZip function and directly link to the download URL.


lindapaiste (approved):

lindapaiste (approved):
Thank you!   I can confirm that this works and that it fixes the issue so it's good to merge but I do have some further thoughts.

Initially when I mentioned using an `<a>` tag I was thinking about replacing the `<button>` itself with the `<a>`.  However that caused [styling issues](https://github.com/processing/p5.js-web-editor/issues/2311#issuecomment-1638982757).  This approach with the hidden `<a>` tag avoids those issues so that's great. 👍 

We also have export logic in the `Nav` component for exporting the current sketch from the editor screen.  So now I'm wondering if it may be better to move your logic into a utility function which can be used in both places.

PR #2319, opened 2023-07-22
Fix: Hover style

### LINKED ISSUE #2223: collection description `EditableInput` text doesn't change color on hover - CSS Specificity
### p5.js version

_No response_

None

### Web browser and version

_No response_

### Actual Behavior

This is the current behavior of the `EditableInput` when there is an existing description on the collection.

normal:
image
hover:
image

image

### Expected Behavior

This is the current behavior when there is no existing description.

hover:
image

### Steps:
1. View the collection details page for a collection which you own
2. Give it a description
3. Hover over the description name
4. Notice that the pencil turns pink but the text does not

----

The text of an existing collection description does not change color when hovered.  If there is no existing description, then the words "Add description" will be pink when hovered.

I believe this is a mistake, but perhaps it's the intended behavior.  Can anyone confirm if we do or don't want the text to be pink when hovered?

We have a style targeting the `.editable-input__label` when it is used inside of `.editable-input--has-value` inside of `.collection-metadata__description`
https://github.com/processing/p5.js-web-editor/blob/7212f76f6261ad52a33c9c87e17e5d90be903b3c/client/styles/components/_collection.scss#L77-L81

This selector is very specific, so it takes priority over the `:hover` style of `button.editable-input__label`.
https://github.com/processing/p5.js-web-editor/blob/7212f76f6261ad52a33c9c87e17e5d90be903b3c/client/styles/components/_editable-input.scss#L7C1-L21

The specific selector only applies to non-empty descriptions (due to the `.editable-input--has-value` class) so the hover effect still works if the description is empty.

We can fix this by appending a `:not(:hover)` to the override style in `_collection.scss`.

adadgarg:
Fixes #2223 

Changes:
Added :not(:hover) .editable-input--has-value so that it shows correct color while hovering.


lindapaiste (approved):
Perfect, thanks!

PR #2320, opened 2023-07-23
Cleanup `getProjectsForUser` function

lindapaiste:
Progress on issue #521 

Changes:
- Combine common logic from `getProjectsForUser` (internal API) and `apiGetProjectsForUser` (public API).
- Add comments explaining the difference between the two handlers.
- Fix the "callback hell" by converting to `async`/`await` syntax.
- Update the test file to support async user lookup.
- **Behavior change** - internal API returns an error if there is no `req.params.username` instead of returning an empty array.

Note: there is a lot of discussion in PR #973 about wrapping an entire Express handler in a `try`/`catch` like I am doing here and whether there are better alternatives.  We may want to revisit this as I'm not sure what the current best practice is.


PiyushChandra17 (approved):
Nicely Done

raclim (approved):
Thanks for your work on this!

PR #2321, opened 2023-07-23
Delete unnecessary `return new Promise` in Redux actions.

lindapaiste:
Changes:
- Remove all unnecessary `new Promise` constructors from Redux actions.

We have a lot of instances where we a wrapping an API call in `return new Promise((resolve, reject) => {`.  This is unnecessary because the API call is already a `Promise`. 

This code:
```
return new Promise((resolve) => {
  someFunction(args)
    .then((response) => {
      /** ... **/
      resolve();
    })
    .catch((error) => {
      /** ... **/
      resolve({ error });
    });
});
```
Is the same as this:
```
return someFunction(args)
  .then((response) => {
    /** ... **/
  })
  .catch((error) => {
    /** ... **/
    return { error };
  });
```
Both return a `Promise` which either resolves to `void` (on success) or resolves to an object with an `error` property (on failure).

There's a lot more cleanup that can be done here in the future, but right now I'm just changing this one thing.  Notes for the future.

- We often resolve to `{ error }` but we don't use this value anywhere.
-I've encountered uncaught errors which are thrown in the `catch` clauses.  We are assuming that the `error` variable always has a property `error.response.data` and this is not always the case, as the `catch` clause catches ALL errors. (This won't crash the app but it logs to the console).
- Some async action creator functions return the Promise and others don't.

This returns a `Promise`:
```
return (dispatch) =>
  apiClient
    .get('path')
```
This returns `void`:
```
return (dispatch) => {
  apiClient
    .get('path')
};
```

### DIFF
--- a/client/modules/IDE/actions/files.js
+++ b/client/modules/IDE/actions/files.js
@@ -89,25 +89,22 @@ export function handleCreateFile(formProps, setSelected = true) {
     const { files } = state;
     const { parentId } = state.ide;
     const projectId = state.project.id;
-    return new Promise((resolve) => {
-      submitFile(formProps, files, parentId, projectId)
-        .then((response) => {
-          const { file, updatedAt } = response;
-          dispatch(createFile(file, parentId));
-          if (updatedAt) dispatch(setProjectSavedTime(updatedAt));
-          dispatch(closeNewFileModal());
-          dispatch(setUnsavedChanges(true));
-          if (setSelected) {
-            dispatch(setSelectedFile(file.id));
-          }
-          resolve();
-        })
-        .catch((error) => {
-          const { response } = error;
-          dispatch(createError(response.data));
-          resolve({ error });
-        });
-    });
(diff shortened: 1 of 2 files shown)


client/modules/IDE/actions/files.js
```diff
@@ -89,25 +89,22 @@ export function handleCreateFile(formProps, setSelected = true) {
...
-          }
-          resolve();
-        })
-        .catch((error) => {
-          const { response } = error;
-          dispatch(createError(response.data));
-          resolve({ error });
-        });
-    });
+    submitFile(formProps, files, parentId, projectId)
```
PiyushChandra17, line 92: I think `return` is missing here, otherwise all the changes looks good to me. Nicely Done.

PiyushChandra17 (changes requested):
I think `return` is missing, otherwise all the refactor looks good to me. Nicely done.

PR #2325, opened 2023-07-25
test: add ErrorModal tests and story

ofhope:
No issue number

Changes:

Adds tests and Storybook story around the ErrorModal component

Screenshot 2023-07-25 at 2 41 03 pm

### DIFF
--- a/client/modules/IDE/components/ErrorModal.stories.jsx
+++ b/client/modules/IDE/components/ErrorModal.stories.jsx
@@ -1,47 +1,30 @@
-import React from 'react';
-
 import ErrorModal from './ErrorModal';
 
 export default {
   title: 'IDE/ErrorModal',
   component: ErrorModal,
   argTypes: {
-    type: {
-      options: [
-        'forceAuthentication',
-        'staleSession',
-        'staleProject',
-        'oauthError'
-      ],
-      control: { type: 'select' }
-    },
-    service: {
-      options: ['google', 'github'],
-      control: { type: 'select' }
-    },
     closeModal: { action: 'closed' }
   }
 };
 
-const Template = (args) => <ErrorModal {...args} />;
-
-export const ForceAuthenticationErrorModal = Template.bind({});
-ForceAuthenticationErrorModal.args = {
-  type: 'forceAuthentication'
+export const ForceAuthenticationErrorModal = {
+  args: {
+    type: 'forceAuthentication'
+  }
 };
-
-export const StaleSessionErrorModal = Template.bind({});
-StaleSessionErrorModal.args = {
-  type: 'staleSession'
+export const StaleSessionErrorModal = {
+  args: {
+    type: 'staleSession'
+  }
 };
-
-export const StaleProjectErrorModal = Template.bind({});
-StaleProjectErrorModal.args = {
-  type: 'staleProject'
+export const StaleProjectErrorModal = {
+  args: {
+    type: 'staleProject'
+  }
 };
-
-export const OauthErrorModal = Template.bind({});
-OauthErrorModal.args = {
-  type: 'oauthError',
-  service: 'google'
+export const OauthErrorModal = {
+  args: {
+    type: 'oauthError'
+  }
 };
--- a/client/modules/IDE/components/ErrorModal.stories.jsx
+++ b/client/modules/IDE/components/ErrorModal.stories.jsx
@@ -1,0 +1,47 @@
+import React from 'react';
+
+import ErrorModal from './ErrorModal';
+
+export default {
+  title: 'IDE/ErrorModal',
+  component: ErrorModal,
+  argTypes: {
+    type: {
+      options: [
+        'forceAuthentication',
+        'staleSession',
+        'staleProject',
+        'oauthError'
+      ],
+      control: { type: 'select' }
+    },
+    service: {
+      options: ['google', 'github'],
+      control: { type: 'select' }
+    },
+    closeModal: { action: 'closed' }
+  }
+};
+
+const Template = (args) => <ErrorModal {...args} />;
+
+export const ForceAuthenticationErrorModal = Template.bind({});
+ForceAuthenticationErrorModal.args = {
+  type: 'forceAuthentication'
+};
+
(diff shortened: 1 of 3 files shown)


client/modules/IDE/components/ErrorModal.stories.jsx
```diff
@@ -0,0 +1,47 @@
...
+    closeModal: { action: 'closed' }
+  }
+};
+
+const Template = (args) => <ErrorModal {...args} />;
+
+export const ForceAuthenticationErrorModal = Template.bind({});
+ForceAuthenticationErrorModal.args = {
+  type: 'forceAuthentication'
+};
```
lindapaiste, line 15: We don't need to do the `Template.bind({});` stuff anymore with the version that you installed. The alternate syntax is actually more lines though 🙃 so it's whatever.

```
export const ForceAuthenticationErrorModal = {
  args: {
    type: 'forceAuthentication'
  }
};
```

client/modules/IDE/components/ErrorModal.stories.jsx
```diff
@@ -0,0 +1,47 @@
...
+  title: 'IDE/ErrorModal',
+  component: ErrorModal,
+  argTypes: {
+    type: {
+      options: [
+        'forceAuthentication',
+        'staleSession',
+        'staleProject',
+        'oauthError'
+      ],
```
lindapaiste: I think it might be better to define these options in the `propTypes` of the component rather than here.  If we do it that way then Storybook will pick up on the options automatically and we'll also get the validation in `prop-types`.

```
ErrorModal.propTypes = {
  type: PropTypes.oneOf([
    'forceAuthentication',
    'staleSession',
    'staleProject',
    'oauthError'
  ]).isRequired,
  closeModal: PropTypes.func.isRequired,
  service: PropTypes.oneOf(['google', 'github'])
};
```

client/modules/IDE/components/ErrorModal.unit.test.jsx
lindapaiste, line 1: The tests look good.  This component renders a message and we're just checking that it renders the correct message for each `type` argument.  There's no internal logic to be tested.  So I wonder if a snapshot test makes more sense?  There's pros and cons to each approach.  A snapshot would detect any unintended changes over time but it doesn't necessarily validate that the correct message is printed for each case -- we'd need to examine the first snapshot manually, or use it in combination with these tests that you wrote.

I'm just thinking out loud here.  There is no problem with these tests. ✔️

client/modules/IDE/components/ErrorModal.stories.jsx
```diff
@@ -0,0 +1,47 @@
...
+    },
+    service: {
+      options: ['google', 'github'],
+      control: { type: 'select' }
+    },
+    closeModal: { action: 'closed' }
+  }
+};
+
+const Template = (args) => <ErrorModal {...args} />;
```
lindapaiste: FYI I tried seeing if wrapping in `<Overlay>` here would show us the actual UI but does not ☹️.
image

Not a problem.  Maybe something we can play with later if we want to.

lindapaiste (dismissed):
Looks good.  I have some specific feedback but there's nothing here that's a problem.

client/modules/IDE/components/ErrorModal.unit.test.jsx
ofhope: We could potentially test the login/sign in clicking. To ensure that it triggers the closeModal function and triggers a History.push. Partly left out due to laziness and partly wasn't sure if it is the right approach with a modal. I was thinking a generic modal should have customisable actions (cancel, submit, login, sign in, accept, decline). And avoid this baked in behaviour.

As to snapshot testing. In the past I've admittedly become very numb to checking PR snapshots. My personal preference is to use testing-library. `('log in')).toBeVisible()` this does check the element is not obscured and avoids opacity: 0 display: none etc. Its minor but offers some additional checking. 🤷🏽

client/modules/IDE/components/ErrorModal.stories.jsx
ofhope: Oh interesting. I'm still familiarising myself with the components in the application. I was wondering where the overlay was 😄  It would make a nicer story seeing the modal represented with the overlay.

client/modules/IDE/components/ErrorModal.stories.jsx
ofhope: Oh awesome I'll check it out. That'll save a fair bit of boiler plate

client/modules/IDE/components/ErrorModal.stories.jsx
ofhope: The action wasn't picked up with this 😞 so left that argType.

client/modules/IDE/components/ErrorModal.stories.jsx
lindapaiste: I have to try it again after the theming changes in #2326.  I think the way that they SCSS gets generated is that it creates selectors like `.light .overlay` and `.dark .overlay` but not just `.overlay`.  So it's possible that the overlay UI will work once it's inside the right parent container.

client/modules/IDE/components/ErrorModal.stories.jsx
lindapaiste: Yeah when I was playing around with it I left the `closeModal` in.  I'm not aware of a trick for those, at least not yet.

FYI, if you want to keep it as a dropdown rather than radio buttons (I don't think it matters?) then you can provide a partial config like `service: {  control: { type: 'select' }  }` which specifies the `control` and it will still infer the `options`.

client/modules/IDE/components/ErrorModal.stories.jsx
lindapaiste: Honestly it feels like a bit of a naming mistake that this is called `ErrorModal` when it doesn't include the modal part.  Really it's more of an `ErrorMessage`!

It could potentially be modified so that it takes the overlay's `title` and `ariaLabel` as props and renders both the `Overlay` and the message.

lindapaiste (approved):

PR #2326, opened 2023-07-25
Add theme switching to Storybook

lindapaiste:
Changes:

- Add a "Theme" option to Storybook controls.
- Create a decorator which applies the selected theme to the story.
- Default view is all 3 themes together, but can also select a single theme.

image
image
image

Could be improved if I make the background fill the entire preview area.

@ofhope - thoughts?


lindapaiste:
Trying out this feature with all of the new stories added in #2398 has revealed a bunch of things which are not handled perfectly.

TODOs:
- Some SVGs (ex. EditableInput pencil) don't work due to a lack of `viewbox`.  I need to look into the `svgr` plugin settings.
- Elements which use `fixed` or `absolute` positioning (ex. modals, FloatingActionButton) will show up on top of each other when viewing all themes at once.
- File uploader does not work in all-theme mode due to using an element id to initialize the dropzone, as this id is not unique when three versions are rendered at once.
- Modals aren't getting the correct text color in the dark and contrast themes.  I suspect they inherit this from some parent when in the app?

ofhope:
This is awesome!  Its really nice to be able to isolate and explore components with a theme.

The `all` view with three together may be questionable. It doesn't play so nice with more complex stories, Modal like things. It'd be great for primitives (button, input, icon etc). But if more themes become available it could clutter the story (probably unlikely that more themes will come just speculating here).

One idea for a view with all themes is to mimic Compositions https://bit.cloud/teambit/design/inputs/select

Screenshot 2023-09-15 at 10 06 14 am

Storybook has MDX support so you could use your `all` view in an MDX document to print out all view variations in a supporting document. This would leave the stories as singular views and allow control over which components had a supporting MDX file with an all theme view.

But these are just ideas for the future. I think whats here is an enhancement by exposing the theme in the first place. It can be iterated on in the future if needed.

.storybook/preview.js
```diff
@@ -21,5 +21,9 @@ export const decorators = [
       </MemoryRouter>
```
ofhope, line 21: It looks like `ThemeProvider` has moved to the withThemeProvider HOC. Might need to remove `<ThemeProvider>` as well.

ofhope (approved):

raclim (approved):

PR #2337, opened 2023-07-30
Fix #2336: `CookieConsent` cannot be hidden properly on very small screens

### LINKED ISSUE #2336: `CookieConsent` cannot be hidden properly on very small screens
### p5.js version

_No response_

None

### Web browser and version

_No response_

### Actual Behavior

The positioning of the `CookieConsent` is always fine when it is visible, as we use `bottom: 0` for this.
image

When it's hidden we use `bottom: -300px` (in rem).  Moving the bottom down by exactly `300px` will hide the banner if the banner is >= `300px`.  This is almost always true, even on the vast majority of mobile devices.  But using fixed numeric positions is prone to edge case bugs.  On a very narrow screen the banner can be taller than `300px` due to text wrapping.  In that case, part of the banner will always be visible and it's impossible to remove this visible part.
image

### Expected Behavior

I expect it to be completely hidden regardless of screen size.

### Steps:
 1. Open the web app
 2. Shrink the window

lindapaiste:
Fixes #2336 

Changes:
- The change in position is based on the height of the element rather than a fixed amount.
- `bottom` is always `0`.
- The change now comes from a `transform` `translateY`.
- I used `105%` instead of `100%` because I was paranoid that the top 1px of the banner might sometimes be visible due to rounding errors if the banner starts exactly below where the screen ends.  Not sure if this is an actual issue but I figured it's better to be safe. 🤷 


PiyushChandra17 (approved):
Nicely Done

raclim (approved):
Thank you for working on this!

PR #2343, opened 2023-08-03
Create tests for `Toast` component

lindapaiste:
Should be reviewed/merged after #2329 

Changes:
- Create unit tests for the end-to-end behavior of the Toast (check that it can be opened and closed).
- Add an ARIA role to the Toast.  I used [`status`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role) rather than [`alert`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role) because the lower priority seemed like it made more sense.


raclim (approved):
Thanks for working on this!

PR #2346, opened 2023-08-04
Mobile overlay fix: works on smaller devices

dewanshDT:
This is a part of mobile responsive project

Changes:
It changes the _overlays.scss file

Before:
image
overlays_scss_before

After:
overlays_scss_after


lindapaiste:
It looks like there’s a vertical scroll bar in the “after” screenshot?  So there’s a container around that content which needs to have a change to its height.  Probably specific to the settings/preferences and not in this overlay.scss file.  I know that I said to make a PR with just the overlay.scss changes so we could fix that later if you want or you could fix it now as part of this PR.

client/styles/components/_overlay.scss
```diff
@@ -23,15 +23,22 @@
   flex-wrap: wrap;
   flex-flow: column;
   max-height: 80%;
+  max-width: 80%;
   position: relative;
   padding-bottom: #{25 / $base-font-size}rem;
+
+  @media (max-width: 650px) {
```
lindapaiste, line 30: My first instinct was that we should centralize these breakpoints into SCSS and JS variables.  Then I thought about it more and we don't necessarily need to use the same breakpoints for every component.  The point at which the overlay no longer fits might be different than the point where the nav bar doesn't fit, for example.  But if we want to add breakpoints to elements which are rendered as children of an overlay (like the preferences) then those would probably need to use the same breakpoints as the overlay.  Curious your thoughts on this.

lindapaiste:
There are some specific issues with specific modals.  I DON'T WANT YOU TO FIX THESE.  I'm just looking things over to see where we are at.  We've always said that the collection lists, etc. would be something that you worked on last only if you had time since your main focus in on the IDE.   Maybe we open some issues and let other contributors handle these?

Components rendered inside `<Overlay/>`:
- [ ] `AddToCollectionList` (list of collections) - needs a max-width, move searchbar below header.
- [ ] `AddToCollectionSketchList` (list of sketches) - needs a max-width, move searchbar below header.
- [ ] `Preferences` - need to remove the explicit height on the container.
- [ ] `About` - Content needs rearranging.  Some of the changes in #2342 may help.
- [x] `Feedback` -- this is in the code but not actually in the app 🙃 
- [x] `ShareModal`
- [ ] `KeyboardShortcutModal` - almost good but we need a max-width on the container to force long lines to wrap.
- [x] `ErrorModal`
- [ ] `CollectionCreate` - form fields need a max width.

lindapaiste (changes requested):
The purpose of this PR is to make overlays fill the screen on mobile devices and you've mostly done that.  The one modal-specific issue which I think is worth addressing now is the height of the "fixed height" overlay, only because that is declared in this same `_overlay.scss` file.  Our intention is that overlays should fill the whole screen on small devices so I think that should apply here too.
https://github.com/processing/p5.js-web-editor/blob/31e64d5dca088c6254d626bbf3cc07f7a3954270/client/styles/components/_overlay.scss#L50-L53

We discussed the possibility of writing media queries "mobile first" with the mobile as the default and a `min-width` media query for the desktop styles.  That might be good here because the `height: 80vh;` only applies on `.overlay--is-fixed-height` if the `min-width` query is met.  If we write it desktop-first then we need two styles: `height: 80vh` and an override to set it back to `height: 100vh;` on small screens. Whereas with mobile-first we only need one style, with a selector and a media query, to apply `height: 80vh`.  But I'll leave that up to you.

client/styles/components/_overlay.scss
dewanshDT: yeah I agree the breakpoints does not need to be same when they are completely unrelated like the navbar and the overlay

dewanshDT:
> It looks like there’s a vertical scroll bar in the “after” screenshot? So there’s a container around that content which needs to have a change to its height. Probably specific to the settings/preferences and not in this overlay.scss file. I know that I said to make a PR with just the overlay.scss changes so we could fix that later if you want or you could fix it now as part of this PR.

yeah I didn't change that due to the reason that these changes were already present in the mega PR that you made me merge in before I started working, I think that we should get that part of  the PR merged in the repo, what are your suggestions on this?

dewanshDT:
> There are some specific issues with specific modals. I DON'T WANT YOU TO FIX THESE. I'm just looking things over to see where we are at. We've always said that the collection lists, etc. would be something that you worked on last only if you had time since your main focus in on the IDE. Maybe we open some issues and let other contributors handle these?
> 
> Components rendered inside `<Overlay/>`:
> 
> * [ ]  `AddToCollectionList` (list of collections) - needs a max-width, move searchbar below header.
> * [ ]  `AddToCollectionSketchList` (list of sketches) - needs a max-width, move searchbar below header.
> * [ ]  `Preferences` - need to remove the explicit height on the container.
> * [ ]  `About` - Content needs rearranging.  Some of the changes in [added styled component to About.jsx #2342](https://github.com/processing/p5.js-web-editor/pull/2342) may help.
> * [x]  `Feedback` -- this is in the code but not actually in the app 🙃
> * [x]  `ShareModal`
> * [ ]  `KeyboardShortcutModal` - almost good but we need a max-width on the container to force long lines to wrap.
> * [x]  `ErrorModal`
> * [ ]  `CollectionCreate` - form fields need a max width.

Yeah I agree, as we leave some of the issues for others to fix, we should mention them in the Issues section, also we can add the good first issue label in them

dewanshDT:
> The purpose of this PR is to make overlays fill the screen on mobile devices and you've mostly done that. The one modal-specific issue which I think is worth addressing now is the height of the "fixed height" overlay, only because that is declared in this same `_overlay.scss` file. Our intention is that overlays should fill the whole screen on small devices so I think that should apply here too.
> 
> https://github.com/processing/p5.js-web-editor/blob/31e64d5dca088c6254d626bbf3cc07f7a3954270/client/styles/components/_overlay.scss#L50-L53
> 
> We discussed the possibility of writing media queries "mobile first" with the mobile as the default and a `min-width` media query for the desktop styles. That might be good here because the `height: 80vh;` only applies on `.overlay--is-fixed-height` if the `min-width` query is met. If we write it desktop-first then we need two styles: `height: 80vh` and an override to set it back to `height: 100vh;` on small screens. Whereas with mobile-first we only need one style, with a selector and a media query, to apply `height: 80vh`. But I'll leave that up to you.

I'll check it out, I've never saw this before

lindapaiste:
@dewanshDT I'd like to get this PR merged.  I think we can fix the fixed height ones by changing the current CSS to this. Can you check it out?
```
 /* Fixed height overlay */ 
 .overlay--is-fixed-height .overlay__body { 
  @media (min-width: 651px) {
     height: 80vh; 
   }
 } 
```

dewanshDT:
Yeah, just a minute

dewanshDT:
check it out

client/styles/components/_overlay.scss
```diff
@@ -49,10 +56,13 @@
 
 /* Fixed height overlay */
 .overlay--is-fixed-height .overlay__body {
-  height: 80vh;
+  height: 100vh;
+
+  @media (min-width: 770px) {
+    height: 80vh;
+  }
```
lindapaiste: Now I'm playing with this and my head is exploding 🤯 but I think that `height: 100vh;` is actually all that we need?  Because the `max-height: 80%;` will apply over 651px so that's the same as `height: 80vh;`

dewanshDT: So I should remove the media query then?

lindapaiste: Yeah I think so.  I don't like to have code that doesn't do anything.  The overlay looks good though!

dewanshDT: done!

lindapaiste (approved):
This is a major improvement to mobile usability!

PR #2347, opened 2023-08-04
FloatingActionButton: added the component to be used in smaller views

dewanshDT:
This is the part of the mobile responsive project

Changes:
Floating action button component is added, to be used in the mobile views


client/modules/IDE/components/FloatingActionButton.jsx
```diff
@@ -0,0 +1,67 @@
...
+  return (
+    <Button
+      data-behaviour={isPlaying ? 'stop' : 'run'}
+      style={{ paddingLeft: isPlaying ? 0 : remSize(5) }}
+      onClick={() => {
+        if (!isPlaying) {
+          props.syncFileContent();
+          dispatch(startSketch());
+        }
+        if (isPlaying) dispatch(stopSketch());
```
lindapaiste: I would use an `else` since these two conditions cover all cases.  IMO when using `if`/`else` it feels cleaner to write the positive condition first.  `if (isPlaying) { } else { }`

client/modules/IDE/components/FloatingActionButton.jsx
```diff
@@ -0,0 +1,67 @@
...
+  ${prop('Button.secondary.default')};
+  aspect-ratio: 1/1;
+  border-radius: 99999px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  box-shadow: rgba(0, 0, 0, 0.24) 0px 3px 8px;
+  /* scale: 0.75; */
+  /* padding-left: 1.8rem; */
+  &[data-behaviour='stop'] {
```
lindapaiste: I'm not a fan of using `data-` attributes for styling.  It feels like a bit of a hack.  I think using a `className` makes more sense.

client/modules/IDE/components/FloatingActionButton.jsx
```diff
@@ -0,0 +1,67 @@
...
+`;
+
+const FloatingActionButton = (props) => {
+  const isPlaying = useSelector((state) => state.ide.isPlaying);
+  const dispatch = useDispatch();
+
+  return (
+    <Button
+      data-behaviour={isPlaying ? 'stop' : 'run'}
+      style={{ paddingLeft: isPlaying ? 0 : remSize(5) }}
```
lindapaiste, line 48: I don't love this but I know it's here to fix a genuine problem so it's best to keep it for now so as not to fall deep into an SVG rabbit hole.  We need this because the `PlayIcon` does not appear visually centered within its viewbox.  IMO that's a problem with the icon and we may want to modify the icon to have whitespace on the side so that it looks centered.

client/modules/IDE/components/FloatingActionButton.jsx
```diff
@@ -0,0 +1,67 @@
...
+const Button = styled.div`
+  position: fixed;
+  right: ${remSize(20)};
+  bottom: ${remSize(20)};
+  height: ${remSize(60)};
+  width: ${remSize(60)};
+  z-index: 3;
+  cursor: pointer;
+  ${prop('Button.secondary.default')};
+  aspect-ratio: 1/1;
```
lindapaiste, line 20: This property [has way more support than I thought](https://caniuse.com/mdn-css_properties_aspect-ratio)!  It's not really necessary since you've defined both a `width` and a `height` but it's not hurting anything so 🤷 .

client/modules/IDE/components/FloatingActionButton.jsx
```diff
@@ -0,0 +1,67 @@
...
+  border-radius: 99999px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  box-shadow: rgba(0, 0, 0, 0.24) 0px 3px 8px;
+  /* scale: 0.75; */
+  /* padding-left: 1.8rem; */
```
lindapaiste: Let's delete these commented-out lines before merging.

client/modules/IDE/components/FloatingActionButton.jsx
```diff
@@ -0,0 +1,67 @@
...
+    ${prop('Button.primary.default')}
+    g {
+      fill: ${prop('Button.primary.default.foreground')};
+    }
+  }
+  > svg {
+    width: 35%;
+    height: 35%;
+    > g {
+      fill: white;
```
lindapaiste: This should probably be a theme property?

client/modules/IDE/components/FloatingActionButton.jsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import styled from 'styled-components';
+import { useDispatch, useSelector } from 'react-redux';
+import PropTypes from 'prop-types';
+import PlayIcon from '../../../images/triangle-arrow-right.svg';
+import StopIcon from '../../../images/stop.svg';
+import { prop, remSize } from '../../../theme';
+import { startSketch, stopSketch } from '../actions/ide';
+
+const Button = styled.div`
```
lindapaiste: Why `styled.div` and not `styled.button`?

client/modules/IDE/components/FloatingActionButton.jsx
lindapaiste, line 1: Just want to note some of the ideas that we've tossed around regarding this element, potentially for future refactorings but not for right now:
- We could export the `styled` part as a reusable `FloatingActionButton` UI component, and make this specific `StartStopButton` be an instance of that common `FloatingActionButton`. (This fits with best practices, but is kind of silly since there are no other floating action buttons anywhere else in the app.)
- We could extract the existing `StartStopButton` from the desktop toolbar into its own component (see #2239).  That component would need to take a `className` prop in order to be stylable with `styled-components`. This floating version for mobile could be a `styled(StartStopButton)`, where we only need to specify the styles to override like the position and size.  The positioning could also potentially be handled by a separate `div` which has the button as a child.

lindapaiste (changes requested):

client/modules/IDE/components/FloatingActionButton.jsx
dewanshDT: yeah exactly that's what I thought, but let's keep it for now we'll remove it when we decide to replace the Icon with a better one.

client/modules/IDE/components/FloatingActionButton.jsx
dewanshDT: yeah 😂😂

client/modules/IDE/components/FloatingActionButton.jsx
dewanshDT: got it!

client/modules/IDE/components/FloatingActionButton.jsx
dewanshDT: yeah I agree

client/modules/IDE/components/FloatingActionButton.jsx
dewanshDT: got it!

client/modules/IDE/components/FloatingActionButton.jsx
dewanshDT: Yeah, the `StartStopButton` is a great idea and since we do not have any more floating action buttons anywhere else in the app but when I made it I thought it would be reusable and we should definitely export the styled part as a floating action button and the start-stop button should be passed in as a child element. let me know when we want to do it

dewanshDT:
hey @lindapaiste, made the suggested changes, check it out!

client/modules/IDE/components/FloatingActionButton.jsx
```diff
@@ -0,0 +1,65 @@
...
+  padding: 0;
+  ${prop('Button.secondary.default')};
+  aspect-ratio: 1/1;
+  border-radius: 99999px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  box-shadow: rgba(0, 0, 0, 0.24) 0px 3px 8px;
+  &.stop {
+    ${prop('Button.primary.default')}
```
lindapaiste, line 27: we've got a color here with no property!

client/modules/IDE/components/FloatingActionButton.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+const Button = styled.button`
+  position: fixed;
+  right: ${remSize(20)};
+  bottom: ${remSize(20)};
+  height: ${remSize(60)};
+  width: ${remSize(60)};
+  z-index: 3;
+  padding: 0;
+  ${prop('Button.secondary.default')};
```
lindapaiste, line 19: Another variable that's not assigned to any property.

lindapaiste (changes requested):
Looking good!

Two small changes and then let's merge this:
1. Delete the two lines I noted which have a `${prop(` that isn't assigned to any property.
2. I know that hover isn't that important on small-screen but let's put a more obvious hover effect on the "stop" button for when it's on a small laptop or whatever.  I'm thinking we change the background to pink.  Might be as simple as changing `&.stop` to `&.stop:not(:hover)`.

lindapaiste (approved):

PR #2350, opened 2023-08-04
New nav component and a new hook

dewanshDT:
Changes:
added the new nav component, which was broken into several smaller components and a new hook useSketchActions


client/modules/IDE/components/Header/Nav.jsx
```diff
@@ -1,414 +1,318 @@
...
-    this.props.cloneProject();
+const LeftLayout = (props) => {
+  switch (props.layout) {
+    case 'dashboard':
+      return <DashboardMenu />;
+    case 'project':
+    default:
+      return <ProjectMenu />;
   }
+};
```
lindapaiste, line 55: I still think that in the future we may want to break it out into two separate components for the different Nav layouts.  But that's not a today change.

client/modules/IDE/components/Header/Nav.jsx
```diff
@@ -1,414 +1,318 @@
...
-            />
-            <span className="nav__item-header">
-              {this.props.t('Nav.BackEditor')}
-            </span>
-          </Link>
-        </li>
-      </ul>
-    );
+  const isLoginEnabled = getConfig('LOGIN_ENABLED');
+  const isAuthenticated = user.authenticated;
```
lindapaiste: Since we only need the `user.authenticated` property that should be all that we select from Redux.
```
const isAuthenticated = useSelector((state) => state.user.authenticated);
```
or better yet
```
const isAuthenticated = useSelector(getAuthenticated);
```

client/modules/IDE/components/Header/Nav.jsx
```diff
@@ -1,414 +1,318 @@
...
-          <NavMenuItem key={key} value={key} onClick={this.handleLangSelection}>
-            {languageKeyToLabel(key)}
-          </NavMenuItem>
-        ))}
-      </NavDropdownMenu>
-    );
-  }
+const ProjectMenu = (props) => {
+  const isUserOwner = useSelector(getIsUserOwner);
+  const project = useSelector((state) => state?.project);
```
lindapaiste: `state` is always defined so there's no need for `?.`.  `state.project` is always an object, but it [doesn't have all of the properties](https://github.com/processing/p5.js-web-editor/blob/b22cc0a8d4fc7d36ab2e4ce6730d9b19a4863fa0/client/modules/IDE/reducers/project.js#L8-L12) when it's not saved.

client/modules/IDE/components/Header/Nav.jsx
```diff
@@ -1,414 +1,318 @@
...
+        <NavMenuItem
+          hideIf={!project?.id || !user.authenticated}
+          onClick={() => dispatch(cloneProject())}
+        >
+          {t('Nav.File.Duplicate')}
+        </NavMenuItem>
+        <NavMenuItem hideIf={!project?.id} onClick={shareSketch}>
+          {t('Nav.File.Share')}
+        </NavMenuItem>
+        <NavMenuItem hideIf={!project?.id} onClick={downloadSketch}>
```
lindapaiste: Just a suggestion but it might be more clear and readable if you add a variable `const isUnsaved = !project.id`.  You use this same `hideIf={!project?.id}` in a few places and maybe the intention would be more obvious if it was `hideIf={isUnsaved}`.

client/modules/IDE/components/Header/Nav.jsx
```diff
@@ -1,414 +1,318 @@
...
-  }).isRequired,
-  t: PropTypes.func.isRequired,
-  setLanguage: PropTypes.func.isRequired,
-  language: PropTypes.string.isRequired,
-  isUserOwner: PropTypes.bool.isRequired,
-  editorLink: PropTypes.string
+  return (
+    <NavDropdownMenu id="lang" title={languageKeyToLabel(language)}>
+      {sortBy(availableLanguages).map((key) => (
+        // eslint-disable-next-line react/jsx-no-bind
```
lindapaiste, line 247: This `jsx-no-bind` rule is one that we break all the time so it def should be disabled for the whole app (I think it is?).  FYI, If you wanted to not break the rule then you would define `handleLangSelection` with a `useCallback` to memoize it, so that the function is not recreated on each render.  But I would probably leave it as-is to keep the code from becoming confusing and because re-rendering `NavMenuItem` is trivial.

client/modules/IDE/components/Header/index.jsx
```diff
@@ -0,0 +1,34 @@
...
+          if (matches)
+            return (
+              <Toolbar
+                syncFileContent={props.syncFileContent}
+                key={project.id}
+              />
+            );
+          return null;
+        }}
+      </MediaQuery>
```
lindapaiste, line 25: Potentially you could leave out the `<MediaQuery>` here and keep the `<Toolbar>` always visible for now.  Then put in the `<MediaQuery>` when in the same PR as the mobile version of the `Nav`.  But let's leave it for now because I expect the mobile/responsive `Nav` to be merged soon enough that it's not a problem.

client/modules/IDE/components/Header/index.jsx
```diff
@@ -0,0 +1,34 @@
...
+    </header>
+  );
+};
+
+Header.propTypes = {
+  syncFileContent: PropTypes.func.isRequired
```
lindapaiste, line 31: In a future PR you can remove the `syncFileContent` prop and access it from your new context instead.

client/modules/IDE/hooks/index.js
```diff
@@ -0,0 +1,77 @@
...
+
+  function canEditProjectName() {
+    return (
+      (project.owner &&
+        project.owner.username &&
+        project.owner.username === currentUser) ||
+      !project.owner ||
+      !project.owner.username
+    );
+  }
```
lindapaiste: I created a selector function for this and I think that the logic is the same?  `true` if the sketch is unowned/unsaved or if the current user is the owner.  It's merged in so you can use it.
https://github.com/processing/p5.js-web-editor/blob/b22cc0a8d4fc7d36ab2e4ce6730d9b19a4863fa0/client/modules/IDE/selectors/users.js#L43-L47

client/modules/IDE/hooks/index.js
lindapaiste, line 1: You can change it later but I definitely think that this hook should be in a named file rather than an `index` file.  It's confusing because there are other files in this directory.  You can re-export it from the index and you get the same nice import paths by doing that.
`export * from './useSketchActions';`
or, if the hook is the default export
`export { default as useSketchActions } from './useSketchActions';`

package.json
```diff
@@ -213,7 +213,7 @@
     "passport-google-oauth20": "^1.0.0",
     "passport-http": "^0.3.0",
     "passport-local": "^1.0.0",
-    "prettier": "^2.2.1",
+    "prettier": "2.2.1",
```
lindapaiste, line 216: This will help with future headaches.

client/modules/IDE/components/Header/Nav.unit.test.jsx
```diff
@@ -1,62 +1,11 @@
...
-    setLanguage: jest.fn(),
-    language: 'en-US',
-    isUserOwner: true
-  };
-
-  it('renders correctly', () => {
-    const { asFragment } = render(<NavComponent {...props} />);
-    expect(asFragment()).toMatchSnapshot();
-  });
-
```
lindapaiste, line 59: It is sooo much cleaner with all those props gone! 🧹

client/modules/IDE/components/Header/Nav.jsx
```diff
@@ -1,414 +1,318 @@
...
-    user: state.user,
-    unsavedChanges: state.ide.unsavedChanges,
-    rootFile: state.files.filter((file) => file.name === 'root')[0],
-    language: state.preferences.language,
-    isUserOwner: getIsUserOwner(state),
-    editorLink: selectSketchPath(state)
-  };
-}
+const AuthenticatedUserMenu = () => {
+  const user = useSelector((state) => state.user);
```
lindapaiste: Seems like we don't need the whole `user` object here either.  Just the `username`.

client/modules/IDE/components/Header/Nav.jsx
```diff
@@ -1,414 +1,318 @@
...
+          aria-label={t('Common.p5logoARIA')}
+          focusable="false"
+          className="svg__logo"
+        />
+      </li>
+      <NavDropdownMenu id="file" title={t('Nav.File.Title')}>
+        <NavMenuItem onClick={newSketch}>{t('Nav.File.New')}</NavMenuItem>
+        <NavMenuItem
+          hideIf={
+            !getConfig('LOGIN_ENABLED') || (project?.owner && !isUserOwner)
```
lindapaiste, line 146: FYI that second condition is the negative of `canEditProjectName`/`selectCanEditSketch`.

lindapaiste (changes requested):
Excellent work!  There is only one thing which needs to change, and it's my fault 🤦‍♀️ because I told you that none of my Redux changes mattered...but I forgot about changing the toast from two actions to one.  I will come up with a patch for that which just changes the actions and doesn't involve Redux Toolkit.

lindapaiste:
Ok here's the fix I came up with for the Redux actions which will allow you to use `dispatch(showToast('Toast.SketchSaved'));` here without refactoring other components and actions which still use two separate actions.  Replace the existing `showToast` function:
https://github.com/processing/p5.js-web-editor/blob/b22cc0a8d4fc7d36ab2e4ce6730d9b19a4863fa0/client/modules/IDE/actions/toast.js#L9-L16

with this:
```
/**
 * Temporary fix until #2206 is merged.
 * Supports legacy two-action syntax:
 *    dispatch(setToastText('Toast.SketchFailedSave'));
 *    dispatch(showToast(1500));
 * And also supports proposed single-action syntax with message and optional timeout.
 *    dispatch(showToast('Toast.SketchFailedSave'));
 *    dispatch(showToast('Toast.SketchSaved', 5500));
 */
export function showToast(textOrTime, timeout = 1500) {
  return (dispatch) => {
    let time = timeout;
    if (typeof textOrTime === 'string') {
      // eslint-disable-next-line no-use-before-define
      dispatch(setToastText(textOrTime));
    } else {
      time = textOrTime;
    }
    dispatch({
      type: ActionTypes.SHOW_TOAST
    });
    setTimeout(() => dispatch(hideToast()), time);
  };
}
```

client/modules/IDE/components/Header/Nav.jsx
```diff
@@ -1,414 +1,318 @@
...
+          }
+          onClick={() => saveSketch(cmRef)}
```
lindapaiste: We've got a problem here!
image
Your `saveSketch` function expects to be called with the `cmContoller` object -- not the ref.  So you need to add a `.current` either here (`saveSketch(cmRef.current)`) or in the `saveSketch` function.

dewanshDT: Got it!

dewanshDT:
> Ok here's the fix I came up with for the Redux actions which will allow you to use `dispatch(showToast('Toast.SketchSaved'));` here without refactoring other components and actions which still use two separate actions.  Replace the existing `showToast` function:
> https://github.com/processing/p5.js-web-editor/blob/b22cc0a8d4fc7d36ab2e4ce6730d9b19a4863fa0/client/modules/IDE/actions/toast.js#L9-L16
> 
> with this:
> ```
> /**
>  * Temporary fix until #2206 is merged.
>  * Supports legacy two-action syntax:
>  *    dispatch(setToastText('Toast.SketchFailedSave'));
>  *    dispatch(showToast(1500));
>  * And also supports proposed single-action syntax with message and optional timeout.
>  *    dispatch(showToast('Toast.SketchFailedSave'));
>  *    dispatch(showToast('Toast.SketchSaved', 5500));
>  */
> export function showToast(textOrTime, timeout = 1500) {
>   return (dispatch) => {
>     let time = timeout;
>     if (typeof textOrTime === 'string') {
>       // eslint-disable-next-line no-use-before-define
>       dispatch(setToastText(textOrTime));
>     } else {
>       time = textOrTime;
>     }
>     dispatch({
>       type: ActionTypes.SHOW_TOAST
>     });
>     setTimeout(() => dispatch(hideToast()), time);
>   };
> }
> ```

Okay 👍

client/modules/IDE/components/Header/Nav.jsx
dewanshDT: we'll do that for sure!

client/modules/IDE/components/Header/Nav.jsx
dewanshDT: okay

client/modules/IDE/components/Header/Nav.jsx
dewanshDT: what should we do then

client/modules/IDE/components/Header/index.jsx
dewanshDT: yeah we should leave it as is for now

client/modules/IDE/components/Header/index.jsx
dewanshDT: yeah that would be a great Idea!

client/modules/IDE/hooks/index.js
dewanshDT: got it so I should remove it from here?

dewanshDT: yeah since it is being used in the toolbar pr I'll remove it in that PR

client/modules/IDE/hooks/index.js
dewanshDT: okay I'll create a separate file for that

package.json
dewanshDT: yeah 😮‍💨

client/modules/IDE/components/Header/Nav.unit.test.jsx
dewanshDT: yes it is 😂

dewanshDT:
hey @lindapaiste! just made all the suggested changes to the PR please check it out.

client/modules/IDE/components/Header/Nav.jsx
lindapaiste: Do nothing, you're good 😄 .  We have to weigh performance gains against keeping the codebase beginner-friendly.

lindapaiste (approved):

PR #2306, opened 2023-07-16
Fix errors sending emails (#2303)

lindapaiste:
Fixes #2303 
(Probably, but I haven't checked it with a valid Mailgun config to make sure that it works all the way through to sending the email.)

We merged a dependabot PR #2267 which included a major version bump of `mjml` and that caused emails to break.

Changes:

- Import `mjml2html` as the default from `mjml` instead of a named import.
- Remove deprecated component `<mj-container>` from templates.
- Convert unitless value `width="192"` to px `width="192px"`.
- Log errors when calling `mjml2html` (useful in development, probably does nothing in production but idk).

I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] has no test errors (`npm run test`)
* [x] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #2361, opened 2023-08-07
`MobileNav` component for smaller devices

dewanshDT:
> This PR is a part of the mobile responsive project

Changes:
added a `MobileNav` component to be used in smaller devices and, modified the `Nav` component to use the component in smaller screens.


client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,388 @@
...
+      overflow: hidden;
+      z-index: 10;
+      right: 10px;
+      transform: translateX(${remSize(6)});
+      width: max-content;
+      transform: scale(0);
+      opacity: 0;
+      transform-origin: top;
+      transition: opacity 100ms;
+      transition: transform 100ms 200ms;
```
lindapaiste: I think if you have two transition lines that the second overrides the first?  It should be one line with both transitions in it.

The transitions are really nice though!  Good work ⭐

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,388 @@
...
+      display: flex;
+      flex-direction: column;
+      gap: ${remSize(2)};
+      box-shadow: rgba(60, 64, 67, 0.3) 0px 1px 2px 0px,
+        rgba(60, 64, 67, 0.15) 0px 1px 3px 1px;
+      min-width: ${remSize(120)};
+      border-radius: ${remSize(2)};
+
+      /* User */
+      li.user {
```
lindapaiste, line 115: Nice specificity!

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,388 @@
...
+        display: flex;
+        width: 100%;
+        > button {
+          width: 100%;
+          padding: ${remSize(8)} ${remSize(15)} ${remSize(8)} ${remSize(10)};
+          font-size: ${remSize(18)};
+          text-align: left;
+          &:hover {
+            background-color: ${prop('Button.primary.hover.background')};
+            color: ${prop('Button.primary.hover.foreground')};
```
lindapaiste, line 150: If this style is very similar to the primary Button then it might possibly make sense to have these buttons be an extension of it. `styled(Button)`.  You can style components which are already styled, so long as they pass through a `className` prop.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,388 @@
...
+      case '/p5/collections':
+        return 'Examples';
+      case `/${user.username}/assets`:
+      case `/${user.username}/collections`:
+      case `/${user.username}/sketches`:
+        return 'My Stuff';
+      default:
+        return project.name;
+    }
+  }
```
lindapaiste, line 224: I don’t love this switch based on location.  It feels a bit fragile, like if we were to add a new page or change a path or something we would have to change it here as well.  Perhaps the alternative is that each page which includes the `Nav` specifies a `mobileTitle` prop which we pass down.  Like `<Nav layout=“dashboard” mobileTitle={t(‘Login’)} />`.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,388 @@
...
+
+  useEffect(() => {
+    setTitle(resolveTitle(pathname));
+  }, [pathname, project]);
```
lindapaiste: The title is a derived value so it should be a `useMemo` or just a plain variable, rather than a state set via `useEffect`.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,388 @@
...
+          <h5>by {project?.owner?.username}</h5>
+        )}
+      </Title>
+      <Options>
+        {user.authenticated ? (
+          <AccoutnMenu />
+        ) : (
+          <div>
+            <IconButton
+              onClick={() => {
```
lindapaiste: I know we’ve discussed this but these should be links.  👊💀🐴

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,388 @@
...
+  const project = useSelector((state) => state.project);
+  const user = useSelector((state) => state.user);
+  const [title, setTitle] = useState();
+
+  // use the useNavigate hook in the react router v6
+  const history = useHistory();
+  const navigate = (url) => history.push(url);
+
+  const { pathname } = useLocation();
+  const editorLink = useSelector(selectSketchPath);
```
lindapaiste, line 201: Isn’t it nice when we have defined selector functions?  Makes it much easier to access the same state in multiple places.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,388 @@
...
+              icon={EditorIcon}
+            />
+          </div>
+        )}
+      </Options>
+    </Nav>
+  );
+};
+
+const AccoutnMenu = () => {
```
lindapaiste: You’ve got a typo in the name here.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,388 @@
...
+            onClick={() => {
+              dispatch(openPreferences());
+            }}
+          >
+            Preferences
+          </button>
+        </li>
+        <li>
+          <button>Language</button>
+        </li>
```
lindapaiste: What are you picturing for the language selection?  Probably a question for @shujuuu.  We could open a modal but it might be cool if we can use a native `<select>` inline with the menu so that clicking on the language brings up the phone’s standard UI for selecting from dropdown menus.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,388 @@
...
+        <b>{t('Nav.Help.Title')}</b>
+        <li>
+          <button>{t('Nav.Help.KeyboardShortcuts')}</button>
```
lindapaiste: I think the one is just a `dispatch` to open the modal.

However — if there’s a way to hide this menu item on touch devices that would be super cool!  It would be a media query that’s specifically for whether it’s touch or keyboard, not based on the width.  The might have a keyboard plugged in to a small tablet, for example.

That’s for the future.  For now open the modal.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,388 @@
...
+  );
+};
+
+MoreMenu.propTypes = {
+  cmController: PropTypes.shape({
+    tidyCode: PropTypes.func,
+    showFind: PropTypes.func,
+    showReplace: PropTypes.func,
+    getContent: PropTypes.func
+  })
```
lindapaiste: We can use your context here!  Either now or later.

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -260,12 +261,11 @@ class IDEView extends React.Component {
         <WarnIfUnsavedChanges currentLocation={this.props.location} />
         <Toast />
         <CmControllerContext.Provider value={{ current: this.cmController }}>
-          <Nav />
+          <Header
+            syncFileContent={this.syncFileContent}
+            key={this.props.project.id}
```
lindapaiste, line 267: I wonder if this `key` prop actually does anything?  I see it’s part of the existing code.  It doesn’t hurt to leave it but I strongly suspect that it’s useless.

client/theme.js
```diff
@@ -74,11 +74,12 @@ const baseThemes = {
     primaryTextColor: grays.dark,
     inactiveTextColor: grays.middleDark,
     backgroundColor: grays.lighter,
+    accentColor: colors.p5jsPink,
 
     Button: {
       primary: {
         default: {
-          foreground: colors.black,
+          foreground: grays.mediumDark,
```
lindapaiste, line 82: This change will have implications on all button text colors throughout the app.  Is that what we want?  @shujuuu  Probably yes but I want to double check.

client/theme.js
```diff
@@ -130,6 +131,10 @@ const baseThemes = {
         foreground: colors.black,
         background: grays.light,
         border: grays.middleLight
+      },
+      secondary: {
+        foreground: colors.black,
+        background: grays.mediumLight
```
lindapaiste, line 137: If `secondary` is its own complete object then does it need a `border` color?

If the only difference between `primary` and `secondary` (and potential `tertiary`) is the background color and the rest is the same then maybe we structure it a bit differently.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,388 @@
...
+
+  const { pathname } = useLocation();
+  const editorLink = useSelector(selectSketchPath);
+
+  function resolveTitle(path) {
+    switch (path) {
+      case '/':
+        return project.name;
+      case '/login':
+        return 'Login';
```
lindapaiste: I bet if you look in the English translations file that you can find translation keys for most of these title strings (except “My Stuff”).

I’m a fan of reusing existing keys/properties because they already have translations in many languages, rather than adding a new “MobileNav” section.

lindapaiste (changes requested):
This is coming along great!  IMO we should clean it up a bit before merging.

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: got it! replaced it with a single line, `transition: opacity 100ms, transform 100ms 200ms;`

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: yeah, but it just has same hover colors, rest is totally different so Idk if extending really makes sense, let me know

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: yeah that's better, will update it in the future for sure, added a TODO

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: yeah that's a great idea, memoized it!

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: yeah replaced all `navigate()` with `<Link>` 😁

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: yes it is ✨

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: got it!

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: yeah I'll ask her, will update this in future

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: okayy!

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: we're already using it, forgot to remove it from here, done!

client/modules/IDE/pages/IDEView.jsx
dewanshDT: should I remove it?

client/theme.js
dewanshDT: yeah this is should be discussed

client/theme.js
dewanshDT: in mobile Panel we don't have borders, I'll check though

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: yeah added the translations.

lindapaiste:
It seems like the "Find" command isn't working.  It's very strange because "Tidy Code" does work.  I confirmed that the `Editor` `showFind()` method is getting called.  It's the internal codemirror part (`this._cm.execCommand('findPersistent');`) which isn't doing anything.

This could be a real rabbit hole so I will look into it.  It seems like the modal is appearing but somehow it's inside of the nav bar and not visible.

image

Edit: turns out it's just a `z-index`/positioning issue.  Let's add a media query to the `_editor.scss` file to override the `top` property of the `.CodeMirror-dialog` at the same breakpoint where the `MobileNav` appears.  Maybe ~100px down from the top?  We need to work on the dialog itself but that's a future problem.  It will be usable on tablet sizes if we fix the `top` position.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,338 @@
...
+  );
+  const dispatch = useDispatch();
+  const { t } = useTranslation();
+  const { newSketch, saveSketch } = useSketchActions();
+
+  const cmRef = useContext(CmControllerContext);
+
+  return (
+    <div>
+      <IconButton icon={MoreIcon} />
```
lindapaiste: Can we add a `transform: rotate(90deg);` to this icon to make the dots vertical?

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,338 @@
...
+  margin-left: auto;
+  display: flex;
+  /* transform: translateX(${remSize(12)}); */
+  svg {
+    fill: ${(props) => props.color};
+  }
+
+  /* Drop Down menu */
+  > div > button:focus + ul,
+  > div > ul > button:focus ~ div > ul {
```
lindapaiste: It seems like you are using CSS-only to open and close the menus, which is very cool, but it might get in the way of implementing the behavior that we discussed where clicking on the * icon would also open the menu.

It also has impacts on accessibility, which, I don't really know what to make of that when we're dealing with mobile.  But a keyboard user won't be able to tab through the menu items.

So we probably need some sort of React state and then use the state to control an `opened` class on the menu list.  I need to think about how we can reuse the focus/blur logic from the `NavBar` here.

lindapaiste: I'm playing with using all the contexts and states from the `NavBar` and I've got it like 90% working but I'm tired and don't want to deal with Git stuff now so I'll send it to you tomorrow.

client/modules/IDE/components/Header/MobileNav.jsx
lindapaiste: `transform 100ms 200ms` -- that's a delay of 200ms and then 100ms for the scale transform.  So the opacity transition will have completed (from 0-100ms) before the transform starts (from 200-300ms).  When opening you will only see the transform and when closing you will only see the opacity.  Seems like that's probably not what you intended?  Maybe should be `100ms` for both with no delay?  If we want different transitions for opening vs closing we could use `react-transition-group`.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,338 @@
...
+  return (
+    <div>
+      <IconButton icon={MoreIcon} />
+      <ul>
+        <b>{t('Nav.File.Title')}</b>
+        <li>
+          <button onClick={newSketch}>{t('Nav.File.New')}</button>
+        </li>
+        <li>
+          <button onClick={() => saveSketch(cmRef.current)}>
```
lindapaiste: On the desktop nav we have some conditions to hide the "Save" button if the user is not logged in, so we should probably do that here.

Although the usability for clicking it when logged out is pretty decent, it opens a modal telling me to log in.

lindapaiste:
A couple minor style things.  Probably to save for later.

- The `box-shadow` seems like a lot.  Do we want just a `border-bottom`?

- The account icon could be a little bit smaller.

- The top of the menu should probably be aligned with the bottom of the nav bar instead of overlapping it.

- We should think about how the user gets to the bottom of the menu when on a short screen (a phone in landscape mode).  Right now it will scroll the whole page down.  We may want a `max-height` on the menu of `calc(100vh - barHeightpx)` and an `overflow: auto` so that the menu can scroll independent of the page.

- If it's easy, we should hide the account menu when on the login page because clicking it won't do anything.

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: I updated the svg

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: when I implemented it I thought most of the users won't use the keyboard on a mobile device and implemented it like this, but what you did is great and we should definitely include that

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: I like that we're showing the modal to log in instead of hiding the save button when the user is not logged in, what do you think?

dewanshDT:
We should definitely make all these changes, let's get your PR merged in first

client/modules/IDE/components/Header/MobileNav.jsx
lindapaiste: Yeah I kind of like it too 🤷 Let's keep it for now but I don't like having different logic on mobile vs desktop.  When you get a chance maybe you can open an issue for discussion on how we should handle this.  Like whether we should show the "Save" option on desktop and have it open the modal.

We could also def make that modal nicer.  Emphasize the "Log In" action (maybe a big button?) and use a friendlier title than "Error".

dewanshDT: yeah I agree In the future we should make it nicer by adding a welcoming message with a login button to continue

client/modules/IDE/components/Header/MobileNav.jsx
lindapaiste: Ok in that case it does not make sense.

I'm still figuring out how to best use styled-components in a clean and composable way.  Like what's the equivalent of the `@mixin` structure in the current SCSS, so we could do things like apply a "button hover" mixin.

dewanshDT: yeah

client/modules/IDE/components/Header/MobileNav.jsx
lindapaiste: Probably the quickest/easiest way to get something that's usable would be to open a modal.  No Redux state, use a `languageModalOpen` component state in `MoreMenu` and conditionally render the modal if it's true.  Not sure about the content of the modal but maybe just a native `<select>` for right now even if it's ugly.  This wouldn't be the final product but it would allow us to go live with the `MobileNav` without creating any usability issues.  We just need to make sure that there is some way to select a language.

IMO this and the "Find" search box are the only things that we need to address right now; and everything else is just a "nice to have" which we can revisit later.

dewanshDT:
@lindapaiste, I just added an overlay to select language, please have a look, it also indicates the selected language.

image

dewanshDT:
hey @lindapaiste, added the use of ProjectName, please check it out.

image

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,393 @@
...
+                  className={classNames({
+                    'current-language':
+                      languageKeyToLabel(key) === languageKeyToLabel(language)
+                  })}
+                  aria-label={languageKeyToLabel(key)}
+                  key={key}
+                  onClick={handleLangSelection}
+                  value={key}
+                >
+                  {languageKeyToLabel(key)}
```
lindapaiste: Since we have `languageKeyToLabel(key)` three times, it might be worth it to define it outside of the return.  Up to you.

```
{sortBy(availableLanguages).map((key) => {
  const label = languageKeyToLabel(key);
  return (
    <button
```

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,393 @@
...
+          ariaLabel="Select Language"
+          closeOverlay={() => setIsLanguageModalVisible(false)}
+        >
+          <LanguageSelect>
+            <div>
+              {sortBy(availableLanguages).map((key) => (
+                <button
+                  className={classNames({
+                    'current-language':
+                      languageKeyToLabel(key) === languageKeyToLabel(language)
```
lindapaiste: This should probably be `key === language`, though what you have works.  Compare on the key/value ("en-US") rather than the label.

client/modules/IDE/components/Header/Toolbar.jsx
```diff
@@ -35,34 +30,6 @@ const Toolbar = (props) => {
...
-      dispatch(showEditProjectName());
-      setTimeout(() => {
-        projectNameInputRef.current?.focus();
-      }, 140);
-    }
-  }
-
-  function handleProjectNameSave() {
-    const newName = nameInputValue;
-    if (newName.length > 0) {
```
lindapaiste, line 61: FYI it will now use the 0-length behavior from the `EditableInput`, which is to close the input and keep the name from before.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,393 @@
...
+    button {
+      font-size: ${remSize(16)};
+      padding: ${remSize(8)} ${remSize(15)} ${remSize(8)} ${remSize(10)};
+      width: 100%;
+      max-width: 80vw;
+
+      &:hover,
+      &:active,
+      &.current-language,
+      &:focus {
```
lindapaiste, line 186: Very nice how you've applied the same style to all of these selectors.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,393 @@
...
+  h3 {
+    text-align: center;
+    margin-bottom: 2rem;
+  }
+
+  div {
+    display: flex;
+    gap: ${remSize(8)};
+    flex-direction: column;
+    position: block;
```
lindapaiste: You should delete `position: block;` as it's invalid.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,393 @@
...
+  justify-content: center;
+
+  h3 {
+    text-align: center;
+    margin-bottom: 2rem;
+  }
+
+  div {
+    display: flex;
+    gap: ${remSize(8)};
```
lindapaiste, line 172: I always forget that you can use `gap` with flexbox.  Very nice!

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,393 @@
...
+      }
+    }
+  }
+`;
+
+const LanguageSelect = styled.div`
+  display: flex;
+  flex-direction: column;
+  position: relative;
+  height: 100%;
```
lindapaiste, line 162: The positioning is great if the screen is big enough to fit the whole list.  We need an `overflow-y: auto` somewhere if the list doesn't fit (like in portrait mode of iPhone SE, for example).  I'm playing now and it seems like adding `overflow-y: auto;` on this div and `max-height: 100%; overflow-y: auto;` on the inner div will do it?

lindapaiste (changes requested):

lindapaiste:
> hey @lindapaiste, added the use of ProjectName, please check it out.
> 
> image

Looking just like the mockups!

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: yeah

client/modules/IDE/components/Header/Toolbar.jsx
dewanshDT: is it okay to leave it like this?

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: yess!

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: just `overflow-y: auto;` worked

client/modules/IDE/components/Header/Toolbar.jsx
lindapaiste: yes.

client/modules/IDE/components/Header/MobileNav.jsx
lindapaiste: For me at least that doesn't work.  The list is still centered so there's a hidden top part of it that is not reachable.
image
I see "Italiano" first but I should see "Deutsch".

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -0,0 +1,395 @@
...
+          <LanguageSelect>
+            <div>
+              {sortBy(availableLanguages).map((key) => {
+                const label = languageKeyToLabel(key);
+                return (
+                  <button
+                    className={classNames({
+                      'current-language': label === languageKeyToLabel(language)
+                    })}
+                    aria-label={label}
```
lindapaiste, line 330: I need to read up on ARIA best practices because I don't know if it's necessary to have an `aria-label` when it's the same as the text content of the button.  But we'll leave it for now.

lindapaiste (approved):
You did it! 🥳 🎊 We've fixed all of the major usability issues and I think it's ready to go!  This will be such a huge improvement for our mobile users.  Next up...the IDE screen.

lindapaiste:
@raclim We've been through a lot of revisions on this PR and we've gotten it to a point where it makes a lot of things better and doesn't make anything worse.  There is still a lot to be improved, but it's ready to release!

raclim:
This is super amazing!!! It looks really awesome so far and I think it'll definitely gain a lot of traction from users over the next few weeks, especially as students start going back to school! Thanks again to you both for all of your work! :) 

I'll try to merge in #2306, #2052, and #2362 as well before starting the process for the next release, and will give an update when that's happening. Please let me know if you both feel otherwise or would prefer to prioritize any others!

PR #2374, opened 2023-08-12
Changed promises to async/await in addProjectToCollection.js

adadgarg:
Changes:
Changed the promises to async/await syntax in the addProjectToCollection.js file.

### DIFF
--- a/server/controllers/collection.controller/addProjectToCollection.js
+++ b/server/controllers/collection.controller/addProjectToCollection.js
@@ -1,15 +1,9 @@
 import Collection from '../../models/collection';
 import Project from '../../models/project';
 
-export default function addProjectToCollection(req, res) {
+export default async function addProjectToCollection(req, res) {
   const owner = req.user._id;
   const { id: collectionId, projectId } = req.params;
-
-  const collectionPromise = Collection.findById(collectionId).populate(
-    'items.project',
-    '_id'
-  );
-  const projectPromise = Project.findById(projectId);
 
   function sendFailure(code, message) {
     res.status(code).json({ success: false, message });
@@ -46,7 +40,6 @@ export default function addProjectToCollection(req, res) {
 
     try {
       collection.items.push({ project });
-
       return collection.save();
     } catch (error) {
       console.error(error);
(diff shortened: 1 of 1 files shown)


server/controllers/collection.controller/addProjectToCollection.js
```diff
@@ -69,9 +62,19 @@ export default function addProjectToCollection(req, res) {
...
-    .catch(sendFailure);
+  try {
+    const collection = await Collection.findById(collectionId).populate(
+      'items.project',
+      '_id'
+    );
+    const project = await Project.findById(projectId);
+
+    const updatedCollection = await updateCollection([collection, project]);
+    if (updatedCollection) {
```
lindapaiste, line 73: What if there’s not an updated collection?  Do we need an `else` here?  We probably also want 404 errors if either the collection or the project are not found.  But I need to take a closer look at this and see what those other functions are doing.

adadgarg: we can just send failure message and then return 

```
  if (!updatedCollection) {
    sendFailure(404, 'Collection not updated');
    return;
  }
```

adadgarg: for collection and project we can also do the same

```
if (!collection) {
      sendFailure(404, 'Collection not found');
      return;
}
if (!project) {
  sendFailure(404, 'Project not found');
  return;
}
```

lindapaiste: Ah ok now I'm looking at this properly and I see that the things I was concerned about are already handled in other places.

The not found errors are in handled inside the `updateCollection` function.  https://github.com/processing/p5.js-web-editor/blob/6f5eb26a3afd29615c15be47f5a04c20ae81bed0/server/controllers/collection.controller/addProjectToCollection.js#L17-L25

If `updateCollection` returns `null` then it will have also called `sendFailure` https://github.com/processing/p5.js-web-editor/blob/6f5eb26a3afd29615c15be47f5a04c20ae81bed0/server/controllers/collection.controller/addProjectToCollection.js#L44-L48

server/controllers/collection.controller/addProjectToCollection.js
```diff
@@ -69,9 +62,19 @@ export default function addProjectToCollection(req, res) {
...
-    .then(updateCollection)
-    .then(populateReferences)
-    .then(sendSuccess)
-    .catch(sendFailure);
+  try {
+    const collection = await Collection.findById(collectionId).populate(
+      'items.project',
+      '_id'
+    );
+    const project = await Project.findById(projectId);
```
lindapaiste, line 70: There's actually a slight performance benefit to the `Promise.all` in the previous code because it allows the two "findById" functions to be executed at the same time instead of one and then the other.

raclim:
Since it's been a while since this PR has been updated, I've marked it with a "closing soon" label, which signifies that this PR will be closed in 10 days. Please feel free to revisit this to make any further updates or indicate that you would like to keep this open!

PR #2377, opened 2023-08-13
Modal tests

ofhope:
Fixes #issue-number

This PR adds storybook documentation and unit tests to the modal component.

Screenshot 2023-08-11 at 6 06 11 am

It also looks like a bable upgrade along the way broke storybook hot module reloading.
I didn't spend a lot of time investigating the cause but did find[ a solution](https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/176) to monkey patching the issue. HMR was still working in storybook but un-viewable without this patch.

Screenshot 2023-08-11 at 6 05 10 am

Changes:

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -1,0 +1,5 @@
+<script>
+    // https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/176#issuecomment-683150213
+    window.$RefreshReg$ = () => {};
+    window.$RefreshSig$ = () => () => {};
+</script>
--- a/client/modules/IDE/components/Modal.stories.jsx
+++ b/client/modules/IDE/components/Modal.stories.jsx
@@ -1,0 +1,17 @@
+import React from 'react';
+import Modal from './Modal';
+
+export default {
+  title: 'IDE/Modal',
+  component: Modal,
+  argTypes: {
+    onClose: { action: 'onClose' }
+  }
+};
+
+export const ModalDefault = {
+  args: {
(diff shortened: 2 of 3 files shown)


lindapaiste:
> It also looks like a bable upgrade along the way broke storybook hot module reloading. I didn't spend a lot of time investigating the cause but did find[ a solution](https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/176) to monkey patching the issue. HMR was still working in storybook but un-viewable without this patch.

THANK YOU!  I've been seeing that error but hadn't had a chance to look into it yet.

lindapaiste (approved):
Looks great!  I love the tests for the inside and outside click behavior and I so grateful for the `window.$RefreshReg$` fix.

lindapaiste:
It's a bummer that it doesn't render with the proper styling.  There's two issues there, one of which I only just now realized!  Maybe you can fix it in a separate PR.
1. The modal needs to have a parent with `class="light"` (or "dark" or "contrast").  This will be fixed by #2326, but we can also put in a quick fix until that gets merged since it might be a while.  We can add a `<div className="light">` to the `decorators` in `preview.js`.
2. The compiled `main.css` file that we are importing in `preview.js` is not up to date.  I don't know what that file is for and probably it should be deleted.  We want to import the uncompiled file `import '../client/styles/main.scss'` as that will be current will all changes.  But in order to do that we also need a Storybook addon to process it: https://storybook.js.org/addons/storybook-addon-sass-postcss

PR #2362, opened 2023-08-07
Move `IDEOverlays` into a separate component

lindapaiste:
Progress on #1458

Merge after #2335 

Changes:
- Move the section of the code for rendering modals and overlays out of `IDEView` and into a new component `IDEOverlays`, as discussed in #1458 

I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] has no test errors (`npm run test`)
* [x] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #2387, opened 2023-08-17
Useable Mobile Editor 🎉

dewanshDT:
Fixes #issue-number

Changes:

The useable editor is now available in mobile!!
image

image

image

before merging this, following PRs need to be merged
#2052 
#2347 
#2362 


client/components/RootPage.jsx
```diff
@@ -4,10 +4,16 @@ import { prop } from '../theme';
 const RootPage = styled.div`
   min-height: 100%;
   display: flex;
+  justify-content: start;
```
lindapaiste, line 7: Does this make a difference?  I think it should be ‘flex-start’ and that’s the default value so it probably doesn’t change anything?

client/modules/IDE/components/Editor/MobileEditor.jsx
```diff
@@ -0,0 +1,90 @@
...
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  padding-bottom: 5rem;
+  transform: ${(props) =>
+    props.expanded ? 'translateX(50%)' : 'translateX(0)'};
+
+  > header {
+    display: flex;
+    ${prop('MobilePanel.secondary')}
```
lindapaiste, line 14: Looks like we’ve got a value with no property?  Here and on line 21.

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -197,6 +201,16 @@ class Editor extends React.Component {
...
+      if (
+        ((metaKey === 'Cmd' && e.metaKey) ||
+          (metaKey === 'Ctrl' && e.ctrlKey)) &&
+        e.shiftKey &&
+        e.key === 'f'
+      ) {
+        e.preventDefault();
+        this.tidyCode();
+      }
+
```
lindapaiste: The removal on line 164 and the addition here is due to a change made to the develop branch in another PR.  Yours should be the same as the current.

lindapaiste (changes requested):
Some first thoughts.  Will continue reviewing later.

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -428,122 +338,7 @@ IDEView.propTypes = {
...
-  setPreviousPath: PropTypes.func.isRequired,
-  showErrorModal: PropTypes.func.isRequired,
-  hideErrorModal: PropTypes.func.isRequired,
-  clearPersistedState: PropTypes.func.isRequired,
-  startSketch: PropTypes.func.isRequired,
-  closeUploadFileModal: PropTypes.func.isRequired,
-  t: PropTypes.func.isRequired,
-  isUserOwner: PropTypes.bool.isRequired
+  }).isRequired
 };
```
lindapaiste: Look at all those props gone!  So satisfying.  You actually don’t need ANY props anymore because you can get the params and location from react-router hooks.

dewanshDT: Yess 😙

client/components/RootPage.jsx
dewanshDT: yes it does when the length of the editor increases

dewanshDT: sometimes the header goes missing it prevents it from that unexpected behaviour

client/modules/IDE/components/Editor/MobileEditor.jsx
dewanshDT: no this is to set the background color, it automatically does that

dewanshDT: we can also change all the `foreground` key in the theme to `color` to set the color and background automaticall by just writing this one line `${prop('some object background and color property')}`

lindapaiste: Right, I see what you’re saying.  You put the “color:” part in the theme object.  So it’s like a mixin rather than a color variable.

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -78,347 +79,256 @@ function WarnIfUnsavedChanges() {
...
 
-  componentDidMount() {
-    // If page doesn't reload after Sign In then we need
-    // to force cleared state to be cleared
-    this.props.clearPersistedState();
+  const [consoleSize, setConsoleSize] = useState(
+    ide.consoleIsExpanded ? 150 : 29
+  );
+  const [sidebarSize, setSidebarSize] = useState(
+    ide.sidebarIsExpanded ? 160 : 20
```
lindapaiste: A lot of my feedback here is going to be stuff that we already discussed on Discord.

You can simplify the handling of the size state by storing just the expanded size.  So `useState(160)`

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -78,347 +79,256 @@ function WarnIfUnsavedChanges() {
...
 
-    this.props.stopSketch();
-    if (this.props.params.project_id) {
-      const { project_id: id, username } = this.props.params;
-      if (id !== this.props.project.id) {
-        this.props.getProject(id, username);
-      }
-    }
+  const cmRef = useRef({});
+  let overlay = null;
```
lindapaiste: The overlay will need to be `useRef` as well.

Although…it seems like the current code is only using this variable to manipulate a style property on the element and I hate that.  You could use a state instead, like const [overlayHidden, setOverlayHidden] = useState(false);  And the on the div use style={{ display: overlayHidden ? ‘none’ : ‘block’ }} or something like that.

dewanshDT: Got it

client/modules/IDE/pages/IDEView.jsx
dewanshDT: Okay

client/modules/IDE/pages/IDEView.jsx
lindapaiste: The overlay will need to be `useRef` as well.

Although…it seems like the current code is only using this variable to manipulate a style property on the element and I hate that.  You could use a state instead, like const [overlayHidden, setOverlayHidden] = useState(false);  And the on the div use style={{ display: overlayHidden ? ‘none’ : ‘block’ }} or something like that.

dewanshDT: Yeah will do it

lindapaiste:
We discussed how a lot of the prevProps comparisons could be avoided if you handled each useEffect separately with the dependency on the prop you are trying to detect changes in.

dewanshDT:
yeah I remember that, will work on all the changes now

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -75,7 +80,7 @@ window.HTMLHint = HTMLHint;
 
 const INDENTATION_AMOUNT = 2;
 
-class Editor extends React.Component {
+class EditorV2 extends React.Component {
```
lindapaiste: Let's rename this back to `Editor` before merging.

dewanshDT: got it! done, check it out @lindapaiste

dewanshDT:
@lindapaiste Also, we're getting errors in integration testing, try running `npm run test`. How can we fix that?

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -77,265 +79,243 @@ function WarnIfUnsavedChanges() {
...
-                primary="second"
-                size={
-                  this.props.ide.consoleIsExpanded ? this.state.consoleSize : 29
-                }
-                minSize={29}
-                onChange={(size) => this.setState({ consoleSize: size })}
-                allowResize={this.props.ide.consoleIsExpanded}
-                className="editor-preview-subpanel"
+                split="vertical"
+                size={ide.sidebarIsExpanded ? sidebarSize : 20}
```
lindapaiste, line 148: Since you have the ternary here, you can delete the `useEffect` on the sidebar size.  You can also simplify the initial state to `  const [sidebarSize, setSidebarSize] = useState(160);`.  The `sidebarSize` state just needs to store the size of the expanded panel.  Then you use either the current expanded `sidebarSize` or the collapsed size of `20`.

Same for the `consoleSize`.

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -77,265 +79,243 @@ function WarnIfUnsavedChanges() {
...
-        )}
-      </RootPage>
-    );
-  }
-}
+                  <Console />
+                </SplitPane>
+              </PreviewWrapper>
+              <EditorSidebarWrapper show={!ide.isPlaying}>
+                <FileDrawer show={ide.sidebarIsExpanded}>
```
lindapaiste: I think it's weird that we have all of these lines (284-310) in the `IDEView` component for the mobile sidebar, when the corresponding code for desktop is in the `Sidebar` component.  We should work on combining the two sidebar codes.

We want the the same dropdown options (add file, add folder, upload file) on both versions.  The ➕ button that you designed for the mobile looks really nice and we can use that on the desktop version too.

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -77,265 +79,243 @@ function WarnIfUnsavedChanges() {
...
+    if (ide.sidebarIsExpanded) {
+      setSidebarSize((prev) => (prev < 160 ? 160 : sidebarSize));
+    }
+  }, [ide.sidebarIsExpanded]);
 
-  componentDidUpdate(prevProps) {
-    if (this.props.isUserOwner && this.props.project.id) {
+  // For autosave
+  useEffect(() => {
+    let autosaveTimeout;
```
lindapaiste: I find it confusing having both a local `autosaveTimeout` and a persisted `autosaveIntervalRef`.  It's probably better to remove the `let autosaveTimeout` and replace the usages of `autosaveTimeout` with `autosaveIntervalRef.current`.

I have to look at this bit closer and make sure that it's doing what it's supposed to.  Right now I'm 🤯

client/modules/IDE/pages/IDEView.jsx
dewanshDT: yeah

client/modules/IDE/pages/IDEView.jsx
dewanshDT: yeah!

dewanshDT:
hey @lindapaiste, Just made all the changes and push the code check it out!

client/modules/IDE/hooks/useKeyDownHandlers.js
```diff
@@ -36,7 +36,7 @@ export default function useKeyDownHandlers(keyHandlers) {
     } else if (isCtrl) {
       handlers.current[`ctrl-${e.key.toLowerCase()}`]?.(e);
     }
-    handlers.current[e.key.toLowerCase()]?.(e);
+    handlers.current[e.key?.toLowerCase()]?.(e);
```
lindapaiste, line 40: Thanks for catching my mistake here! I assumed that keyboard events would always have an `e.key` but they do not. 🤦‍♀️ 

The `?.` on the previous two seem to have gotten overwritten when merging (Rachel's `else if` change).  Instead let's put `if (!e.key) return;` as the first line of the `handleEvent`.

dewanshDT: got it!

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -66,296 +70,212 @@ function WarnIfUnsavedChanges() {
...
+  }, []);
+
+  // for setting previous location
+  useEffect(() => {
+    dispatch(setPreviousPath(locationRef.current.pathname));
+  }, [location.pathname]);
 
-  componentDidUpdate(prevProps) {
-    if (this.props.isUserOwner && this.props.project.id) {
+  // For autosave
```
lindapaiste: Let's define some of the complicated `boolean` logic outside of the `useEffect` so that we can minimize the dependencies that we need.

```suggestion
  const autosaveAllowed = isUserOwner && project.id && preferences.autosave;
  const shouldAutosave = autosaveAllowed && ide.unsavedChanges;

  // For autosave - send to API after 5 seconds without changes
```

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -66,296 +70,212 @@ function WarnIfUnsavedChanges() {
...
+  }, [
+    isUserOwner,
+    project.id,
+    preferences.autosave,
+    ide.unsavedChanges,
+    ide.justOpenedProject,
+    selectedFile.name,
+    dispatch,
+    autosaveProject
+  ]);
```
lindapaiste:
```suggestion
}, [shouldAutosave, dispatch]);
```

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -66,296 +70,212 @@ function WarnIfUnsavedChanges() {
...
+  useEffect(() => {
+    dispatch(setPreviousPath(locationRef.current.pathname));
+  }, [location.pathname]);
 
-  componentDidUpdate(prevProps) {
-    if (this.props.isUserOwner && this.props.project.id) {
+  // For autosave
+  useEffect(() => {
+    const handleAutosave = () => {
       if (
```
lindapaiste:
```suggestion
```

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -66,296 +70,212 @@ function WarnIfUnsavedChanges() {
...
-        this.props.preferences.autosave &&
-        this.props.ide.unsavedChanges &&
-        !this.props.ide.justOpenedProject
+        isUserOwner &&
+        project.id &&
+        preferences.autosave &&
+        ide.unsavedChanges &&
+        !ide.justOpenedProject &&
+        selectedFile.name === prevFileNameRef.current
```
lindapaiste: As we discussed, we can do the conditional checks before we start the timer.  So we don't need any `if` here.
```suggestion
```

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -66,296 +70,212 @@ function WarnIfUnsavedChanges() {
...
       ) {
```
lindapaiste:
```suggestion
```

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -66,296 +70,212 @@ function WarnIfUnsavedChanges() {
...
-            clearTimeout(this.autosaveInterval);
-          }
-          this.autosaveInterval = setTimeout(this.props.autosaveProject, 20000);
-        }
-      } else if (this.autosaveInterval && !this.props.preferences.autosave) {
-        clearTimeout(this.autosaveInterval);
-        this.autosaveInterval = null;
+        dispatch(autosaveProject());
+        console.log('saveee');
       }
```
lindapaiste:
```suggestion
```

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -66,296 +70,212 @@ function WarnIfUnsavedChanges() {
...
-          if (this.autosaveInterval) {
-            clearTimeout(this.autosaveInterval);
-          }
-          this.autosaveInterval = setTimeout(this.props.autosaveProject, 20000);
-        }
-      } else if (this.autosaveInterval && !this.props.preferences.autosave) {
-        clearTimeout(this.autosaveInterval);
-        this.autosaveInterval = null;
+        dispatch(autosaveProject());
+        console.log('saveee');
```
lindapaiste:
```suggestion
      dispatch(autosaveProject());
```

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -66,296 +70,212 @@ function WarnIfUnsavedChanges() {
...
-    clearTimeout(this.autosaveInterval);
-    this.autosaveInterval = null;
-  }
 
-  handleBeforeUnload = (e) => {
-    const confirmationMessage = this.props.t('Nav.WarningUnsavedChanges');
-    if (this.props.ide.unsavedChanges) {
-      (e || window.event).returnValue = confirmationMessage;
-      return confirmationMessage;
+    if (preferences.autosave) {
```
lindapaiste:
```suggestion
    if (shouldAutosave) {
```

lindapaiste (changes requested):

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -66,296 +70,191 @@ function WarnIfUnsavedChanges() {
...
+
+    dispatch(stopSketch());
+    if (props.params.project_id) {
+      const { project_id: id, username } = props.params;
+      if (id !== project.id) {
+        dispatch(getProject(id, username));
       }
     }
-  }
+  }, []);
```
lindapaiste: we discussed the need for some dependencies here.

lindapaiste (approved):

PR #2390, opened 2023-08-17
The My stuff Page in mobile 🧸

dewanshDT:
This is a part of the mobile editor project

Changes:

image
image
image


client/modules/IDE/components/CollectionList/CollectionListRow.jsx
```diff
@@ -228,16 +236,10 @@ class CollectionListRowBase extends React.Component {
             {this.renderCollectionName()}
           </span>
         </th>
+        <td>{formatDateCell(collection.createdAt, mobile)}</td>
+        <td>{formatDateCell(collection.updatedAt, mobile)}</td>
```
lindapaiste, line 240: I have concerns that showing two dates without labels is confusing to users.  But I know that you are implementing what was shown in the mockups.  Maybe something to ask @shujuuu?

client/modules/User/pages/DashboardView.jsx
```diff
@@ -151,7 +156,9 @@ class DashboardView extends React.Component {
           </div>
 
           <div className="dashboard-content">
-            {this.renderContent(currentTab, username)}
+            <MediaQuery maxWidth={770}>
+              {(mobile) => this.renderContent(currentTab, username, mobile)}
+            </MediaQuery>
```
lindapaiste, line 161: Nice 👍 I do kinda wonder if it makes more sense to determine `isMobile` closer to where we actually use it, rather than passing down a `mobile` prop through multiple components.  But this is fine for now.  Especially since the `SketchListRow` etc. are still function components and cannot use a `react-responsive` hook.

client/styles/base/_base.scss
```diff
@@ -2,41 +2,50 @@
...
+			color: getThemifyVariable("modal-border-color");
+		}
+	}
+	/* Handle */
+	::-webkit-scrollbar-thumb {
+		background: #888;
+	}
+	/* Handle on hover */
+	::-webkit-scrollbar-thumb:hover {
+		background: #555;
```
lindapaiste, line 30: What's the logic between applying scrollbar styling above 550px only?  Is it because we assume that small devices have good native scrollbar UI?

client/styles/layout/_dashboard.scss
```diff
@@ -6,8 +6,4 @@
   max-width: #{1012 / $base-font-size}rem;
   margin: 0 auto;
   width: 100%;
-
-  @include themify() {
-    border: 1px solid getThemifyVariable('modal-border-color');
-  }
```
lindapaiste, line 12: Deleting this removes the border around the table when on desktop.  I don't think you meant to do that.  Probably instead of deleting these lines you need to wrap them in a media query so that it only applies above 770px.

client/styles/components/_dashboard-header.scss
```diff
@@ -4,7 +4,62 @@
...
+            opacity: 1;
+          }
+        }
+      }
+
+      .dashboard-header__actions {
+        padding: 0;
+
+        > :first-child {
+          display: none;
```
lindapaiste, line 45: I don't feel comfortable publishing this with the "Create Collection" button hidden because that is removing functionality.  Here's a suggestion for a "good enough for now" solution:

Put a media query on the `.dashboard-header__actions` div to add the style `flex-direction: column-reverse;` which will render the button on a separate line below the search bar.  You can use `align-items: flex-end;` or `align-items: flex-start` to control the left-right alignment.

This will keep it fully functional until you get a "+" button working.

client/styles/components/_sketch-list.scss
```diff
@@ -2,11 +2,75 @@
...
+            padding-left: 0;
+            width: 100%;
+            font-weight: bold;
+            margin-bottom: #{8 / $base-font-size}rem;
+          }
+
+          > td {
+            padding-left: 0;
+            width: 30%;
+          }
```
lindapaiste, line 56: at some point we will want to make the dates be a smaller font size and a lighter color.

lindapaiste (changes requested):

client/modules/IDE/components/CollectionList/CollectionListRow.jsx
dewanshDT: yeah, I think in mobile we should only show something like "updated just now", "updated 1h ago", or "updated 13 Aug and not the created date.

client/modules/User/pages/DashboardView.jsx
dewanshDT: yeah 😀

client/styles/base/_base.scss
dewanshDT: yeah exactly and having that 10px wide scrollbar that looks beautiful in bigger devices is kinda ugly in mobile devices.

client/styles/layout/_dashboard.scss
dewanshDT: yeah I'll add a mediaquery instead

client/styles/components/_dashboard-header.scss
dewanshDT: added the "**+**" button, and functionality instead

client/styles/components/_sketch-list.scss
dewanshDT: yeah will do that

dewanshDT:
hey @lindapaiste, just pushed the required changes check'em out

dewanshDT:
hey @lindapaiste, I used Regular expressions for making the check of my stuff page url more robust and also created a hook `useWhatPage`, so that we can know from anywhere what page we are in currently.

dewanshDT:
@lindapaiste, I've also add the filter button

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -238,6 +238,7 @@ const MobileNav = () => {
         )}
       </Title>
       <Options>
+        {title === 'My Stuff' && <StuffMenu />}
```
lindapaiste: It is safer to look at `pageName` rather than `title` because eventually we will have a translation on `"My Stuff"`.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -261,6 +262,39 @@ const MobileNav = () => {
...
+            {t('DashboardView.CreateCollection')}
+          </NavMenuItem>
+        </ParentMenuContext.Provider>
+      </ul>
+      {createCollectionVisible && (
+        <Overlay
+          title={t('DashboardView.CreateCollectionOverlay')}
+          closeOverlay={() => setCreateCollectionVisible(false)}
+        >
+          <CollectionCreate />
```
lindapaiste, line 305: Very nice re-use of the existing `<CollectionCreate />` component 👍 

I'm just thinking to myself here.  Somewhere down the line this could be refactored in such a way that the existing overlay handling in `DashboardView` applies to both mobile and desktop.  That's way in the future, a few steps past `// TODO: remove the switch and use a props like mobileTitle <Nav layout=“dashboard” mobileTitle={t(‘Login’)} />`.  Where we have multiple separate Nav components like `StuffNav`, `EditorNav`, etc. and can pass down an `onClickCollectionCreate` prop to `StuffNav` only.

client/modules/IDE/hooks/useWhatPage.js
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable no-useless-escape */
+import { useEffect, useState } from 'react';
+import { useSelector } from 'react-redux';
+import { useLocation } from 'react-router-dom';
+
+/**
+ *
+ * @returns {"home" | "myStuff" | "login" | "signup" | "account" | "examples"}
```
lindapaiste, line 7: ❤️

client/modules/IDE/hooks/useWhatPage.js
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable no-useless-escape */
```
lindapaiste: It's considering the `\/` as a "useless" escape because you are using `new RegExp(string)` rather than creating the regex like `const myRe = /text/;`.  The `/` doesn't have any special meaning when inside a `string` so you don't need to escape it.

client/modules/IDE/hooks/useWhatPage.js
```diff
@@ -0,0 +1,33 @@
...
+/**
+ *
+ * @returns {"home" | "myStuff" | "login" | "signup" | "account" | "examples"}
+ */
+const useWhatPage = () => {
+  const projectName = useSelector((state) => state.project.name);
+  const username = useSelector((state) => state.user.username);
+  const { pathname } = useLocation();
+
+  const [pageName, setPageName] = useState(projectName);
```
lindapaiste: This doesn't really need to be a state.  You could use a `useMemo` instead of `useState` + `useEffect`, like we did in another place.

client/modules/IDE/hooks/useWhatPage.js
```diff
@@ -0,0 +1,33 @@
...
+  );
+
+  useEffect(() => {
+    if (myStuffPattern.test(pathname)) setPageName('myStuff');
+    else if (pathname === '/login') setPageName('login');
+    else if (pathname === '/signup') setPageName('signup');
+    else if (pathname === '/account') setPageName('account');
+    else if (pathname === '/p5/collections' || pathname === '/p5/sketches')
+      setPageName('examples');
+  }, [pathname]);
```
lindapaiste: At some point I want to enable the eslint `react-hooks/exhaustive-deps` rule but I'm thinking about the dependencies now.  The result depends on `myStuffPatttern` which depends on `username`.  You can move the `const myStuffPattern =` inside the body of the `useEffect` (or `useMemo`) and add `username` to the dependencies.

client/modules/User/components/DashboardTabSwitcher.jsx
```diff
@@ -31,8 +38,23 @@ Tab.propTypes = {
   to: PropTypes.string.isRequired
 };
 
+// It is good for right now, because we need to separate the nav dropdown logic from the navBar before we can use it here
```
lindapaiste, line 41: Can we make a todo list for future refactorings?  Maybe add a section to that Google doc?

client/modules/User/components/DashboardTabSwitcher.jsx
```diff
@@ -57,6 +79,51 @@ const DashboardTabSwitcher = ({ currentTab, isOwner, username }) => {
...
+        {(mobile) => (
+          <FilterOptions className={mobile ? 'visible' : 'hidden'}>
+            <div>
+              <IconButton icon={FilterIcon} />
+              <ul>
+                <li>
+                  <button
+                    onClick={() => dispatch(toggleDirectionForField('name'))}
+                  >
+                    Name
```
lindapaiste: We have translation keys that we can use here.  There's a different set of keys for `SketchList` and `CollectionList` but the values are the same so whatever.  Can use `t('CollectionList.HeaderCreatedAt')` etc.

client/modules/User/components/DashboardTabSwitcher.jsx
```diff
@@ -57,6 +79,51 @@ const DashboardTabSwitcher = ({ currentTab, isOwner, username }) => {
...
+                    Updated
+                  </button>
+                </li>
+                <li>
+                  <button
+                    onClick={() =>
+                      dispatch(toggleDirectionForField('numItems'))
+                    }
+                  >
+                    # of sketches
```
lindapaiste: We need to show this one only when viewing the collections list, as it doesn't make sense or do anything on the sketches list.

client/styles/base/_base.scss
lindapaiste: I wonder if there's a better media query to use instead of `width`.  Like checking if it's a touch device.  Something to look into.

lindapaiste (changes requested):
It's coming along great!

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: yeah

client/modules/IDE/components/Header/MobileNav.jsx
dewanshDT: yeah that's a nice idea!

client/modules/IDE/hooks/useWhatPage.js
dewanshDT: got it!

client/modules/IDE/hooks/useWhatPage.js
dewanshDT: yeah

client/modules/User/components/DashboardTabSwitcher.jsx
dewanshDT: yeah we should, and we can do it in github instead of google doc, we should create a todo list in github

dewanshDT:
yeah there's a media query to check for touch devices.
```css
/* Styles for touch devices */
@media (hover: none) and (pointer: coarse) { }

/* Styles for non-touch devices */
@media (hover: hover) and (pointer: fine) {}
```

dewanshDT:
hey @lindapaiste, update this PR, check it out

client/modules/IDE/hooks/useWhatPage.js
```diff
@@ -0,0 +1,31 @@
...
+      `(/${username}/(sketches/?$|collections|assets)/?)`
+    );
+
+    if (myStuffPattern.test(pathname)) return 'myStuff';
+    else if (pathname === '/login') return 'login';
+    else if (pathname === '/signup') return 'signup';
+    else if (pathname === '/account') return 'account';
+    else if (pathname === '/p5/collections' || pathname === '/p5/sketches')
+      return 'examples';
+    return projectName;
```
lindapaiste: Might make more sense to return something like `'project'` or `'editor'`, since the return type is a union of page types.

lindapaiste:
Let's hide the sorting button on the "Assets" tab for now, since it doesn't work properly there.  

We may revisit putting in searching and sorting on the assets later on.  Some of the refactoring that I've been doing on the table components should help.

client/modules/User/components/DashboardTabSwitcher.jsx
```diff
@@ -57,6 +79,53 @@ const DashboardTabSwitcher = ({ currentTab, isOwner, username }) => {
...
+                  </button>
+                </li>
+                {currentTab === TabKey.collections && (
+                  <li>
+                    <button
+                      onClick={() =>
+                        dispatch(toggleDirectionForField('numItems'))
+                      }
+                    >
+                      # of sketches
```
lindapaiste:
```suggestion
                      {t('CollectionList.HeaderNumItems')}
```

client/modules/IDE/hooks/useWhatPage.js
dewanshDT: ohh shit, I totally forgot, its `'home'`

client/styles/base/_base.scss
dewanshDT: check it out I've pushed it

client/modules/IDE/components/CollectionList/CollectionListRow.jsx
shujuuu: Make sense @dewanshDT, let's go with that proposal!

lindapaiste (dismissed):
I think this is at a point where it's good enough the merge.  Just need to fix the failing tests!  It's probably best if we put in a separate PR to install `jest-styled-components` and regenerate all snapshots.  You can also just regenerate the snapshots here if you want to save that for later.

dewanshDT:
> I think this is at a point where it's good enough the merge. Just need to fix the failing tests! It's probably best if we put in a separate PR to install `jest-styled-components` and regenerate all snapshots. You can also just regenerate the snapshots here if you want to save that for later.

so you're going to make a pr for the `jest-styled-components`, or you want me to do that?

dewanshDT:
@lindapaiste let's merge this 🎉

lindapaiste (approved):
Looks great! 🥇

PR #2396, opened 2023-09-02
Jest styled components

dewanshDT:
Changes:
added the `jest-styled-components`


lindapaiste (approved):
Yesssss thank you!

@raclim We don't love this solution because it requires us to update the snapshots whenever we change the styles.  Ideally we would ignore the styles and just look at the HTML.  It's definitely an improvement because it gives us stable class names and ensures that these snapshots won't fail due to changes elsewhere.

PR #2398, opened 2023-09-08
docs: add default stories for IDE components

ofhope:
Changes:

Adds default stories to IDE components. At least to the ones that make sense. There are a few like Toast, Timer and EditorAccessibility which is either hidden or needs some redux magic set up to trigger output.

There is more that could be done for each story to offer any other variations. But for a first pass I think it's good to get a simple story. I think would ease development for things like file upload or copyable field as they are now surfaced as an isolated story.

Screenshot 2023-09-08 at 1 48 27 pm
Screenshot 2023-09-08 at 1 48 18 pm
Screenshot 2023-09-08 at 1 47 56 pm
Screenshot 2023-09-08 at 1 47 29 pm
Screenshot 2023-09-08 at 1 47 24 pm


client/modules/IDE/components/AssetPreview.stories.jsx
```diff
@@ -0,0 +1,13 @@
...
+  title: 'IDE/AssetPreview',
+  component: AssetPreview
+};
+
+export const Default = {
+  args: {
+    url: 'https://p5js.org/assets/img/p5js.svg',
+    name: 'P5 Logo'
+  }
+};
```
lindapaiste, line 13: We can add stories for audio and video files (can be done later).

client/modules/IDE/components/AssetSize.stories.jsx
```diff
@@ -0,0 +1,12 @@
+import AssetSize from './AssetSize';
+
+export default {
+  title: 'IDE/AssetSize',
+  component: AssetSize
+};
+
+export const Default = {
+  args: {
+    totalSize: 123
```
lindapaiste, line 10: This story shows up with a control for the `totalSize`, but changing it doesn't do anything because the default export gets the size from Redux.

Once we get the styles working and once we get this using the `args` then it might be cool to have stories for empty vs. full vs. somewhere in the middle.

client/modules/IDE/components/EditableInput.stories.jsx
```diff
@@ -0,0 +1,13 @@
...
+  title: 'IDE/EditableInput',
+  component: EditableInput
+};
+
+export const Default = {
+  args: {
+    value: 'an@email.com',
+    label: 'Example'
+  }
+};
```
lindapaiste, line 13: We aren't seeing the pencil icon here due to an issue with `svgr`.  I think it's removing the `viewbox`.  Just FYI, we can fix that later.

client/modules/IDE/components/EditorAccessibility.stories.jsx
```diff
@@ -0,0 +1,19 @@
...
+    lintMessages: [
+      {
+        severity: 'warn',
+        line: '10',
+        message: 'Warning Foo',
+        id: '123'
+      }
+    ]
+  }
+};
```
lindapaiste: I'm not sure what makes sense for a story here, as this component is hidden/invisible.  We would want to review the text that we are outputting (I recall it being not quite right at the moment).  We might want to create an export which has the text contents of the component with no styling.

client/modules/IDE/components/NewFileForm.stories.jsx
```diff
@@ -0,0 +1,8 @@
+import NewFileForm from './NewFileForm';
+
+export default {
+  title: 'IDE/NewFileForm',
+  component: NewFileForm
+};
+
+export const Default = {};
```
lindapaiste, line 8: The `NewFileForm` is only used inside of the `NewFileModal` so it might be redundant to have stories for both. (the same for the `NerFolderForm`)

TBH we could consider combing the two component files as there's not much left in the `NewFileModal` after we made a common `Modal` component.

client/modules/IDE/components/SketchList.stories.jsx
```diff
@@ -0,0 +1,25 @@
...
+    username: 'Joe Blogs',
+    loading: false,
+    sketches: [
+      {
+        id: '1',
+        name: 'Play Sketch',
+        createdAt: Date.now().toString(),
+        updatedAt: Date.now().toString()
+      }
+    ]
```
lindapaiste, line 23: I was wondering why these sketches don't show up in the list.  Again, it's because the component ignores this prop and gets the actual value from Redux.

lindapaiste (dismissed):
Excellent work 👍 You've got me thinking about lots of things and I've added many comments with things that we can improve in the future.

The biggest thing is how these components interact with Redux. We have two approaches here to make these stories make sense:

1. Each component exports a connected version which gets the state from Redux and an unconnected version which accepts props.  We would use the unconnected version for Storybook.  This is easy for the old components which use `connect`.  But for the ones that I've rewritten, like `Timer`, I've generally been making it so that the component takes no props and accesses Redux internally.

2. We pass the `args` of the story into a Redux provider so that we can specify the correct state.  This requires setup and more work on the Storybook side, but no changes to the components which is a pro. But it also requires the story to know the structure of the Redux state, ie. that `projectSavedTime` should go in `state.project.updatedAt`, and that is a con to me because Redux should be a black box.

client/modules/IDE/components/EditorAccessibility.stories.jsx
ofhope: I'm not terribly familiar with the components so I wasn't sure if it made sense having this here.

Probably makes sense to remove it form as a story and rely on unit tests to qualify behaviour if its not a visual thing.

client/modules/IDE/components/NewFileForm.stories.jsx
ofhope: I like the idea of having fewer Modals and reusing the generic one. I think thats capturing the layout primitives

client/modules/IDE/components/SketchList.stories.jsx
ofhope: Ahhh! I was wondering that myself but was unsure.

client/modules/IDE/components/NewFileForm.stories.jsx
ofhope: I've removed `NerFolderForm`

ofhope:
@lindapaiste is there anything here you'd like me to action? I think the comments are follow up items.

lindapaiste (approved):
Sorry that this was sitting for so long!  I'm going to approve and merge this because it is a huge step forwards.  We can continue to workout some of the kinks with future PRs.

PR #2400, opened 2023-09-09
Account Settings Page UI for mobile devices

dewanshDT:
Changes:

Account Settings Page UI Update in small devices
image

### DIFF
--- a/client/styles/components/_forms.scss
+++ b/client/styles/components/_forms.scss
@@ -57,12 +57,17 @@
 
 .form__input {
   max-width: 90vw;
-  width: #{360 / $base-font-size}rem;
+  // width: #{360 / $base-font-size}rem;
+  width: 100%;
   height: #{40 / $base-font-size}rem;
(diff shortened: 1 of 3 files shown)


client/styles/components/_forms.scss
```diff
@@ -57,13 +57,18 @@
 
 .form__input {
   max-width: 90vw;
-  width: #{360 / $base-font-size}rem;
+  // width: #{360 / $base-font-size}rem;
+  width: 100%;
```
lindapaiste, line 60: This change has unintended consequences as it's used on all form inputs.  Check out the login and signup pages on desktop.  We might need an override that it specific to the account settings page.

lindapaiste (changes requested):

client/styles/components/_forms.scss
dewanshDT: hey @lindapaiste, sorry for late response, I just fixed the problem, please check it out

lindapaiste (approved):

Issue #2358, opened 2023-08-07
Convert legacy React class components to function components


lindapaiste:
### Increasing Access

It keeps the codebase up to date and makes it easier for new contributors.

### Feature enhancement details

[`React.Component` is considered a legacy API and is not recommended.](https://react.dev/reference/react/Component)  We should convert all class components to function components which use React hooks.

---

This is mainly a "Progress Report" for myself because with so many unmerged PRs sometimes I forget what I've already worked on!

Here are the class components which remain the codebase, and the PRs to convert them:
- [ ] EmailVerificationView
- [x] App #2308 
- [x] Overlay #2309 
- [x] AddToCollectionList #2353
- [x] AddToCollectionSketchList #2364
- [ ] ConsoleInput
- [x] CopyableInput #2360
- [ ] Editor -- this one is too annoying/difficult due to the potential for stale closures.  It will likely wait until upgrading to codemirror v6, when the component needs a lot of changes anyways.
- [ ] FileNode
- [x] Toolbar #2352
- [x] Preferences #2335 
- [ ] Searchbar
- [x] IDEView done by @dewanshDT, not submitted yet
- [x] APIKeyForm #2307 
- [x] Collection #2366
- [x] CollectionView #2302 
- [ ] DashboardView
- [ ] Tables -- should be addressed together (see #1442)
  - [ ] SketchListRowBase
  - [x] SketchList #2376
  - [ ] AssetListRowBase
  - [x] AssetList #2376
  - [ ] CollectionListRowBase
  - [x] CollectionList #2376

vijayabhaskar-ev:
Convert EmailVerificationView to function component #2401

KSSaiTeja:
i want to work to work to convert the collectionListRowbase component and make a PR. can i do it?

shauryag2002:
Hi @lindapaiste , It seems like you are working on this issue quite long time,
if not able to do it !
Can @lindapaiste assign it to me?

lindapaiste:
Thank you everyone for all of your work on this!  It looks like all components have been covered in pending PRs.

We do not need any more new PRs for this issue.

KSSaiTeja:
@lindapaiste could you please take my PR into consideration and review it

up1512001:
Can I work on DashboardView if anyone didn't?

thakkarnetram:
Ref #2447 .
I have updated the Assets List to a functional component

lindapaiste:
[...]

PR #2403, opened 2023-09-10
Fix: Searchbar class component to functional component

m-Chetan:
Changed Searchbar class component to functional component.

Ref #2358


raclim:
Thanks for opening this! 

A few changes I might make here are using the arrow function definition on line 8, and potentially finding a way to compress the `useEffect()` hook on line 17. Since there's a few other pull requests that are refactoring the `Searchbar` component, so I'm going to close this for now, but please feel free to submit another PR!

PR #2409, opened 2023-09-11
Converted ConsoleInput to functional components

rain2003:
Ref #2358 

Changes: I have changed the ConsoleInput.jsx file from legacy to functional component.

### DIFF
--- a/client/modules/IDE/components/ConsoleInput.jsx
+++ b/client/modules/IDE/components/ConsoleInput.jsx
@@ -1,39 +1,33 @@
+import React, { useRef, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
-import React from 'react';
 import CodeMirror from 'codemirror';
 import { Encode } from 'console-feed';
 
 import RightArrowIcon from '../../../images/right-arrow.svg';
 import { dispatchMessage, MessageTypes } from '../../../utils/dispatcher';
 
-// heavily inspired by
-// https://github.com/codesandbox/codesandbox-client/blob/92a1131f4ded6f7d9c16945dc7c18aa97c8ada27/packages/app/src/app/components/Preview/DevTools/Console/Input/index.tsx
+const ConsoleInput = ({ theme, dispatchConsoleEvent, fontSize }) => {
+  const [commandHistory, setCommandHistory] = useState([]);
+  const [commandCursor, setCommandCursor] = useState(-1);
 
-class ConsoleInput extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      commandHistory: [],
-      commandCursor: -1
-    };
-  }
+  const codemirrorContainerRef = useRef(null);
+  const codemirrorRef = useRef(null);
 
-  componentDidMount() {
-    this._cm = CodeMirror(this.codemirrorContainer, {
-      // eslint-disable-line
-      theme: `p5-${this.props.theme}`,
+  useEffect(() => {
+    codemirrorRef.current = CodeMirror(codemirrorContainerRef.current, {
+      theme: `p5-${theme}`,
       scrollbarStyle: null,
       keymap: 'sublime',
       mode: 'javascript',
       inputStyle: 'contenteditable'
     });
 
-    this._cm.on('keydown', (cm, e) => {
+    const handleKeyDown = (cm, e) => {
       if (e.key === 'Enter' && !e.shiftKey) {
         e.preventDefault();
         e.stopPropagation();
         const value = cm.getValue();
-        if (value.trim(' ') === '') {
+        if (value.trim() === '') {
           return false;
         }
         const messages = [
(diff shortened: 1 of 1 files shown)


client/modules/IDE/components/ConsoleInput.jsx
```diff
@@ -1,39 +1,33 @@
...
 import PropTypes from 'prop-types';
-import React from 'react';
 import CodeMirror from 'codemirror';
 import { Encode } from 'console-feed';
 
 import RightArrowIcon from '../../../images/right-arrow.svg';
 import { dispatchMessage, MessageTypes } from '../../../utils/dispatcher';
 
-// heavily inspired by
-// https://github.com/codesandbox/codesandbox-client/blob/92a1131f4ded6f7d9c16945dc7c18aa97c8ada27/packages/app/src/app/components/Preview/DevTools/Console/Input/index.tsx
```
lindapaiste, line 10: I would probably leave this in? Even though the code is not as similar anymore.

client/modules/IDE/components/ConsoleInput.jsx
```diff
@@ -47,92 +41,84 @@ class ConsoleInput extends React.Component {
...
-        this.setState((state) => {
-          const newCursor = Math.min(
-            state.commandCursor + 1,
-            state.commandHistory.length - 1
-          );
-          this._cm.getDoc().setValue(state.commandHistory[newCursor] || '');
-          const cursorPos = this._cm.getDoc().getLine(0).length - 1;
-          this._cm.getDoc().setCursor({ line: 0, ch: cursorPos });
-          return { commandCursor: newCursor };
+        setCommandCursor((prevCursor) => {
```
lindapaiste, line 54: Love that you are using a prevState callback!

client/modules/IDE/components/ConsoleInput.jsx
```diff
@@ -47,92 +41,84 @@ class ConsoleInput extends React.Component {
...
+          const newCursor = Math.min(prevCursor + 1, commandHistory.length - 1);
+          cm.getDoc().setValue(commandHistory[newCursor] || '');
```
lindapaiste, line 56: It seems not quite right to be performing these side effects inside of a set state.  But that's how it was before so let's leave it for now.

client/modules/IDE/components/ConsoleInput.jsx
```diff
@@ -47,92 +41,84 @@ class ConsoleInput extends React.Component {
...
-    this._cm.getWrapperElement().style[
-      'font-size'
-    ] = `${this.props.fontSize}px`;
-    this._cm.refresh();
-  }
+    return () => {
+      codemirrorRef.current.off('keydown', handleKeyDown);
+      codemirrorRef.current = null;
+    };
+  }, [theme, dispatchConsoleEvent, fontSize]);
```
lindapaiste, line 90: We have to be careful here because we don't want to re-create the entire codemirror instance when `theme` or `fontSize` change.

I think (need to try it out to be sure) that you can remove the `theme: `p5-${theme}`,` and the setting of the fontSize from this main `useEffect` as they are already covered by your next `useEffect`.

client/modules/IDE/components/ConsoleInput.jsx
```diff
@@ -47,92 +41,84 @@ class ConsoleInput extends React.Component {
...
 
-  componentWillUnmount() {
-    this._cm = null;
-  }
+  useEffect(() => {
+    if (codemirrorRef.current) {
+      codemirrorRef.current.setOption('theme', `p5-${theme}`);
+      codemirrorRef.current.getWrapperElement().style[
+        'font-size'
+      ] = `${fontSize}px`;
```
lindapaiste, line 97: This part of the code has always felt a bit wonky to me.  I think that you can delete these lines from all `useEffect`s and instead put `style={{ fontSize }}` on the `<div ref={codemirrorContainerRef}>` element

client/modules/IDE/components/ConsoleInput.jsx
```diff
@@ -1,39 +1,33 @@
...
 import { Encode } from 'console-feed';
 
 import RightArrowIcon from '../../../images/right-arrow.svg';
 import { dispatchMessage, MessageTypes } from '../../../utils/dispatcher';
 
-// heavily inspired by
-// https://github.com/codesandbox/codesandbox-client/blob/92a1131f4ded6f7d9c16945dc7c18aa97c8ada27/packages/app/src/app/components/Preview/DevTools/Console/Input/index.tsx
+const ConsoleInput = ({ theme, dispatchConsoleEvent, fontSize }) => {
+  const [commandHistory, setCommandHistory] = useState([]);
+  const [commandCursor, setCommandCursor] = useState(-1);
```
lindapaiste, line 11: Let's put a `// eslint-disable-next-line no-unused-vars` above this line to suppress the warning.  It's definitely not the best code to have a state which isn't used but that's not your fault, that's how it is currently.

lindapaiste (changes requested):
This code doesn't actually work, does it?  If I type `let x = 1 + 2` into the editor console and hit enter, it just creates extra codemirror instances! 😭 

So definitely more work is needed here.  I suspect it's the way that we are passing down the `dispatchConsoleEvent` prop, in which case it can be fixed by connecting this component to Redux directly.  But I'm still working on figuring it out.

lindapaiste:
> I suspect it's the way that we are passing down the `dispatchConsoleEvent` prop, in which case it can be fixed by connecting this component to Redux directly. But I'm still working on figuring it out.

This is a very complicated component!

I confirmed that I can type and submit to the console if the only dependency of the `useEffect` is `[dispatch]`.  However the arrow up and arrow down to navigate through previous statements still doesn't work and now I'm seeing why.  Those functions depend on `commandHistory`, but we do not have `commandHistory` as a dependency of the `useEffect` and so we will always have a stale/outdated value of `commandHistory` inside the effect.  We can't just add `commandHistory` as a dependency because then we would go back to having the problem where it keeps creating new CodeMirror instances.

You might be able to get it working if you use one `useState` where the state is an object with properties `commandCursor` and `commandHistory` like it was before.  Then you can access the `commandHistory` inside your `setState` from the `prevState`.

rain2003:
I tried to use the above method but still am not getting anywhere can you please tell what's wrong in this I have even tried visiting other websites which use a similar type of console box and looked at their code but still can't find out what's wrong with mine.
[consoleinputfunctionaltxt.txt](https://github.com/processing/p5.js-web-editor/files/12647406/consoleinputfunctionaltxt.txt)

PR #2421, opened 2023-09-12
covert SketchListRowBase Class to Functional compontent

mubacodestudio:
Fixes Ref #2358 

Changes: change SketchListRowBase to Functional component

### DIFF
--- a/client/modules/IDE/components/SketchList.jsx
+++ b/client/modules/IDE/components/SketchList.jsx
@@ -1,5 +1,5 @@
+import React, { useState, useRef } from 'react';
 import PropTypes from 'prop-types';
-import React from 'react';
 import { Helmet } from 'react-helmet';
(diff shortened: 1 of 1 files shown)


lindapaiste:
I'm going to close this PR because it has conflicts with other recent changes to the `develop` branch but also has major problems of its own.  

When running this branch I am unable to view the `/sketches` page and I get an error:
> Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

You've removed the part where the `SketchListRowBase` gets connected to Redux which means that a lot of its props are now missing, as they are neither passed down from the parent nor provided by Redux.  Those missing function props will cause errors when those functions are called.

In the future please run your code and make sure that it works before submitting PRs.

PR #2438, opened 2023-09-14
confusing icons hover issue fix #2255

### LINKED ISSUE #2255: Confusing icons in `QuickAddList`
### Increasing Access

It will be easier to make selections if it is obvious which items have already been selected.

This screenshot shows the UI when I click "Add Sketch" on a collection details page.  The first sketch in the list ("Stormy Sandalwood") is already in  the collection and the other two are not.  But there is no way to know that unless you hover over the ✔️ icon, where you will see either a ❌ to remove the item if it is already added or a ➕ to add the item if not.  This is not great in general but it's totally unusable on touch devices where there is no hover.  A user will not know what clicking the icon is going to do: will it add or remove?

The ✔️ does not mean anything as all items show the ✔️.  In my opinion, we should only show the ✔️ for items which are already in the collection.

As for the items which are not in the collection, we could show the ➕ all the time and not just on hover.  I've also had a look at @shujuuu's [mobile design for this component](https://www.figma.com/file/27wGlxCRns7tyXoEIbcmTA/p5-Mobile---2020?type=design&node-id=199-126&t=VAxy4ru1Z5s71j7n-0).  She is using more of a "radio button" type look, where there is a very clear and obvious ☑️ for items which are added and an empty circle ⚪ for items which can be added by clicking.  If @dewanshDT gets around to creating this radio button element for his GSoC project then we should use that element on the web version as well.  It would definitely be an improvement to usability.

image

We show this `QuickAddList` modal:
- With a list of collections, when clicking "add to collection" from the editor.
- With a list of sketches, when clicking "add sketch" on the collection.

luckyklyist:
Fix Confusing Icons #2255 

### Changes Made
I've addressed the issue of confusing icons that occurred due to the Sass hover styles. To enhance the user experience, I reorganized the code related to hover-specific styles. Now, users can clearly see whether they've added a sketch without needing to hover over the button.

https://github.com/processing/p5.js-web-editor/assets/35479077/1a8533bc-3081-4db7-996b-2d449e14794e


luckyklyist:
@lindapaiste can you review this pr ?

lindapaiste:
I think this is an improvement so I should probably merge it as-is.  Just need to load up the branch and run it myself.

Since we are no longer showing different icons on hover the CSS could be cleaned up further, if you want to. We can use the JS to return only the one matching icon and remove all of the `display: none;` and `display: inline-block;` switching from the CSS, rather than having both in the HTML and using CSS to hide one.

There are things which are still weird about this `QuickAddList` which can be handled in future PRs.  We have an "are you sure?" dialog before removing a sketch via the normal list, but we don't have any confirmation here.  And also there's an `aria-label="Descending"` which make no sense 🙃.  There's a correct label on the parent `button` so it should be `aria-hidden="true"` instead.  But those things aren't related to the confusing icons.

lindapaiste (dismissed):
This is definitely an improvement!  It still feels a bit confusing though, especially because the "+" and "x" icons are so visually similar.  Do you think it would be better if we used the ✔️ instead of the "x" for items in the collection?

luckyklyist:
I have changed the + to ✔️ . 
Screenshot 2023-09-23 at 11 10 42 PM

raclim (approved):
Thanks for your work on this! :)

PR #2450, opened 2023-09-16
fixes the logout misbehave #2449

### LINKED ISSUE #2449: Login Session Still Persists Despite Logout
### p5.js version

1.7.0

Windows

### Web browser and version

chrome, edge, Brave

### Actual Behavior

When we try to log out, the session Id is not deleted

Screenshot (92)
Screenshot (95)
Screenshot (93)
Screenshot (94)

As a result, after clicking logout, when I refresh the page, I again enter the login state on the website.

### Expected Behavior

The sessionId got deleted and after refresh, the user can not go back to the login state.

### Steps:
1. click logout.
2. hit refresh
3. you get back to login state

Ankush263:
Fixes #2449

Changes:

* [ ✅] has no linting errors (`npm run lint`)
* [✅ ] has no test errors (`npm run test`)
* [✅ ] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* [✅ ] is descriptively named and links to an issue number, i.e. `Fixes #123`

### DIFF
--- a/server/controllers/session.controller.js
+++ b/server/controllers/session.controller.js
@@ -31,11 +31,9 @@ export function getSession(req, res) {
 }
 
 export function destroySession(req, res, next) {
-  req.logout((err) => {
-    if (err) {
-      next(err);
-      return;
-    }
-    res.json({ success: true });
+  res.cookie('sessionId', 'loggedout', {
+    expires: new Date(Date.now() + 10 * 1000),
+    httpOnly: true
   });
+  res.status(200).json({ status: 'success' });
 }
--- a/server/controllers/session.controller.js
+++ b/server/controllers/session.controller.js
@@ -31,9 +31,6 @@ export function getSession(req, res) {
 }
 
 export function destroySession(req, res, next) {
-  res.cookie('sessionId', 'loggedout', {
-    expires: new Date(Date.now() + 10 * 1000),
-    httpOnly: true
-  });
-  res.status(200).json({ status: 'success' });
+  res.clearCookie('sessionId');
+  res.json({ success: true });
 }


server/controllers/session.controller.js
```diff
@@ -31,11 +31,9 @@ export function getSession(req, res) {
...
 export function destroySession(req, res, next) {
-  req.logout((err) => {
-    if (err) {
-      next(err);
-      return;
-    }
-    res.json({ success: true });
+  res.cookie('sessionId', 'loggedout', {
+    expires: new Date(Date.now() + 10 * 1000),
+    httpOnly: true
```
lindapaiste: I'm trying to understand this.  It seems like you are setting a new cooking which will expire in 10 seconds, and that cookie has an invalid value.  Wouldn't it make more sense to use [`res.clearCookie`](http://expressjs.com/en/5x/api.html#res.clearCookie)?

Why delete the existing [`req.logout`](https://www.passportjs.org/concepts/authentication/logout/)?  If we do need to clear the cookie, it should be in addition to calling `req.logout`.

Ankush263: OK, I'm gonna try it with a different method

server/controllers/session.controller.js
```diff
@@ -31,11 +31,6 @@ export function getSession(req, res) {
...
 
 export function destroySession(req, res, next) {
-  req.logout((err) => {
-    if (err) {
-      next(err);
-      return;
-    }
-    res.json({ success: true });
-  });
+  res.clearCookie('sessionId');
```
Ankush263, line 34: Hey @lindapaiste, I use `clearCookie` here...
I think this is one of the optimized solutions

raclim:
Thanks for working on this and for raising this issue! I think we might go with the solution in #2473 for now, but please feel free to reopen this again if the issue persists!

PR #2452, opened 2023-09-18
Fix: cursor on Login button becomes `cursor: not-allowed` in pristine state

### LINKED ISSUE #2228: Log In button sometimes shows `cursor: not-allowed` when autofilled with a saved password
### p5.js version

_No response_

Windows

### Web browser and version

Chrome 113.0.5672.93

### Actual Behavior

`cursor: not-allowed` 🚫 

This is extremely minor, but I've noticed an issue with the form state on the login page where it can sometimes show the wrong cursor when hovering the "Log In" button.  It still works when I click it, but it looks as though it is not clickable.  This corrects itself as soon as I click anywhere within the window.  It only happens if I start directly on the `https://editor.p5js.org/login` page.  Chrome autofills my username and password so the form is complete before the window is focused.

### Expected Behavior

`cursor: pointer` 👆 

I can't capture the cursor in my screenshots, but I also see that the text is not getting the correct CSS style in the auto-fill situation.  Again, it fixes itself as soon as I click anywhere inside the window.

Screenshot 2023-06-04 12 57 25
Screenshot 2023-06-04 13 01 55

### Steps:
1. Save your login details using the Chrome password manager
2. Navigate directly to URL [https://editor.p5js.org/login](https://editor.p5js.org/login)
3. Hover over the "Log In" button

rajatmohan22:
Fixes #2228 

Upon some investigation, I found that the `value` prop of  the input field in a web form is not able to recognize Google Autocomplete with a password manager. Seems like it is bypassing Reacts state management.

[screen-capture.webm](https://github.com/processing/p5.js-web-editor/assets/77490034/0b17027c-bf25-4873-9256-40bcb9d38c50)

Changes:

### DIFF
--- a/client/modules/User/components/LoginForm.jsx
+++ b/client/modules/User/components/LoginForm.jsx
@@ -20,12 +20,7 @@ function LoginForm() {
       validate={validateLogin}
       onSubmit={onSubmit}
     >
-      {({
-        handleSubmit,
-        submitError,
-        submitting,
-        modifiedSinceLastSubmit
-      }) => (
+      {({ handleSubmit, submitError, submitting, modifiedSinceLastSubmit }) => (
         <form className="form" onSubmit={handleSubmit}>
           <Field name="email">
             {(field) => (
--- a/client/modules/User/components/LoginForm.jsx
+++ b/client/modules/User/components/LoginForm.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Form, Field } from 'react-final-form';
 import { useDispatch } from 'react-redux';
@@ -7,7 +7,22 @@ import { validateLogin } from '../../../utils/reduxFormUtils';
 import { validateAndLoginUser } from '../actions';
 
 function LoginForm() {
+  const [contentPresent, setContentPresent] = useState(false);
   const { t } = useTranslation();
+
+  useEffect(() => {
+    const checkEmailElement = () => {
+      if (
+        document.getElementById('email') &&
+        document.getElementById('password')
+      ) {
+        setContentPresent(true);
+      } else {
+        setContentPresent(false);
+      }
+    };
+    checkEmailElement();
+  }, []);
 
   const dispatch = useDispatch();
   function onSubmit(formProps) {
@@ -71,7 +86,10 @@ function LoginForm() {
           {submitError && !modifiedSinceLastSubmit && (
             <span className="form-error">{submitError}</span>
           )}
-          <Button type="submit" disabled={submitting || pristine}>
+          <Button
+            type="submit"
+            disabled={(submitting || pristine) && !contentPresent}
+          >
             {t('LoginForm.Submit')}
           </Button>
         </form>
--- a/client/modules/User/components/LoginForm.jsx
+++ b/client/modules/User/components/LoginForm.jsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { Form, Field } from 'react-final-form';
 import { useDispatch } from 'react-redux';
--- a/client/modules/User/components/LoginForm.jsx
+++ b/client/modules/User/components/LoginForm.jsx
@@ -7,22 +7,7 @@ import { validateLogin } from '../../../utils/reduxFormUtils';
 import { validateAndLoginUser } from '../actions';
 
 function LoginForm() {
-  const [contentPresent, setContentPresent] = useState(false);
   const { t } = useTranslation();
-
-  useEffect(() => {
-    const checkEmailElement = () => {
-      if (
-        document.getElementById('email') &&
-        document.getElementById('password')
-      ) {
-        setContentPresent(true);
-      } else {
-        setContentPresent(false);
-      }
-    };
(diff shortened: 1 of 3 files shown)


client/modules/User/components/LoginForm.jsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
```
lindapaiste:
```suggestion
import React from 'react';
```

client/modules/User/components/LoginForm.jsx
```diff
@@ -71,7 +71,7 @@ function LoginForm() {
           {submitError && !modifiedSinceLastSubmit && (
             <span className="form-error">{submitError}</span>
           )}
-          <Button type="submit" disabled={submitting || pristine}>
+          <Button type="submit" disabled={submitting}>
```
lindapaiste, line 68: Awesome!

This seems like the best solution.  I wish there was some way that the `react-final-form` package would know that the field is not `pristine` when it's been autofilled.  But I don't think that's possible?  When I look at the HTML of the page in dev tools I see that the `value` prop of the inputs is not set when autofilled if the page has not been clicked yet.  Then as soon as I click the window the `input` changes from `value`  to `value="lindapaiste@gmail.com"`.  It's very strange because I don't understand how Chrome is able to display the value without injecting it into the HTML.

lindapaiste (changes requested):
Looks good but let's clean up the `import` line before merging.

client/modules/User/components/LoginForm.jsx
rajatmohan22: > Awesome!
> 
> This seems like the best solution. I wish there was some way that the `react-final-form` package would know that the field is not `pristine` when it's been autofilled. But I don't think that's possible? When I look at the HTML of the page in dev tools I see that the `value` prop of the inputs is not set when autofilled if the page has not been clicked yet. Then as soon as I click the window the `input` changes from `value` to `value="lindapaiste@gmail.com"`. It's very strange because I don't understand how Chrome is able to display the value without injecting it into the HTML.

I agree. It seems like a browser security measure aimed at preventing data harvesting threats. They may be trying to avoid situations where websites could exploit automatic form filling to collect sensitive user data without their consent.

If you visit Amazon's login page, you'll notice a similar login page behavior where the login button isn't disabled in the Pristine stages ( probably due to the same reasons ).

client/modules/User/components/LoginForm.jsx
rajatmohan22: Done !

rajatmohan22:
@lindapaiste could you have a look? I also feel that we need to make sure that the value prop when i inspect the page doesn't hold the actual value of the entered password (which is currently happening).

We could use a hash+salt instead. However, I feel we could have a different PR for that

rajatmohan22:
@lindapaiste I was wondering if we could disable the value prop for the password field. I feel it is kind of unsafe to have it record the value that the user types.

Could we do something like this?:

`<Field name="password">
            {(field) => {
              const { value, ...restOfFieldInput } = field.input;
              return (
                <p className="form__field">
                  <label htmlFor="password" className="form__label">
                    {t('LoginForm.Password')}
                  </label>
                  <input
                    className="form__input"
                    aria-label={t('LoginForm.PasswordARIA')}
                    type="password"
                    id="password"
                    autoComplete="current-password"
                    value={console.log(field.input.value)}
                    {...restOfFieldInput}
                  />
                  {field.meta.touched && field.meta.error && (
                    <span className="form-error">{field.meta.error}</span>
                  )}
                </p>
              );
            }}
          </Field>`

rajatmohan22:
@lindapaiste, could you please take a look? I've removed unnecessary imports. The only thing I believe needs attention is the 'value' prop storing the user's input. Do you think it's safe to leave it this way, or should I make changes here if necessary?

lindapaiste:
I don’t think that the value prop is a problem?  You can definitely research what the best practices are for dealing with an input type=“password”, but let’s tackle that in a separate PR.

If it’s easy, can you roll back the changes in the package-lock.json file and in the Button file?  It’s it’s a pain then no worries.

rajatmohan22:
> I don’t think that the value prop is a problem? You can definitely research what the best practices are for dealing with an input type=“password”, but let’s tackle that in a separate PR.
> 
> If it’s easy, can you roll back the changes in the package-lock.json file and in the Button file? It’s it’s a pain then no worries.

Oh no problem, It's done!

rajatmohan22:
@lindapaiste, not a problem at all. I've rolled them back. My apologies for missing that.

lindapaiste (dismissed):

lindapaiste:
We’re just about ready to merge but it’s failing lint checks due to a formatting issue.  We use a package called Prettier so that all code is formatted nicely and consistently.  Now that you removed one of the destructured properties, Prettier thinks that it should all be on one line instead of on separate lines.  If you run the `lint-fix` command that should take care of it.

rajatmohan22:
> We’re just about ready to merge but it’s failing lint checks due to a formatting issue. We use a package called Prettier so that all code is formatted nicely and consistently. Now that you removed one of the destructured properties, Prettier thinks that it should all be on one line instead of on separate lines. If you run the `lint-fix` command that should take care of it.

@lindapaiste The test passed! Thanks for the suggestion

lindapaiste (approved):

PR #2459, opened 2023-09-18
add CSS for vertical alignment of checkbox labels

### LINKED ISSUE #2458: add CSS for vertical alignment of checkbox labels
### Increasing Access

**p5.js version**
v1.4.1

**What is your operating system?**
Windows OS

**### Problem:** 
The checkbox and Text are not Linear.
**_Default Behavior:_**

image

**_After adding CSS:_**

image

### **Changes:** 

- Add margin for good styling.

BamaCharanChhandogi:
Fixes #issue-number

Changes:


BamaCharanChhandogi:
@lindapaiste  Could you please check?

lindapaiste (changes requested):
Thanks for raising this issue!  It's a good issue but we need to be careful about how we implement the fix.

### 1.  It should not change the appearance of other elements.  

The same `preference__option` class name is also used on the radio options.  We don't want to change the margins on all options.  We only want to change the checkbox options.  We could add additional class names to get greater specificity.  We can also use [sibling selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator) to select option labels appearing alongside checkboxes.

### 2. It should work the same for all browsers and devices.

When I look at the existing UI in dev tools in Chrome (on Windows) I can see that the `margin` around the checkbox is coming from the user agent stylesheet, which is the built-in Chrome styles.
image
image

A different browser may have different margins.  So we probably want to set an explicit margin for checkboxes when they appear inside the Preferences modal.

BamaCharanChhandogi:
> Thanks for raising this issue! It's a good issue but we need to be careful about how we implement the fix.
> 
> ### 1. It should not change the appearance of other elements.
> The same `preference__option` class name is also used on the radio options. We don't want to change the margins on all options. We only want to change the checkbox options. We could add additional class names to get greater specificity. We can also use [sibling selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator) to select option labels appearing alongside checkboxes.
> 
> ### 2. It should work the same for all browsers and devices.
> When I look at the existing UI in dev tools in Chrome (on Windows) I can see that the `margin` around the checkbox is coming from the user agent stylesheet, which is the built-in Chrome styles. image image
> 
> A different browser may have different margins. So we probably want to set an explicit margin for checkboxes when they appear inside the Preferences modal.

Hi @lindapaiste, I changed the class name, what next?

BamaCharanChhandogi:
@raclim please check

BamaCharanChhandogi:
@raclim check!

BamaCharanChhandogi:
@raclim  please review!

BamaCharanChhandogi:
@lindapaiste Please Check!

raclim:
Sorry it took a while to get to this! I think I'm still seeing the same error, and don't see a change in the checkbox alignment on my end. Since it's been a while since this PR was opened, I'm going to close this for now. I'm sorry that we couldn't get this in, but please feel free to open a new PR for this issue or check out some of the other ones!

PR #583, opened 2018-03-01
Remove accessibility CSS; Auto focus

MathuraMG:
This PR addresses #566 and #504 

We also noticed that there is a function in the reducer that can be completely removed - `startAccessibleSketch` . We were thinking of creating a new issue for this and clearing it out across the board.  Does that sound good?

@lm-n 
@cleezyITP

Before your pull request is reviewed and merged, make sure you

* [x] there are no linting errors -- `npm run lint`
* [x] code is in uniquely-named feature branch, and has been rebased on top of latest master. If you're asked to make more changes make sure you rebase onto master then too!
* [x] pull request is descriptively named and links to an issue number, i.e. `Fixes #123`

Thank you!

PR #2469, opened 2023-09-22
removed function startAccessibleSketch() solved issue #585

### LINKED ISSUE #585: Remove startAccessibleSketch()
#### Nature of issue?

- [ ] Found a bug
- [ ] New feature request

Remove `startAccessibleSketch` from reducer. 

@MathuraMG https://github.com/processing/p5.js-web-editor/pull/583

leftover19:
Fixes #585

Changes:
Removed the function startAccessibleSketch() 
Edited ide.js and Toolbar.js
Website worked fine without this function. Verified.

* has no linting errors (`npm run lint`)
* has no test errors (`npm run test`)
* is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* is descriptively named and links to issue number `#585`

### DIFF
--- a/client/modules/IDE/actions/ide.js
+++ b/client/modules/IDE/actions/ide.js
@@ -269,14 +269,6 @@ export function startSketch() {
   };
 }
 
-export function startAccessibleSketch() {
-  return (dispatch) => {
-    dispatch(clearConsole());
-    dispatch(startAccessibleOutput());
-    dispatch(startSketchAndRefresh());
-  };
-}
-
 export function stopSketch() {
   return (dispatch) => {
     dispatchMessage({
--- a/client/modules/IDE/components/Header/Toolbar.jsx
+++ b/client/modules/IDE/components/Header/Toolbar.jsx
@@ -4,12 +4,7 @@ import PropTypes from 'prop-types';
 import { useTranslation } from 'react-i18next';
 import { useDispatch, useSelector } from 'react-redux';
 import { Link } from 'react-router-dom';
-import {
-  openPreferences,
-  startAccessibleSketch,
-  startSketch,
-  stopSketch
-} from '../../actions/ide';
+import { openPreferences, startSketch, stopSketch } from '../../actions/ide';
 import {
   setAutorefresh,
   setGridOutput,
@@ -50,7 +45,6 @@ const Toolbar = (props) => {
         className="toolbar__play-sketch-button"
         onClick={() => {
           props.syncFileContent();
-          dispatch(startAccessibleSketch());
           dispatch(setTextOutput(true));
           dispatch(setGridOutput(true));
         }}


lindapaiste (changes requested):
Hey thanks for looking into this.  Honestly the Redux code is a complete mess and I've largely avoided opening that can of worms.

This particular function has to do with content that is only visible to screen readers so you're not going to see the changes just by looking at the app.  I'm pretty sure that the `dispatch(startAccessibleSketch());` which you deleted on line 53 should be replaced by `dispatch(startSketch());` instead.  I can explain more later this weekend.  It involves a hidden button so it's a bit 🤯 

I'm trying to understand the context of the discussion back in 2018.  It seems like at one point we were using `state.ide.isAccessibleOutputPlaying` to determine whether or not to render the accessibility text (hidden text describing the sketch).  Then we switched to using separate properties `state.preferences.textOutput` and `state.preferences.gridOutput`.  So the  `state.ide.isAccessibleOutputPlaying` property is no longer needed and there is some code related to it which can be deleted.  You can actually delete more than what you have removed here because we no longer need that state property in the reducer.

I'm on the fence about the keeping the actions that set it and those action name constants.  The way that we are dispatching of a bunch of actions in a row ([introduced here](https://github.com/processing/p5.js-web-editor/pull/583/files#diff-655f086d081b9ee40d8c3976580ec47bcf01d16cb2d517d8fcaac2f2f1f7f18e)) is weird and [not a best practice](https://redux.js.org/style-guide/#avoid-dispatching-many-actions-sequentially).  I actually think that it makes much more sense to `dispatch` a single `startAccessibleSketch` action which would simultaneously enable the text/grid output and start the sketch.  But that's getting into bigger changes that are potentially beyond the scope of this PR.

https://github.com/processing/p5.js-web-editor/blob/ca74574cd1830d900aad12665f15fc5f09251f14/client/modules/IDE/components/Header/Toolbar.jsx#L51-L56

https://github.com/processing/p5.js-web-editor/blob/ca74574cd1830d900aad12665f15fc5f09251f14/client/modules/IDE/pages/IDEView.jsx#L201-L203

client/modules/IDE/components/Header/Toolbar.jsx
```diff
@@ -50,7 +45,6 @@ const Toolbar = (props) => {
         className="toolbar__play-sketch-button"
         onClick={() => {
           props.syncFileContent();
-          dispatch(startAccessibleSketch());
```
lindapaiste, line 53: We need to continue starting the sketch here.  It's super confusing but this button on lines 49-61 is a hidden version that's shown on screen readers only. Lines 62-73 are the button that you see.  So removing this completely will prevent the sketch from playing on screen readers.

client/modules/IDE/actions/ide.js
```diff
@@ -269,14 +269,6 @@ export function startSketch() {
...
 }
 
-export function startAccessibleSketch() {
-  return (dispatch) => {
-    dispatch(clearConsole());
-    dispatch(startAccessibleOutput());
-    dispatch(startSketchAndRefresh());
-  };
-}
-
```
lindapaiste, line 279: I need to look into the full chain of actions which result from calling `startAccessibleSketch()` and see which are them are and are not necessary.  The `startAccessibleOutput()` is the one which can be removed but idk about `clearConsole()` and `startSketchAndRefresh()` and how this compares to calling the normal `startSketch()`.

It might be that we modify this function instead of removing it.

leftover19:
I understand, Thank you for such detailed explanation.

PR #2471, opened 2023-09-23
changed title for better seo

### LINKED ISSUE #2251: Better SEO title needed on collection details pages
### Increasing Access

- Help users find relevant collections via search engines
- Allow users to better identify what pages they have open in their browser when looking at the tab titles

When looking at the details page for an individual collection, for example https://editor.p5js.org/codingtrain/collections/bD7HTvWYL, the `<title>` meta tag is not very descriptive as it does not contain the name of the collection.

The current title is:
```
<title>p5.js Web Editor | codingtrain's collections</title>
```
A better title would be something along the lines of:
```
<title>p5.js Web Editor | Nature of Code Videos Chapter 3</title>
```
We could include the username, but it verges on too long depending on the collection name (it should be roughly 50-70 characters for SEO purposes but this one is 78 characters) 
```
<title>p5.js Web Editor | Nature of Code Videos Chapter 3 | Collection by codingtrain</title>
```

I personally prefer putting the website at the end of the title rather than at the start, but we put it at the start in all other places.  So that's a separate issue.

Akhilbisht798:
Fixes #2251

Changes:


client/modules/User/components/Collection.jsx
```diff
@@ -418,7 +416,7 @@ class Collection extends React.Component {
       >
         <article className="collection">
           <Helmet>
-            <title>{this.getTitle()}</title>
+            <title>{this.getTitle() + title}</title>
```
lindapaiste: The `title` variable will be `null` when the collection is still loading, and in that situation I think this will render `"p5.js Web Editor | null"`?  Because the `null` is cast to a `string` when using the `+` operator.

client/modules/User/components/Collection.jsx
```diff
@@ -181,12 +181,10 @@ class Collection extends React.Component {
...
-    if (this.props.username === this.props.user.username) {
-      return this.props.t('Collection.Title');
-    }
-    return this.props.t('Collection.AnothersTitle', {
-      anotheruser: this.props.username
-    });
+    const collectionTitle = this.props.t('Collection.Title');
+    const titleSplits = collectionTitle.split('|');
+    const title = titleSplits[0].concat('| ');
+    return title;
```
lindapaiste: We're basically just extracting the `"p5.js Web Editor"` portion of the title, right?  We don't have any property in the translations files which is just  `"p5.js Web Editor"` but I've wondered before if we should add one.  Like `t('Common.SiteName')` or something.  FYI.

Akhilbisht798: Oh, i was thinking same. Even searched for entire i8n for it. But though i might need to discuss that before putting in. So i stick with the safe option.
Okay now i will implement it.

client/modules/User/components/Collection.jsx
Akhilbisht798: Oh, Okay i though same and checked that, in almost all cases it was correct but still that might be a case .
So i think we might just render `Your Collection` and `Someone Collection` in case of null. To avoid title be null.

Akhilbisht798: Is this sound good to you?

Akhilbisht798:
@lindapaiste, Hey i have made changes according to your suggestion.

lindapaiste:
Hey, sorry for not keeping on top of this!  I was working on something else today which would benefit from your `Common.SiteName` translation key.

There's a bit of a problem with the title in the loading case because `getCollectionOwner` returns a title which already has the site name in it.  But then we join it with the site name so we end up with "p5.js Web Editor | p5.js Web Editor | My collections".

This PR is 95% there so I'm going to go ahead and fix it so that we can get it merged.

lindapaiste (approved):

Akhilbisht798:
Hey, Thanks for letting me about the problem. It was working properly last time when I checked, but I suppose your commit will take care of this problem.

Issue #2292, opened 2023-07-14
Server Error When Logging Out


raclim:
### p5.js version

1.6.0

### What is your operating system?

Mac OS

### Web browser and version

114.0.5735.198 (Official Build) (arm64) 

### Actual Behavior

Originally found by @lindapaiste with [added details and a potential solution.](https://github.com/processing/p5.js-web-editor/pull/2271#issuecomment-1633469530). The content of this issue is based off of her comment. 

Currently there is a 500 internal server error when logging out on localhost from the `http://localhost:8000/editor/logout` API endpoint. 

### Expected Behavior

Error does not appear on logout. 

### Steps:
1. Ensure that you've logged into your account for the p5.js Web Editor. 
2. Logout by clicking on the dropdown menu on the upper-right corner, and then click on "Log Out". 
3.  (I am using Chrome DevTools) Open your browser's Developer Tools and make sure you're in the "Console" panel, which is where the error will appear.

lindapaiste:
I'm reopening this because I am getting a 500 error when trying to log out on `localhost` and in production.

The error is suggesting that we need to use `express-session` middleware but we already are.  Something might be set up incorrectly.  I'm also wondering if #2473 might have put the actions in the wrong order, like maybe it's an error to call `req.logout` after the session has already been destroyed.  I have not investigated yet.

Triggered by this line:
https://github.com/processing/p5.js-web-editor/blob/1cf2dc3c8aa555110bab2b215544ca3abba92139/server/controllers/session.controller.js#L39

Error:
```
Error: Login sessions require session support. Did you forget to use `express-session` middleware?
    at SessionManager.logOut (C:\Users\Linda Paiste\PhpstormProjects\p5.js-web-editor\node_modules\passport\lib\sessionmanager.js:64:33)
    at IncomingMessage.logout (C:\Users\Linda Paiste\PhpstormProjects\p5.js-web-editor\node_modules\passport\lib\http\request.js:67:26)
    at C:\Users\Linda Paiste\PhpstormProjects\p5.js-web-editor\server\controllers\/session.controller.js:39:9
    at tryCatcher (C:\Users\Linda Paiste\PhpstormProjects\p5.js-web-editor\node_modules\bluebird\js\release\util.js:16:23)
    at Promise.successAdapter [as _fulfillmentHandler0] (C:\Users\Linda Paiste\PhpstormProjects\p5.js-web-editor\node_modules\bluebird\js\release\nodeify.js:23:30)
[...]

PR #2473, opened 2023-09-23
Fix: Logout Glitch

### LINKED ISSUE #2449: Login Session Still Persists Despite Logout

rajatmohan22:
Fixes: #2449

@lindapaiste I noticed your comments on the issue. You mentioned that you're not an expert on cookies, and I'm in the same boat 😄. However, after some investigation, here's what I've discovered:

**1. Why doesn't the error occur on localhost?**
Ans: Seemingly, browsers have different policies regarding cookies for different domains. Therefore, cookies and session management may behave differently compared to an external server ( This is actually evident if you open the networks tab on both localhost and the prod site and check the `status code` of the session.) This could lead to differences in how sessions are managed and why the issue doesn't occur locally.

**2. How do we fix it?**
Ans: Deleting the session. What's happening now is that, on logout, the session Id is changing. However, we want it gone!
Apparently Passport provides another function for this. It is called: `session.destroy`.

Please check it out, and let me know if any changes is required.

Here is a brief video/gif:

with `req.session.destroy`
[screen-capture (1).webm](https://github.com/processing/p5.js-web-editor/assets/77490034/65425538-cf13-4639-a76f-4b52aa060aae)

with `req.logout`:
[screen-capture (3).webm](https://github.com/processing/p5.js-web-editor/assets/77490034/9e0df71f-54a5-4d5e-892c-760f5368e046)

Changes:

### DIFF
--- a/server/controllers/session.controller.js
+++ b/server/controllers/session.controller.js
@@ -36,14 +36,12 @@ export function destroySession(req, res, next) {
       next(err);
       return;
     }
-    req.logout((err) => {
-      if (err) {
-        next(err);
+    req.logout((error) => {
+      if (error) {
+        next(error);
         return;
       }
-
       res.json({ success: true });
-      res.redirect('/login');
     });
   });
 }
--- a/server/controllers/session.controller.js
+++ b/server/controllers/session.controller.js
@@ -31,11 +31,19 @@ export function getSession(req, res) {
 }
 
 export function destroySession(req, res, next) {
-  req.logout((err) => {
+  req.session.destroy((err) => {
     if (err) {
       next(err);
       return;
     }
-    res.json({ success: true });
+    req.logout((err) => {
+      if (err) {
+        next(err);
+        return;
+      }
+
+      res.json({ success: true });
+      res.redirect('/login');
+    });
   });
 }


rajatmohan22:
@lindapaiste If the uploaded video isn't clear, I can create a brief YouTube video explanation and share the link here

rajatmohan22:
@lindapaiste Have a look at both the videos:

- Using `req.logout` alone creates a new session.

- Using `req.session.destroy` deletes the session permanently.

lindapaiste:
Thanks for researching this! At first glance it looks good.

rajatmohan22:
@lindapaiste,Is there a way to verify this? because it still runs on localhost even in production mode. So I can't reproduce the issue.

rajatmohan22:
Maybe I must deploy on Heroku to verify ?

rajatmohan22:
@lindapaiste 
I found this link online: https://copyprogramming.com/howto/either-req-logout-or-req-session-destroy-does-not-work
It discusses the distinctions between using `req.logout` and `req.session.destroy`. (Please note, I made a minor mistake earlier – `req.session.destroy` is actually an Express function, not a Passport one.)

I want to highlight a small paragraph from the website. Which says:  It is not recommended to implement logout from the server side, as the token is not stored on the server. Therefore, it's best to let clients handle the logout by deleting the token. Additionally, it's important to avoid storing tokens in the database unless necessary, and if they must be stored, they should be hashed like passwords. Storing tokens in a database can be risky if there is a security vulnerability, as it could give intruders too much power.

Currently, we're handling the hashing part, but we're not deleting the token on the server upon logout. As I mentioned earlier, we're modifying it instead. Don't think it would be a good modification to implement in the codebase?

rajatmohan22:
@lindapaiste, this works. I deployed a test app. Repo link: https://github.com/rajatmohan22/AuthTest1.

Additionally, we can include HTML tags to clear the browser cache. Do you want me to do anything else to test it out particularly?

rajatmohan22:
@raclim could you verify this?

raclim (approved):
Thanks for working on this! I also think this looks good to me for now.

rajatmohan22:
@raclim Thank you! appreciate it.

lindapaiste:
@raclim We need to roll this back.

I had not run it before (FYI, if I say something like "At first glance it looks good." that usually means I looked at the code but didn't run it).  Now that I am running it I am getting a 500 error from the API when logging out, [detailed here](https://github.com/processing/p5.js-web-editor/issues/2292#issuecomment-1791419467). 

The passport logout function throws an error if there is no `req.session`.  This PR is calling `req.session.destroy` first and `req.logout` second.  Therefore the logout fails because the session is gone.

rajatmohan22:
Refer to #2473 for detailed error explanations. I'm also super open to providing a line-by-line breakdown. I believe the code is error-free. However, we can also consider using the 'destroy' method **without** Passport's 'logout' if desired.

raclim:
After giving another look at it I'm also getting the same 500 error here. Sorry I think I didn't give it as thorough of a glance as I should've at the time!

I'm going to revert this PR for now, and we can take a look at the new one submitted for this.

PR #2487, opened 2023-10-03
fix: changed-ConsoleInput-to-function-based

umangutkarsh:
Ref #2358

Changes: Converted legacy React class component to function component (ConsoleInput component)


lindapaiste:
I don’t have time to review this right now, but have you run this on localhost and confirmed that it works when you type a line in the console?  I suspect that it does not work, as your code is very similar to a previous PR that had a lot of problems.  Honestly I think we are going to leave this one as a class component because it is too complicated and confusing to convert it.  Please read my comments on #2409.

PR #2491, opened 2023-10-04
Done with the problem faced of loosing screen.

### LINKED ISSUE #2483: Limit Horizontal Scrolling of Output Page to Prevent Loss at Screen Edge
### p5.js version

_No response_

Windows

### Web browser and version

Chrome

### Actual Behavior

Currently, when users drag the output page to the right, it continues to scroll indefinitely, which can result in the output page disappearing off-screen.

https://github.com/processing/p5.js-web-editor/assets/127239756/d5c0455b-5cd7-458a-99f9-13223641788c

### Expected Behavior

The output page in the code compiler should not be draggable to infinity when the user tries to scroll horizontally. Instead, it should have a limit and stop scrolling at a certain point, for example, at 95% of the screen width. This will prevent users from accidentally losing access to the output page.

https://github.com/processing/p5.js-web-editor/assets/127239756/1ca391da-3254-48ca-bfb0-0a221de3a596

Or, after reaching 100% of the screen width, there should be a noticeable separation or thick strip, clearly indicating the end of the output and distinguishing it from the input section. This design improvement will ensure users have full control over the output while maintaining a clear visual separation between the input and output sections.

### Steps to reproduce

1. Open the code compiler.
2. Execute some code that generates a lengthy output.
3. Try to scroll the output page horizontally to the right.

perminder-17:
Fixes #2483 

Changes: I have make the preview screen where we cannot loose it. Please check the video.

https://github.com/processing/p5.js-web-editor/assets/127239756/a6bef211-1ebf-43ad-8dd0-4eda7c5d170e

I have done my best to change it.


perminder-17:
U may check @lindapaiste whenever u are free.

lindapaiste (changes requested):
It looks like you're combining two approaches here.  You are setting a `maxSize` state but not ever using it.  Instead you're setting the `maxSize` of the `SplitPane` directly from the `window.innerWidth`.

The code that you have here **will** actually work, but in a way that is very roundabout.  Resizing the window calls the `resize` event handler, which calls `setMaxSize`.  That changes the state of the component and forces a re-render, so the props of the `SplitPane` will get re-evaluated using the current value of `window.innerWidth`.

To keep the code cleaner I recommend the following changes:

1. Remove `window.innerWidth` as a dependency of the `useEffect`.  You only need to add the `resize` event listener once.  The dependencies should be `[]` or `[setMaxSize]` (both will only evaluate the effect once).  If that is confusing, consider that the handler could be written as `const updateInnerWidth = (e) => { setMaxSize(e.target.innerWidth);  };`.
2. Assign the state to a variable `maxSize` and use this to control the `SplitPane` : `maxSize={maxSize * 0.965}`.
3. As I'm writing this, I realize the issue of having an `undefined` value the first render and now I suspect that you probably wrote it the way that you did to avoid getting an `NaN` value from multiplication with `undefined`.  How about we set the initial value of the state from the `window`?  `const [maxSize, setMaxSize] = useState(window.innerWidth);`

perminder-17:
I sincerely apologize for the delay @lindapaiste . Upon reevaluating my approach, I have recognized the shortcomings and made the necessary adjustments as per your guidance. Thanks for your suggestion.

raclim (approved):
Thanks for working on this! I'm going to merge this in as it seems to resolve the issue for now. Please feel free to add any further comments or changes before the next release around next Tuesday, January 16.

perminder-17:
Thanks for the merge @raclim and I will definetly add any further changes before 16th of Jan. Really thanks for considering me. :)

PR #2507, opened 2023-10-16
Fix for Build Breaking When Deployed to Release

raclim:
A majority of these changes are temporary solutions to address current issues with deploying a production build to Release. 

Changes:
- Reinstalls `@pmmmwh/react-refresh-webpack-plugin` as a dependency rather than a devDependency to temporarily address error where Release cannot find this package. (The root cause appears to be an issue where Babel and Webpack appear to interpret the build as different environments i.e `development` vs `production`. Some possible solutions were found but I had some difficulty implementing them. Since this temporary solution doesn't seem too severe and to get the next release out soon, I'm thinking this could probably be addressed in a separate issue?)

- Also bumps up the above package a patch release number (`0.5.10` -> `0.5.11`).

- Removes lines 14 & 97 (which dispatches the action `stopSketch()`) from `client/modules/IDE/pages/IDEView.jsx` to address an error where the target origin does not match the recipient for the iFrame. 

Changes that have been updated with @lindapaiste's suggestions commented below:

- Comments out lines related to `isOverlayVisible` and `setIsOverlayVisible` in `client/modules/IDE/pages/IDEView.jsx` to address an issue where the `PreviewFrame` was not receiving any mouse/keyboard interactions due to setting `display: block` on line 196. I wasn't sure what the exact vision was for here (I'm sorry if it's already noted somewhere!) but I decided to temporarily comment this out for now!

- Removes lines 1 & 52 from `client/routes.jsx` to address an error where the `component` property for `Route.proptypes` cannot be set (this also appears exclusively in Release for me). 


lindapaiste:
> * Comments out lines related to `isOverlayVisible` and `setIsOverlayVisible` in `client/modules/IDE/pages/IDEView.jsx` to address an issue where the `PreviewFrame` was not receiving any mouse/keyboard interactions due to setting `display: block` on line 196. I wasn't sure what the exact vision was for here (I'm sorry if it's already noted somewhere!) but I decided to temporarily comment this out for now!

The intention of the overlay stuff was to prevent mouse events from firing on the sketch while dragging the resize panel.  The mistake here (which I should have caught!) is that the initial state should be `false` instead of `true`.  In the previous class-component version we were directly manipulating HTML to do this.
https://github.com/processing/p5.js-web-editor/blob/c58ff78eebdb2b4d258aa6bc72d6bbdfbfee81ba/client/modules/IDE/pages/IDEView.jsx#L315-L320

I think if you change the initial state to `false` then you should be able to add those lines back in.

There's probably a much cleaner way to implement the intended behavior using [`pointer-events: none`](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events) in the CSS, rather than displaying an invisible overlay.

lindapaiste:
> * Removes lines 1 & 52 from `client/routes.jsx` to address an error where the `component` property for `Route.proptypes` cannot be set (this also appears exclusively in Release for me).

I don't recall why I wrote this as two separate assignments.  I wonder if it works this way?

```
Route.propTypes = { 
  ...RouterRoute.propTypes,
  component: PropTypes.elementType.isRequired
};
```
instead of
```
Route.propTypes = { ...RouterRoute.propTypes };
Route.propTypes.component = PropTypes.elementType.isRequired;
```

raclim:
Sounds good, thanks so much for the suggestions! I'll try those out and if all looks good, will hopefully be able to deploy before the end of the day!

PR #2543, opened 2023-10-29
fix : Folders auto-expand on save

### LINKED ISSUE #2530: Folders auto-expand on save
### p5.js version

p5.js v1.7.0 July 10, 2023

Windows

### Web browser and version

1.59.120 Chromium: 118.0.5993.88 (Official Build) (64-bit) 

### Actual Behavior

When I save my work in the the p5.js Web editor (or an auto-save occurs), all the folders expand to reveal their contents, even after I contract them.

### Expected Behavior

Folders should retain their collapsed or expanded state when a save or auto-save occurs.

### Steps:
1. Collapse a folder. (You may need to create one and upload files into it first.)
2. Save your work (or wait for an auto-save to occur).
3. Observe that the folder you just collapsed has expanded again.

7070Shreyash:
Fixes #2530 
Changes:
This PR fixes the folder auto-expand bug by appropriately modifying the files state in redux , adding necessary variables to capture the previous state before saving .

[Screencast from 30-10-23 02:08:20 AM IST.webm](https://github.com/processing/p5.js-web-editor/assets/117765947/16d2ce2d-cddf-4741-9746-3e4f0b092243)


lindapaiste (commented):
There are definitely some issues here with regards to mutating objects.  I also suspect that you are changing code in more places than what's necessary.  Possibly all that you need are the 6 lines in the `case ActionTypes.SET_PROJECT` reducer (with some tweaks).  

I have other thoughts that I will put in a comment on the issue.

client/modules/IDE/reducers/files.js
```diff
@@ -159,8 +160,22 @@ const files = (state, action) => {
...
+          file.isFolderClosed = corrospondingObj.isFolderClosed;
+        }
+        return file;
+      });
       return setFilePaths(action.files);
     case ActionTypes.SET_PROJECT:
+      action.files = action.files.map((file) => {
+        const corrospondingObj = state.find((obj) => obj.id === file.id);
+        if (corrospondingObj && corrospondingObj.fileType === 'folder') {
+          file.isFolderClosed = corrospondingObj.isFolderClosed;
```
lindapaiste: You can avoid mutating the action like this:
```
case ActionTypes.SET_PROJECT:
    const files = action.files.map((file) => {
        const corrospondingObj = state.find((obj) => obj.id === file.id);
        if (corrospondingObj && corrospondingObj.fileType === 'folder') {
            const isFolderClosed = corrospondingObj.isFolderClosed || false;
            return { ...file, isFolderClosed };
        }
        return file;
    });
    return setFilePaths(files);
```

FYI there are some mutations in the existing code in this reducer😢.  But let's not add any new ones.

I have to look at the code to remind myself what the difference is between `NEW_PROJECT` and `SET_PROJECT`.  If we need to use the same code in both places then we can move it to a helper function.  But I suspect that it's not needed for the `NEW_PROJECT` case because if it's a new project then the `corrospondingObj` would always be `undefined`.

lindapaiste: >I suspect that it's not needed for the NEW_PROJECT case because if it's a new project then the corrospondingObj would always be undefined.

I was wrong.  It is needed in both places.  But it seems that `NEW_PROJECT` and `SET_PROJECT` are identical so now I'm wondering why we have two actions that do the same thing.

7070Shreyash: Hey @lindapaiste , Thank you for your invaluable and clear response . I have made the suitable changes.

lindapaiste (approved):

PR #2597, opened 2023-11-12
Fix #2499 - typo in error "please enter a email"

### LINKED ISSUE #2499: Translation Typos in Sign Up Form Text
### p5.js version

2.7.3

Mac OS

### Web browser and version

Chrome and 117.0.5938.132

### Actual Behavior

Validation typo in sign up page `email`
Screenshot 2023-10-13 at 12 08 07 PM

### Expected Behavior

This should state `Please enter an email` instead of `Please enter a email`
Screenshot 2023-10-13 at 12 11 50 PM

### Steps:
1. Open p5.js web editor
2. Click on sign up 
3. Click on email input
4. Notice the typo in email's validation

--------------------------------------------------------

To Note: Also similar to https://github.com/processing/p5.js-web-editor/issues/2560, raised by @apsinghdev. Referenced here to help organize issues.

lindapaiste:
Fixes #2499 

### Changes:
- Delete the untranslated form error "Please enter a ____" in components `AccountForm` and `SignupForm`.

### Explanation:

The primary validation of this form comes from a form-level validation function.  This function already checks for empty email and password and returns translated error messages.
https://github.com/processing/p5.js-web-editor/blob/a1a5efd44f64f4a449f07379098dcdfe049558eb/client/utils/reduxFormUtils.js#L33-L35
Additionally, we have a field-level validation function which calls the `'/signup/duplicate_check'` API endpoint.
https://github.com/processing/p5.js-web-editor/blob/a1a5efd44f64f4a449f07379098dcdfe049558eb/client/modules/User/components/SignupForm.jsx#L10-L32
We don't need to call the duplicate check API if the field is empty.  So we check for this and return early.  But the message that we are returning there is not translated and not semantically correct ("a email" instead of "an email").

We can instead return an empty string `''` (or just `return;` but eslint prefers the consistent string return value).  Returning a falsey value instead of an error message means that this check passes and does not set an error, but that's okay because it is caught later on by the form-level validation function, `validateSignup` or `validateSettings`, where it is handled properly.

### Future Tasks:
- `duplicateCheck` function is the same in both files.
- Message "This email is already taken." comes from the server and is not translated.
- We call the API more than is necessary because we re-query when other fields change (not just the field being checked).  This is the standard behavior of the form package but there are some workarounds.  See https://github.com/final-form/react-final-form/issues/170
- We my want to throttle the API calls instead of calling on every keystroke.
- We should delay the duplicate email check until the email passes the "invalid email" regex check.  We don't need to check while they are typing "linda..." because we wouldn't accept that email anyways.  We only need to check after they have typed the "@gmail.com" part.


raclim (approved):
This looks good to me so far, thanks for the context and for pointing out some other issues around it! Maybe we'd also want them in a separate issue at some point 😄

PR #2622, opened 2023-11-19
Use react-router for legal page tabs.

### LINKED ISSUE #2621: Discussion: `react-tabs` vs. `react-router` links
### Increasing Access

Using standard HTML `<a>` tags for links is more semantically correct and allows for standard right-click behavior.

There are multiple places in the app where we use tabs to switch between content.  Some of these are coded using the `react-tabs` package, while others use `<Link>` elements from `react-router-dom`.

It is my opinion that we should use the `react-router-dom` approach in situations where the tabs have different URLs but I want to open this up for discussion.

### Advantages of `react-router-dom`
* We can use the router to control the content on the page, instead of manually switching based on `pathname`
https://github.com/processing/p5.js-web-editor/blob/185a0b1ed3cef02957bd9bc04d25b33c33eee11d/client/modules/Legal/pages/Legal.jsx#L37-L45
* We don't need any callback function when changing tabs.
https://github.com/processing/p5.js-web-editor/blob/185a0b1ed3cef02957bd9bc04d25b33c33eee11d/client/modules/Legal/pages/Legal.jsx#L47-L59
* We can automatically highlight the current tab using the `NavLink` component (instead of `Link`) https://reactrouter.com/en/main/components/nav-link
* Links render as HTML `<a>` elements which is more semantically correct and allows for standard right-click behavior.
image
* Screen readers understand links, and they can be focused using Tab navigation.

### Current Usage
image
Legal page - uses `react-tabs`
https://github.com/processing/p5.js-web-editor/blob/185a0b1ed3cef02957bd9bc04d25b33c33eee11d/client/modules/Legal/pages/Legal.jsx

image
Dashboard page - uses `react-router-dom` (with manual routing and highlighting that I would like to clean up)
https://github.com/processing/p5.js-web-editor/blob/185a0b1ed3cef02957bd9bc04d25b33c33eee11d/client/modules/User/components/DashboardTabSwitcher.jsx

image
Preferences modal - uses `react-tabs`
This would stay as-is because there is no URL change involved
https://github.com/processing/p5.js-web-editor/blob/185a0b1ed3cef02957bd9bc04d25b33c33eee11d/client/modules/IDE/components/Preferences/index.jsx

lindapaiste:
Fixes #2621 

The primary purpose is to control the routing of the Legal page declaratively (with routes) rather than imperatively (`history.push`).  The `react-tabs` package is fine on its own but doesn't work elegantly alongside `react-router`.

Changes:

- `RouterTab`
  - Extracted the `Tab` component from `DashboardTabSwitcher.jsx` into a reusable `RouterTab` component.
  - Remove the `isSelected` prop and use the `react-router` `NavLink` component which automatically detects if a link is for the current page.
- `_dashboard-header.scss`
  - The `NavLink` means that the `dashboard-header__tab--selected` class is now applied to the child `<a>` rather than the parent `<li>`, so I had to move around some styles in the `.scss` file to keep things correct.
- Legal
  - Use the `RouterTab` instead of the `react-tabs` package.
  - Made the `Legal` component into a container that renders the correct content based on props, rather than switching children.
   - Move the policy fetch `useEffect` and the `Helmet` up into the parent since it is the same for all tabs.
  - _Add a `Loader` while loading._ // TODO: this is not visible and should probably be a child of the PolicyContainer.
  - `PrivacyPolicy` etc. are now specific instances of `Legal`.
- `routes.jsx`
  - Render the specific page for each URL.  This eliminates the need to look at the path inside the `Legal` component.
- `DashboardTabSwitcher`
  - This should look identical but there are some minor changes to the HTML structure:
    - I swapped the `div` and `ul` tags so that the `li` elements are direct children of the `ul`.
    - There is no more `h4` in each tab.  Just `<li><a/></li>`.  I don't think this matters?  But now that I'm writing this I think I might put it back?
    - The active tab uses an `a` instead of a `span`.


PiyushChandra17 (approved):
Looks good 👍

raclim (approved):
This is great, thanks so much for working on this!

PR #2628, opened 2023-11-22
fix autocapitalize in mweb safari

### LINKED ISSUE #2623: mWeb(safari): First letter gets automatically capitalized in username or email input
### p5.js version

2.9.3

Mac OS

### Web browser and version

_No response_

### Actual Behavior

First letter gets automatically capitalized in username or email input in login page without pressing shift button on `iOS - mWeb(safari)`

first letter

### Expected Behavior

First letter should not get capitalized automatically in username or email input 

### Steps:
1. Open the editor web app in iOS - mWeb(safari)
2. Go to the login page
3. Tap on username or email
4. Verify that the first character gets capitalized automatically without pressing `shift`

PiyushChandra17:
Fixes #2623 

Changes:

Added `autoCapitalize` attribute and set it to `none`

fix-autocapitalize


lindapaiste:
Awesome!  I didn't know about this `autocapitalize` attribute.  Now I am looking at the [specs on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize) and I see this:
> The `autocapitalize` attribute never causes autocapitalization to be enabled for an [`<input>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) element with a [type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#type) attribute whose value is `url`, `email`, or `password`.

And I see that we have these two inputs set to `type="text"` instead of the more specific `type="email"`.  So I think we should change the `type`.  We can keep the `autocapitalize` to be safe, but it seems like it's not strictly needed if we have the correct `<input>` type.

PiyushChandra17:
@lindapaiste Even if we set to `type="email"` doesn't seem to work, if we remove `autoCapitalize` attribute. Can you please verify. I think same is happening on sign up `username` form field aswell where we have specified `type="text"`. I think `autoCapitalize` is mandatory here.

lindapaiste:
> @lindapaiste Even if we set to `type="email"` doesn't seem to work, if we remove `autoCapitalize` attribute. Can you please verify. I think same is happening on sign up `username` form field aswell where we have specified `type="text"`. I think `autoCapitalize` is mandatory here.

Sorry to drag you in circles but now I realize that you had it right the first time.  I didn't see that the login input accepts email OR username, so we should stay with `type="text"` and `autoCapitalize="none"` because the input is not always an email address.

On the sign up form there is an email field which is always an email, lines 82-89.  That one has `type="email"` so I'm not sure if we need the `autoCapitalize` there or not.  I think not?

PiyushChandra17:
@lindapaiste Yes, we don't need `autoCapitalize` here:

https://github.com/processing/p5.js-web-editor/blob/f725abd1505a1fecd8b58ed2f4a57367a6051d27/client/modules/User/components/SignupForm.jsx#L81-L88

soniyadotp:
is this issue closed?

raclim (approved):

PR #2688, opened 2023-12-04
Converting Editor component from class component to functional component

### LINKED ISSUE #2687: Editor Component should be changed to Functional Component
### Increasing Access

It is important to update components according to new Functional components.

The component uses useRef for setting a reference to the Codemirror instance which can be used in any useEffect and function. I am referencing a pull request for these issue.

Typicalcodes:
Fix #2687

Changes: Changed a important Component into New Functional Component. I have used react hooks .

### DIFF
--- a/client/modules/IDE/components/Editor/index.jsx
+++ b/client/modules/IDE/components/Editor/index.jsx
@@ -1,7 +1,5 @@
 // TODO: convert to functional component
-
-import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useCallback, useEffect, useMemo, useRef } from 'react';
 import CodeMirror from 'codemirror';
 import Fuse from 'fuse.js';
 import emmet from '@emmetio/codemirror-plugin';
@@ -9,7 +7,7 @@ import prettier from 'prettier/standalone';
 import babelParser from 'prettier/parser-babel';
 import htmlParser from 'prettier/parser-html';
 import cssParser from 'prettier/parser-postcss';
-import { withTranslation } from 'react-i18next';
+import { useTranslation } from 'react-i18next';
 import StackTrace from 'stacktrace-js';
 import 'codemirror/mode/css/css';
 import 'codemirror/mode/clike/clike';
@@ -23,7 +21,6 @@ import 'codemirror/addon/fold/comment-fold';
 import 'codemirror/addon/fold/foldcode';
 import 'codemirror/addon/fold/foldgutter';
 import 'codemirror/addon/fold/indent-fold';
-import 'codemirror/addon/fold/xml-fold';
 import 'codemirror/addon/comment/comment';
 import 'codemirror/keymap/sublime';
 import 'codemirror/addon/search/searchcursor';
@@ -39,9 +36,9 @@ import 'codemirror-colorpicker';
 import { JSHINT } from 'jshint';
 import { CSSLint } from 'csslint';
 import { HTMLHint } from 'htmlhint';
-import classNames from 'classnames';
 import { debounce } from 'lodash';
-import { connect } from 'react-redux';
+import classNames from 'classnames';
+import { useDispatch, useSelector } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import MediaQuery from 'react-responsive';
 import '../../../../utils/htmlmixed';
@@ -54,9 +51,7 @@ import '../../../../utils/codemirror-search';
 import beepUrl from '../../../../sounds/audioAlert.mp3';
 import RightArrowIcon from '../../../../images/right-arrow.svg';
 import LeftArrowIcon from '../../../../images/left-arrow.svg';
-import { getHTMLFile } from '../../reducers/files';
 import { selectActiveFile } from '../../selectors/files';
-
 import * as FileActions from '../../actions/files';
 import * as IDEActions from '../../actions/ide';
 import * as ProjectActions from '../../actions/project';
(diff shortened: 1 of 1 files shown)


client/modules/IDE/components/Editor/index.jsx
```diff
@@ -141,546 +228,443 @@ class Editor extends React.Component {
...
-      // then need to make CodeMirror documents
-      this.initializeDocuments(nextProps.files);
-    }
-    if (this.props.files.length !== nextProps.files.length) {
-      this.initializeDocuments(nextProps.files);
+  }, [mergedState.autocompleteHinter]);
+
+  useEffect(() => {
+    if (file.id) {
+      prevProps.current = {
```
lindapaiste: looks like you're setting `prevProps.current` on every render down on line 556 so I'm not sure if you need this effect.

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -81,36 +76,128 @@ window.HTMLHint = HTMLHint;
...
+      EditorAccessibilityActions,
+      FileActions,
+      ProjectActions,
+      IDEActions,
+      PreferencesActions,
+      UserActions,
+      ConsoleActions
+    ),
+    dispatch
+  );
```
lindapaiste: I wouldn't bother binding anything here.  I would call `dispatch(clearLintMessage())` etc. directly.

There could potentially be a slight benefit to binding if you were to do it inside a `useMemo` such that it only gets executed once.  In that case you would have a stable prop for the places where you pass an action creator down like `onClick={funcprop.expandSidebar}`.  But with the way that it is currently you'll get a new function on each render regardless, so you can write `onClick={() => dispatch(expandSidebar())}` and it's the same as what you have now.

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -81,36 +76,128 @@ window.HTMLHint = HTMLHint;
...
+      mode = 'application/json';
+    } else if (fileName.match(/.+\.(frag|glsl)$/i)) {
+      mode = 'x-shader/x-fragment';
+    } else if (fileName.match(/.+\.(vert|stl)$/i)) {
+      mode = 'x-shader/x-vertex';
+    } else {
+      mode = 'text/plain';
+    }
+    return mode;
+  }
```
lindapaiste: I would move this function outside of the component since it does not use any variables of the component.

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -81,36 +76,128 @@ window.HTMLHint = HTMLHint;
...
+  function tidyCode() {
+    const mode = cm.current.getOption('mode');
+    if (mode === 'javascript') {
+      prettierFormatWithCursor('babel', [babelParser]);
+    } else if (mode === 'css') {
+      prettierFormatWithCursor('css', [cssParser]);
+    } else if (mode === 'htmlmixed') {
+      prettierFormatWithCursor('html', [htmlParser]);
+    }
+  }
```
lindapaiste: This could potentially be moved outside of the component as well, it you were to take the CodeMirror instance as an argument.

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -81,36 +76,128 @@ window.HTMLHint = HTMLHint;
...
-  componentDidMount() {
-    this.beep = new Audio(beepUrl);
-    // this.widgets = [];
-    this._cm = CodeMirror(this.codemirrorContainer, {
-      theme: `p5-${this.props.theme}`,
-      lineNumbers: this.props.lineNumbers,
+  // Inialize main Content
+  const beep = new Audio(beepUrl);
+
+  const updateLintingMessageAccessibility = debounce((annotations) => {
```
lindapaiste, line 192: I wonder how the debouncing works here since we're recreating this function on every render.  Does it do anything?  I would suspect not.  This is the sort of thing that make this component 🤯.

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -141,546 +228,443 @@ class Editor extends React.Component {
...
+      mergedState.autocloseBracketsQuotes
+    );
+    cm.current.setOption('lineNumbers', mergedState.lineNumbers);
+    cm.current.setOption('lineWrapping', mergedState.linewrap);
+  }, [
+    mergedState.theme,
+    mergedState.autocloseBracketsQuotes,
+    mergedState.lineNumbers,
+    mergedState.linewrap
+  ]);
```
lindapaiste, line 256: This looks good ✔️

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -141,546 +228,443 @@ class Editor extends React.Component {
...
+          docs[fileitem.id] = CodeMirror.Doc(
+            fileitem.content,
+            getFileMode(fileitem.name)
+          ); // eslint-disable-line
         }
-      }, 1000)
-    );
+      });
+      cm.current.swapDoc(docs[file.id]);
+    }
```
lindapaiste: I think this needs to be in a separate effect because this effect does not have a dependency on `files`.

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -141,546 +228,443 @@ class Editor extends React.Component {
...
+
+    // eslint-disable-next-line react/prop-types
+    props.provideController({
+      tidyCode,
+      showFind,
+      showReplace,
+      getContent
+    });
```
lindapaiste, line 316: Seems like this is duplicated on line 550.

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -141,546 +228,443 @@ class Editor extends React.Component {
...
+    cm.current.on('change', handleChange);
+    return () => {
+      if (cm.current) {
+        cm.current.off('change', handleChange);
       }
-    }
-    if (this.props.fontSize !== prevProps.fontSize) {
-      this._cm.getWrapperElement().style[
+    };
+  }, [file.id, mergedState.autorefresh, mergedState.isPlaying]);
```
lindapaiste: This looks good ✔️ .  It seems to have all of the necessary dependencies and I love that you are removing the previous `handleChange` when setting the new one.

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -141,546 +228,443 @@ class Editor extends React.Component {
...
-    const editorHolderClass = classNames({
-      'editor-holder': true,
-      'editor-holder--hidden':
-        this.props.file.fileType === 'folder' || this.props.file.url
-    });
-
-    return (
+    }
+    updatedom();
+  }, [file.id]);
```
lindapaiste: You examine so many `mergedState` variables inside the `updatedom` function, but you are only running that function when the `file.id` changes.  That's going to be a problem.

lindapaiste (changes requested):
I've added some first thoughts with lines and sections that are potential problems.  I haven't run your branch yet.

As I said in the issue this conversion is extremely complex and it probably takes more effort than it's worth to get it right.

client/modules/IDE/components/Editor/index.jsx
Typicalcodes: Yes I am thinking same at this time. But I forgot why I have done this. I will figure out and change this soon.

client/modules/IDE/components/Editor/index.jsx
Typicalcodes: See my latest i have changed lot of things.

client/modules/IDE/components/Editor/index.jsx
Typicalcodes: i have fixed this in new commit.

PR #2693, opened 2023-12-05
dropdown menu positioned properly

### LINKED ISSUE #2678: Clicking `hello! username` user-menu dropdown causes the console to move when the username is short
### p5.js version

2.9.3

None

### Web browser and version

Safari and Version 16.6

### Actual Behavior

Inconsistency in `hello! username` user-menu dropdown in `Safari` web browser

https://github.com/processing/p5.js-web-editor/assets/47579287/60fd5867-e6e7-46f9-ba07-1c585e7a0246

### Expected Behavior

user-menu dropdown should be consistent across all web browsers, currently the page gets slightly upwards when user clicks on the dropdown or hovering to translation dropdown

### Steps:
1. Open the editor web app in `safari` web browser
2. Click on `hello! username` dropdown 
3. Notice the inconsistency 
4. Same is reproducible when user click the dropdown > hovers to language translation

AdityaNik:
Fixes #2678 

Changes:
Adjusted the styling of components to resolve the horizontal scrollbar issue.
Modified the width of '.nav__dropdown' and padding of the '.nav__items-right' to ensure it accommodates content without overflow.

**Before:**

p1

**After:**

image


lindapaiste:
This will make the dropdown skinnier in all instances and cause "My Collections" to wrap.  I don't think this is the approach that we want because it will be a downgrade from what we have currently when the username is long.

If we do go with something along these lines then I would recommend using a `min-width` on the dropdown rather than a `padding-right` so that we are only adding the extra space when it is actually necessary (for short usernames but not for long ones).

AdityaNik:
@lindapaiste thank you for the response.
Now, I have used 'floating-ui' for positioning dropdown  plz review

client/components/Nav/NavDropdownMenu.jsx
```diff
@@ -22,8 +23,28 @@ export function useMenuProps(id) {
 function NavDropdownMenu({ id, title, children }) {
   const { isOpen, handlers } = useMenuProps(id);
 
+  let floatingProps;
+  if (id === 'file' || id === 'edit' || id === 'sketch' || id === 'help') {
```
lindapaiste: I would stay away from hard-coding which menus get which position.  If we want to support both directions then we can use a prop like `align="left"`.

That said, right now all menus are positioned from the bottom left. So I think we can use the same `placement: 'bottom-start'` for all.

The `floating-ui` package will handle shifting it over if it doesn't fit within the window.  I think we want the `shift` middleware for that?
https://floating-ui.com/docs/useFloating#visibility-optimizers
https://floating-ui.com/docs/shift

client/components/Nav/NavDropdownMenu.jsx
```diff
@@ -22,8 +23,28 @@ export function useMenuProps(id) {
...
+    floatingProps = {
+      whileElementsMounted: autoUpdate,
+      placement: 'bottom-start',
+      strategy: 'absolute'
```
lindapaiste: I'm looking at the docs and this is the correct `strategy` ✔️

client/components/Nav/NavDropdownMenu.jsx
```diff
@@ -22,8 +23,28 @@ export function useMenuProps(id) {
 function NavDropdownMenu({ id, title, children }) {
   const { isOpen, handlers } = useMenuProps(id);
 
+  let floatingProps;
+  if (id === 'file' || id === 'edit' || id === 'sketch' || id === 'help') {
+    floatingProps = {
+      whileElementsMounted: autoUpdate,
```
lindapaiste: We need to be aware of this warning from [the docs](https://floating-ui.com/docs/useFloating#whileelementsmounted).

> Warning
> Only use this option if your floating element is conditionally rendered, not hidden with CSS. Use an effect to call and clean up `autoUpdate` in the latter scenario.

We are currently using CSS to show and hide the menu:
https://github.com/processing/p5.js-web-editor/blob/f21b84772c85927f8016d7c23aef7ad1a5a41b64/client/styles/components/_nav.scss#L108-L114

I personally don't like using CSS to hide things so I would be fine if we switched to conditional rendering and kept the `autoUpdate`.

lindapaiste (changes requested):
This is looking good.  I need to run your branch and see it in action and also I need to familiarize myself with the various settings of `floating-ui`.

client/components/Nav/NavDropdownMenu.jsx
AdityaNik: conditional rendering would be a better option.

client/components/Nav/NavDropdownMenu.jsx
AdityaNik: using `shift({ mainAxis: true })` is working

client/components/Nav/NavDropdownMenu.jsx
AdityaNik: `{isOpen && (
        <ul
          className="nav__dropdown"
          ref={refs.setFloating}
          style={floatingStyles}
        >
          <ParentMenuContext.Provider value={id}>
            {children}
          </ParentMenuContext.Provider>
        </ul>
      )}`

conditionally rendering is working, but the `nav_dropdown-item` is not navigating on click.
@lindapaiste can you please help me with this...

AdityaNik: @lindapaiste any suggestions on this one

AdityaNik:
@lindapaiste plz review

raclim:
Thanks for your work on this! I'm personally not entirely sure about this addition yet, and found that the issue still persists after running your branch. Since it's been a while since this PR was submitted, I'm going to close this for now. I'm really sorry that we couldn't get this in, but please feel free to reopen this PR with any updates or check out our other issues!

client/components/Nav/NavDropdownMenu.jsx
PiyushChandra17: @lindapaiste I think `placement: bottom` makes more sense to me than `bottom-start` and since this is the default behaviour, we can remove placement option, also in `shift` middleware, mainAxis is true by default so we don't need to specify that

client/components/Nav/NavDropdownMenu.jsx
PiyushChandra17: @lindapaiste Since `strategy: 'absolute` is default behaviour so we don't need this in options

PR #2720, opened 2023-12-14
fix: #2719 Shortcuts Ctrl + Shift + 1 and Ctrl + Shift + 2 aren't wor…

### LINKED ISSUE #2719: Bug: Shortcuts Ctrl + Shift + 1 and Ctrl + Shift + 2 aren't working.
### p5.js version

v1.9.0

Windows

### Web browser and version

120.0.6099.109 (Official Build) (64-bit) (cohort: M120 Rollout) 

### Actual Behavior

Shortcuts Ctrl + Shift + 1 and Ctrl + Shift + 2 should enable and disable the accessible output options, but they aren't doing anything.

### Expected Behavior

Shortcuts Ctrl + Shift + 1 and Ctrl + Shift + 2 should enable and disable the accessible output options.

### Steps:
1. Go to the web editor. 
2. Try the commands Ctrl + Shift + 1 to enable the accessible output options or disable them with Ctrl + Shift + 2 if they are already enabled.
3. Go to Accessibility in the settings and see that the accessibility options aren't changed
### Snippet:
```js

// Paste your code here :)

```

aryanas159:
…king.

Fixes #2719 
Changes:
Added a mapping for '!' -> 1, '@' -> 2, because due to the shift  key Ctrl + Shift + 1 was being registered as Ctrl + Shift + !, which wasn't triggering the event.

https://github.com/processing/p5.js-web-editor/assets/114330931/9fdfc241-1421-42d7-b61b-01e9253a57f6


lindapaiste:
Great catch on the problem! 🧤 

> The [`KeyboardEvent`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent) interface's [`key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) read-only property returns the value of the key pressed by the user, **taking into consideration the state of modifier keys such as Shift** as well as the keyboard locale and layout.

I wonder if there's a better way to handle this which doesn't involve manual mapping.  Like if there is a property other than `e.key` which we can look at.  It seems like maybe [`e.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) would work?

> The `KeyboardEvent.code` property represents a physical key on the keyboard (as opposed to the character generated by pressing the key). In other words, this property returns a value that isn't altered by keyboard layout or the state of the modifier keys

I also want to look into how CodeMirror handles their keymaps, and maybe we can copy something from their source code.

aryanas159:
Great idea @lindapaiste e.code works well, I have pushed the changes.
Also, I checked out the CodeMirror repo and it doesn't seem like they have a keymapping for ctrl+shift+{any number}.

aryanas159:
Hey, @lindapaiste could you please review the changes?

raclim (approved):
Thanks for your work on this!

PR #2727, opened 2023-12-15
Changed arrow color on ConsoleInput

### LINKED ISSUE #2721: Consoleinput arrow is not Adjusting to changed theme.
### p5.js version

_No response_

Windows

### Web browser and version

_No response_

### Actual Behavior

Arrow in consoleinput is not adjusted to Themes.
looking

### Expected Behavior

It should change colour according to theme.
shoullook

### Steps:
1.  Start the editor 
2. Change the theme

Typicalcodes:
Fixes #2721 

Changes:

* [x] is descriptively named and links to an issue number, i.e.


lindapaiste:
Looking at your changes I can see that we have an existing theme variable `console-active-arrow-color` specifically for this element.  But it was not being applied because we were only targeting `path` and this SVG uses `polygon`[1].  You fixed that 👍.  How does it look if we use your change in the targeting (`& g, & polygon, & path {`) but leave the color variable like it was before?  Better or worse?

[1] This sort of thing is a frequent problem throughout the app.  At some point in the future I would really like to overhaul how we recolor SVGs so that we only need to target the `<svg>` element, and the SVG file would use `currentColor` to pass down that color.

PiyushChandra17 (commented):
👍

Sudarshan-21 (approved):
It is working fine and looks great too, but dont you think changing color to something similar to the 'console-header-color' would be more impressive.

raclim (approved):
Thanks for your work on this, I think this is great for now!

PR #2405, opened 2023-09-10
changed EmailVerificationView Component to functional component

Akhilbisht798:
This pr is progress on issue #2358.

Changes:
- Changes EmailVerificationVeiw component to functional Component
- 
I have verified that this pull request:

* [X] has no linting errors (`npm run lint`)
* [X] has no test errors (`npm run test`)
* [X] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* [X] is descriptively named and links to an issue number, i.e. `Fixes #123`

Issue #2179, opened 2023-03-23
`mapStateToProps` and `mapDispatchToProps` often include unused props


lindapaiste:
Related to #824 and #2042

I've noticed that a lot of components are accessing props through the Redux `connect` HOC which are not actually used in the component itself.  The `mapStateToProps` and `mapDispatchToProps` don't get checked for unnecessary fields in the way that the `PropTypes` definitions do so it can be hard to spot them.

For example, the `IDEView` gets the `editorAccessibility` state and all `EditorAccessibilityActions` but does not use any of them.

The bulk importing of action creator objects with the `import * as EditorAccessibilityActions from '../actions/editorAccessibility';` syntax makes it extra difficult to see which of those actions, if any, are actually used.

This is one of many reasons that the [hooks are better](https://redux.js.org/style-guide/#use-the-react-redux-hooks-api), but we still have a lot of class component that can't use hooks yet.

FYI, the `mapDispatchToProps` setup with [`bindActionCreators`](https://redux.js.org/api/bindactioncreators) is not necessary because you can use the [object notation](https://react-redux.js.org/using-react-redux/connect-mapdispatch#defining-mapdispatchtoprops-as-an-object) to pass the action creators without binding.

PR #2736, opened 2023-12-17
fix: infinite api requests when the verification token is invalid

### LINKED ISSUE #2652: Expired verification-link result in infinity verification request-loop and (at the end) browser (or part of it) crash
### Most appropriate sections of the p5.js website?

User Verification process (Verify Email)

Mac OS

### Web browser and version

Mozilla Firefox Developer Edition 121.0b2

### Actual Behavior

I clicked the "Verify Email", and the page just mass looping through the verification process.

billede

billede

billede

billede

billede

### Expected Behavior

I expect the page to use the verification token once, and not spamming the p5.js backend.

If the token are invalid (or already used), tell me that it's invalid.

### Steps to reproduce

Open a link with a invalid token like  [https://editor.p5js.org/verify?t=letthefunbegin](https://editor.p5js.org/verify?t=letthefunbegin) (or a expired one)

billede

And let the fun begin ;-) ....

I have Consent-O-meter + uBlock Origin installed in my browser.

I ended up in the situation, as I accidentally closed the verification tab just after I clicked the "Verify Email" button in the email. So I reopened the tab, and let me just say that I should be verificed a MASSIVE amount of times, as this point 😉.

### Would you like to work on the issue?

Sadly, no. I dont have time to do so. But handling "undefined" data, and a expired link case, shouldn't take long 😃.

Also, it doenst really look like that's part of the repo:
billede

aryanas159:
Fixes #2652

Changes:
Because of the continuous prop changes the useEffect was executing infinitely, I have added a state that allows only one API call to check where the token in invalid or not.

https://github.com/processing/p5.js-web-editor/assets/114330931/b22e4dd2-3463-4d95-8d79-5d4c3ff98636


lindapaiste:
> Changes: Because of the continuous prop changes the useEffect was executing infinitely

Yikes!

> I have added a state that allows only one API call to check where the token in invalid or not.

This is definitely an improvement and should solve the problem.  It feels like a bit of a band-aid solution though.  You've identified that the core problem is having `props` in the `useEffect` dependencies, which causes the effect to run on every re-render.  We can change the dependencies to `[location, props.verifyEmailConfirmation]` so that it depends only on the one function and not on the `props` object.  I'm not sure if that works though because I'm not sure if `props.verifyEmailConfirmation` is a stable reference, especially since we're using `bindActionCreators` (which is [completely unnecessary](https://github.com/processing/p5.js-web-editor/issues/2179)).  Maybe it's time to clean up this component and bring it into the modern age.  See the checklist in https://github.com/processing/p5.js-web-editor/pull/2405#issuecomment-1718326111.  The following code will definitely only re-execute the effect if the token changes.

```

const location = useLocation();

const browserHistory = useHistory();

const dispatch = useDispatch();

const emailVerificationTokenState = useSelector(
  (state) => state.user.emailVerificationTokenState
);

const verificationToken = useMemo(() => {
  const searchParams = new URLSearchParams(location.search);
  return searchParams.get('t');
}, [location.search]);

useEffect(() => {
  if (verificationToken) {
    dispatch(verifyEmailConfirmation(verificationToken));
  }
}, [dispatch, verificationToken]);
```

aryanas159:
@lindapaiste Great suggestion, I have made the following changes you listed in https://github.com/processing/p5.js-web-editor/pull/2405#issuecomment-1718326111. It works well since the dependency array contains the verification token which is memoized there are no infinite loops.
Let me know if these kinds of changes are required for other components as well.

aryanas159:
@lindapaiste, could you please review the changes?

client/modules/User/pages/EmailVerificationView.jsx
```diff
@@ -1,32 +1,31 @@
...
+    if (verificationToken) {
+      dispatch(verifyEmailConfirmation(verificationToken));
     }
-  }, [location, props]);
-
+  }, [dispatch, verificationToken]);
   let status = null;
-
-  if (verificationTokenFromLocation() == null) {
+  if (verificationToken == null) {
```
lindapaiste:
```suggestion
  if (!verificationToken) {
```

Let's change this so that we get the "That link is invalid." message if the token is an empty string.  ie. `https://editor.p5js.org/verify?t=`

lindapaiste (changes requested):
Looking good.  Just one minor change.

lindapaiste:
I went ahead and committed my suggestion because I want to get this bug fixed.

As a side note, it's weird that the message that we show when the token is `invalid` is `"Something went wrong."` and not the `"Token is invalid or has expired."` message that we get back from the API.  But that's something to address in a separate issue.

lindapaiste (approved):
It's so much cleaner now, thank you!

PR #2763, opened 2023-12-23
added preview for sketches

### LINKED ISSUE #2532: Adding examples preview
### Increasing Access

I think as a user when you scroll through the examples section, the User is generally confused by such a number of examples and cannot determine the actual behaviour of each shape; in other words, it is not very visually appealing; if the user is unsure of any shape he will have to check each sketch one by one, And it will take time as it will always redirect to the code editor, I am sure that would not be a very pleasant experience as a user. There should be a preview button. There is already an existing dropdown button which contains the download button. But it can be updated with a [+] expand view. It will display the preview, increase user accessibility, and make it easier to use.

### Feature request details

In this feature, We will update the down arrow section, as shown in the screenshot. It will then expand the div where the preview of the sketch will be running, displaying the animation of the shape the div will also contain three buttons on the left side or right side,
1) share button (to share the shape)
2)Copy(To copy the default code)
3)Download
Making this feature will save users time, and it can be implemented, and the code would be running on click, so it will be convenient too.
Screenshot 2023-10-24 165035

KrishavRajSingh:
Fixes #2532 

I used <EmbedFrame /> the method that was used in #2168  and succeeded in making the previews, but it wasn't working for some sketches, especially the ones with images.

Changes:
1> Added a MenuItem for Preview in Sketch's Dropdown list.
2> Created a ExamplePreview Component to view the Preview. Used EmbedFrame to preview the sketches but it is not working for all the sketches. **Need help in this.**


lindapaiste:
Thanks for working on this!  I’ll take a look after the holidays and see if I can figure out the problems.

client/modules/IDE/components/SketchList.jsx
```diff
@@ -35,7 +36,8 @@ class SketchListRowBase extends React.Component {
     super(props);
     this.state = {
       renameOpen: false,
-      renameValue: props.sketch.name
+      renameValue: props.sketch.name,
+      preview: false
```
lindapaiste, line 40: 👍

client/modules/IDE/components/SketchList.jsx
```diff
@@ -144,6 +146,22 @@ class SketchListRowBase extends React.Component {
           >
             {this.props.t('SketchList.DropdownAddToCollection')}
           </MenuItem>
+          <MenuItem
+            hideIf={!this.props.user.authenticated}
```
lindapaiste: We don't need any `hideIf`.  Anyone can open the preview.

client/modules/IDE/components/SketchList.jsx
```diff
@@ -194,6 +212,16 @@ class SketchListRowBase extends React.Component {
...
         </tr>
+        {this.state.preview ? (
+          <ExamplePreview
+            url={`/editor/${this.props.username}/projects/${slugify(
+              this.props.sketch.name,
+              '_'
+            )}`}
+          />
+        ) : (
+          <div>loading</div>
```
lindapaiste: If `this.state.preview` is `false` then we would show nothing.  Not a loader.  You can do `{this.state.preview && (` or keep the ternary and use `) : null}` for the `false` case.

client/modules/User/components/ExamplePreview.jsx
```diff
@@ -0,0 +1,83 @@
...
+      // Find example projects
+      const response = await axios.get(url);
+      const projects = response.data;
+      if (!projects) return null;
+      setSketch(projects);
+      console.log('sketch', sketch);
+      return projects;
+    }
+    getRandomProject();
+  }, [url]);
```
lindapaiste: In the 404 page this part of the code gets all sketches from the p5 user and returns a random one.  Here we want to get the one specific sketch by its id.
https://github.com/processing/p5.js-web-editor/blob/41529a3bb401e6516ef2d62d9fca9ba8aa21bf43/client/modules/IDE/actions/project.js#L57-L58

client/modules/User/components/ExamplePreview.jsx
```diff
@@ -0,0 +1,83 @@
...
+      content: file.content
+        // Fix links to assets
+        .replace(
+          /'assets/g,
+          "'https://github.com/processing/p5.js-website/raw/main/src/data/examples/assets"
+        )
+        .replace(
+          /"assets/g,
+          '"https://github.com/processing/p5.js-website/raw/main/src/data/examples/assets'
+        )
```
lindapaiste: hmmm...this will only work for sketches on the p5 user.  But the code in `SketchList` adds this functionality to all users' sketches.  So we would need to get the files from s3.  We have code to do that in other places but it's really complex.  I think it's already handled inside the `EmbedFrame` though?

client/modules/User/components/ExamplePreview.jsx
```diff
@@ -0,0 +1,83 @@
...
+        // Change canvas size
+        .replace(
+          /createCanvas\(\d+, ?\d+/g,
+          instanceMode
+            ? 'createCanvas(p.windowWidth, p.windowHeight'
+            : 'createCanvas(windowWidth, windowHeight'
```
lindapaiste: I'm not sure what size we want?

client/modules/User/components/ExamplePreview.jsx
```diff
@@ -0,0 +1,83 @@
...
+  }, [sketch]);
+  console.log(files, 'files');
+  return (
+    <React.Fragment>
+      {files.length > 0 ? (
+        <div style={{ position: 'relative', width: '100%', height: '10rem' }}>
+          <GlobalStyle />
+          <EmbedFrame
+            files={files}
+            basePath={`${window.location.pathname}/3D:_basic_shader`}
```
lindapaiste: I have to refresh my memory on what the `basePath` is supposed to be.

client/modules/User/components/ExamplePreview.jsx
```diff
@@ -0,0 +1,83 @@
...
+            textOutput={false}
+            gridOutput={false}
+            isPlaying
+          />
+        </div>
+      ) : (
+        <div>loading</div>
```
lindapaiste, line 56: Loading is fine here ✔️

client/modules/IDE/components/SketchList.jsx
```diff
@@ -194,6 +212,16 @@ class SketchListRowBase extends React.Component {
           <td>{formatDateCell(sketch.updatedAt, mobile)}</td>
           {this.renderDropdown()}
         </tr>
+        {this.state.preview ? (
+          <ExamplePreview
+            url={`/editor/${this.props.username}/projects/${slugify(
+              this.props.sketch.name,
+              '_'
+            )}`}
```
lindapaiste: Only the p5 user has slugified project URLs.  URLs for other users are based on the id.  But I don't think the `url` is what we want to pass down to the `ExamplePreview` anyways.  If we have the entire `sketch` object (I think we do?) then we should pass that down, and then we don't need any fetching in the `ExamplePreview`.  If it's just a summary of the sketch then we would pass down the `id` and `username` of the sketch and then fetch the complete details in the `ExamplePreview`.

lindapaiste (changes requested):

lindapaiste:
I pushed some changes to your branch.   It seems like the sketch preview works now but we probably need to think through the UI a bit, like we probably want an "x" button over the preview or some other easy way to close.

KrishavRajSingh:
> I pushed some changes to your branch. It seems like the sketch preview works now but we probably need to think through the UI a bit, like we probably want an "x" button over the preview or some other easy way to close.

Yeah Sure, I will make the change

raclim:

Since it's been a while since this PR was submitted and I think that the UI for this issue might need more thought, I'm going to close this for now. I'm really sorry that we couldn't get this in, but please feel free to revisit this PR when further UI updates are fleshed out or check out our other issues!

PR #2767, opened 2023-12-24
Update user.controller.js, Propagating the error via the call back

koushik-dasss:
Changes: 
 The function either should reject the error or the error should be logged or the original function must be told that there was an error , but it doesn't work like that . 
 This code  fixes by propagating the error via the call back . So , no ignorance to the error is made .


lindapaiste:
We use error-first callbacks in a lot of places.  For whatever reason this callback has been passing the user argument as the first argument.  If you change it here then you have to change a bunch of other places too, because the existing code that uses this function is expecting the user object to be the first argument in the callback function.  So you need to change those callbacks.

PR #2768, opened 2023-12-24
Update user.controller.js Fixing the error of ignoring the ERROR case…

koushik-dasss:
Using return promise logic of JS to get the ERROR CASES fixed without ignoring like the current code does .
Fixes  IGNORANCE OF ERROR CASES by adding return promise method .

Screenshot 2023-12-24 112207

### DIFF
--- a/server/controllers/user.controller.js
+++ b/server/controllers/user.controller.js
@@ -29,11 +29,18 @@ const random = (done) => {
   });
 };
 
-export function findUserByUsername(username, cb) {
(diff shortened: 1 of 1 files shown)


lindapaiste:
Copying my comment from #2772 because this PR has the same issue.

The changes to the function look good. You’ve changed the signature of the function so that it no longer accepts a callback as an argument. So you need to make sure that you have updated all the places where this function is used so that the work with an async/promise call instead of passing a callback.

PR #2783, opened 2023-12-26
Fixed #2647, fixed overlay button blue border-box!

### LINKED ISSUE #2647: Position of `cross` in about modal is not aligned `center` in blue border box
### p5.js version

2.9.3

None

### Web browser and version

_No response_

### Actual Behavior

Position of `cross` in about modal is not aligned `center` in blue border box

about-x

### Expected Behavior

Position of `cross` in about modal should be aligned `center` in blue border box

### Steps:
1. Open the editor web app
2. Open the about modal
3. Click on Close
4. Notice it's not aligned center wrt box

vivekbopaliya:
Fixes #2647, Position of cross in about modal is now aligned center in blue border box.

Before: 
Screenshot (174)

Changes:
Screenshot (175)

- [x] has no linting errors (`npm run lint`)
- [x] has no test errors (`npm run test`)
- [x] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
- [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

### DIFF
--- a/client/modules/IDE/components/AddToCollectionList.jsx
+++ b/client/modules/IDE/components/AddToCollectionList.jsx
@@ -39,7 +39,7 @@ class CollectionList extends React.Component {
       props.getProject(props.projectId);
     }
 
-    this.props.getCollections(this.props.username);
+    this.props.getCollections(this.props.user.username);
 
     this.state = {
       hasLoadedData: false
--- a/client/modules/IDE/components/AddToCollectionList.jsx
+++ b/client/modules/IDE/components/AddToCollectionList.jsx
@@ -39,7 +39,7 @@ class CollectionList extends React.Component {
       props.getProject(props.projectId);
     }
 
-    this.props.getCollections(this.props.user.username);
+    this.props.getCollections(this.props.username);
 
     this.state = {
       hasLoadedData: false
--- a/client/styles/components/_overlay.scss
+++ b/client/styles/components/_overlay.scss
@@ -50,10 +50,17 @@
 }
 
 .overlay__close-button {
-  @include icon();
-  padding: #{3 / $base-font-size}rem 0 #{3 / $base-font-size}rem #{10 / $base-font-size}rem;
+  display: flex;
+  align-items: center;
+  justify-content: center;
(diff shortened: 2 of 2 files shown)


lindapaiste (commented):
It looks good, thanks for the screenshots.

I'm playing around with the CSS and it seems like the code which actually is required for the fix is only the `display: flex` and the `padding`.  The rest doesn't make a difference (at least for me in Windows Chrome).  The button is only the size of the "x" + padding, so the `align-items` and `justify-content` don't do anything as there is no extra space to distribute.  The SVG is already `16px` based on the sizes in the `exit.svg` file (though IMO it generally makes more sense to size SVGs using CSS).  It doesn't need `cursor: pointer` because that is the default cursor for all `<button>` HTML elements.

These changes make the "x" further from the edge and also shift the content a little bit.  The "About" is no longer vertically aligned with the "x" but that's easily fixed with an `align-items: center;` on the parent `.overlay__header` element.  The whole header is now a bit taller and the content is shifted down, so I suspect that we want some amount of negative margin?  That allow us to have a larger clickable target and a square and centered box around the "x" without moving the layout.  See [comment](https://github.com/processing/p5.js-web-editor/issues/2647#issuecomment-1830170963).

I'd love for @raclim to chime in on what exactly we want the final result to look like.  Should the "x" stay in the same exact position that it is now?  Should the spacing on the top be equal to the spacing on the side?  How much spacing do we want?

vivekbopaliya:
OH YES GOOD CALL! there always have to be multiple solutions to this as i didn't try many things with` .overlay_header`, but around time, when i was trying to center the blue box, it ended up having more space on the left. So, I added a bit of padding around the button and justify-item center.

Your solution looks cleaner than mine. Go ahead with it.

Thanks for the pictures.

raclim:
Thank you for your work on this! 

Due to recent updates, I'm ultimately going to close this due to the reasons laid out in https://github.com/processing/p5.js-web-editor/issues/2647#issuecomment-2004694237. I'm really sorry we couldn't get this in, but please be on the lookout for the new designs!

PR #2799, opened 2023-12-28
Use styled-components in CollectionListRow

adadgarg:
Progress on #1760 

Changes:

- Migrate styles from client/styles/components/_sketch_list.scss into individual styled components in the client/modules/IDE/components/CollectionList/CollectionListRow.jsx file.
- Looks identical to before on Windows Chrome.


client/theme.js
```diff
@@ -248,6 +254,9 @@ export default {
   ...baseThemes,
   [Theme.contrast]: extend(baseThemes[Theme.dark], {
     inactiveTextColor: grays.light,
+    modalBorderColor: grays.middleDark,
+    searchBackgroundColor: grays.darker,
+    tableRowStripeColor: grays.dark,
```
lindapaiste: This isn't necessary due to the `extend(baseThemes[Theme.dark]`.  We put that in as a helped to avoid having to repeat values which are the same as the dark theme.

lindapaiste:
This looks good at first glance.  I wonder how it works alongside some of the pending PRs that I have re: common table components.  I noticed that you didn't delete the classes from the `.scss` file.  Possibly because they are used in multiple places?  We have the same styles on multiple tables so possibly the conversion to styled-components becomes part of creating a common table component/system, where we have a `TableBase` with the core styles that can be extended or overwritten for specific tables.

client/theme.js
adadgarg: Deleted redundant section.Thanks for pointing it out!

adityagarg06:
> This looks good at first glance. I wonder how it works alongside some of the pending PRs that I have re: common table components. I noticed that you didn't delete the classes from the `.scss` file. Possibly because they are used in multiple places? We have the same styles on multiple tables so possibly the conversion to styled-components becomes part of creating a common table component/system, where we have a `TableBase` with the core styles that can be extended or overwritten for specific tables.

Yeah, I didn't remove the classes as it was used in other components. I agree with the idea of creating a common table styles component and extending it for all the other tables. This same can be done for the `Form` components.

adityagarg06:
@lindapaiste I have created a primary Table core styles component. It is mainly used in SketchList, Collection, CollectionList, CollectionCreate, and CollectionListRow components. Should I create a PR, and in which directory should I add this file?

raclim (approved):

PR #2811, opened 2023-12-30
Fixed checkbox color (issue-#2808)

### LINKED ISSUE #2808: Color of checkbox should be of p5 editor's theme but its not
### p5.js version

_No response_

Windows

### Web browser and version

Chrome

### Actual Behavior

Screenshot 2023-12-30 183221

The color of checkbox should be of p5 editor's theme according to this code but this code is not working and the color of checkbox is default

### Expected Behavior

https://github.com/processing/p5.js-web-editor/assets/112943430/5258643e-5a13-45b7-aacf-cfef8a5cc23e

The expected behaviour should be like this...

### Steps:
1. Go to the p5 web editor website
2. In tool section below the header, there is a checkbox

PiyushThapaa:
Fixes #2808 

Changes:
I corrected the code in toolbar.scss file line 147

* [x] is descriptively named and links to an issue number, i.e. `Fixes #2808 `

### DIFF
--- a/client/styles/components/_toolbar.scss
+++ b/client/styles/components/_toolbar.scss
@@ -144,6 +144,6 @@
 .checkbox__autorefresh{
 	cursor: pointer;
 	@include themify(){
-		color:getThemifyVariable('logo-color');
+		accent-color:getThemifyVariable('logo-color');
 	}
 }


client/styles/components/_toolbar.scss
```diff
@@ -144,6 +144,6 @@
 .checkbox__autorefresh{
 	cursor: pointer;
 	@include themify(){
-		color:getThemifyVariable('logo-color');
+		accent-color:getThemifyVariable('logo-color');
```
lindapaiste, line 147: Looks good, and it seems like this property has pretty good support: https://caniuse.com/mdn-css_properties_accent-color  It's definitely working for me!

I'm wondering if we want some sort of fallback for older browsers which don't support `accent-color`.  Like will the `color` setting work there or will it do nothing?  I'm reading some articles and it seems like it would do nothing but I'm not sure.  Do you know?  (it's ok if you don't)

PiyushThapaa: @lindapaiste See, the color property is not working anywhere because it is not the property to change the color of a input checkbox ( as far as i know ). May be accent color will not work in some older browsers but color property will not work anywhere because it is not the property to change the color of a input checkbox. So its better to use accent color.
I run this code locally and in my browsers it works perfectly file... You can see that in the issue #2808 in expected behaviour.

lindapaiste: Ok yeah I was wondering if there was an easy way to get backwards compatibility but it seems like there is not.  So this looks good.

lindapaiste (approved):

PR #2820, opened 2024-01-01
Refactor: Resolve repeated code between server's previewGeneration.js and client's EmbedFrame.jsx

### LINKED ISSUE #2805: Repeated code between server `previewGeneration.js` and client `EmbedFrame.jsx`
### Increasing Access

Tasks:
- [ ] Add unit tests.

Questions:

Shyam-Raghuwanshi:
Fixes #2805

Changes:


raclim:
Thanks for your work on this! 

I think that although what's here so far does resolve the issue of repeated code between the two files, this doesn't address some of the other tasks listed in the issue (i.e "Add explanatory comments in the code, Add unit tests, Make improvements to the code itself...") that would help implement a more thorough, long-term solution. 

Due to the number of PRs we have at the moment, I'm going to close this one for now. I'm sorry that we couldn't get this in yet, but please feel free to reopen this PR and update it with the needed changes, thanks!

PR #2822, opened 2024-01-01
Fix Button with `display="inline"` to use the same styles as the regular button

### LINKED ISSUE #2821: `Button` component has a different set of styles with prop `display="inline"`
### Increasing Access

We want to have a useful set of basic UI components that we can use throughout the app.  They should be customizable via props but the output should be predictable.

We are not using the inline version anywhere because it is not useful in it's current state.  If we fix it up then it can be used for issues like #2734.

Our `Button` UI component has a prop `display` which supports values `"block"` and `"inline"`.  I would expect that a `<Button/>` with `display="inline"` would look the same as the normal button, with the only difference being that it is on the same line instead of a new line.  In reality it uses a completely different styled component.

https://github.com/processing/p5.js-web-editor/blob/e77bc6ac3e5fd499a12821fef8ecac91a172fa3a/client/common/Button.jsx#L185-L189

https://github.com/processing/p5.js-web-editor/blob/e77bc6ac3e5fd499a12821fef8ecac91a172fa3a/client/common/Button.jsx#L83-L108

The inline version is always the same color regardless of the `kind="primary"` and `kind="secondary"` props, as the color is hard-coded.

This is what I get when using `inline` buttons:
Screenshot 2024-01-01 14 44 10

The wildest part is that it's not even inline!  It has `display: flex` instead of `display: inline-flex`.
https://github.com/processing/p5.js-web-editor/blob/e77bc6ac3e5fd499a12821fef8ecac91a172fa3a/client/common/Button.jsx#L85

I would expect something more like this, which is the `block` style button but displayed inline:
Screenshot 2024-01-01 14 42 53

My proposal is that we should delete the `StyledInlineButton` styled component.  Instead, we should adjust the existing `StyledButton` to change its `display` property based on the `display` prop.  This ensures that all other styling is consistent regardless of the `display` prop.

lindapaiste:
Fixes #2821 

This does not change any UI anywhere in the app.  It removes some unused code and enables us to use the `<Button>` component inline in the future.  There is more that could be refactored and cleaned up but I'm keeping this PR minimal.

Changes:
- Modify the core `StyledButton` component to use either `display: inline-flex` or `display: flex`, depending on the value of the `display` prop.
- Deleted the `StyledIconButton` component which was not used anywhere.  Confusingly, the existing `IconButton` component which we use on mobile is an extension of the `StyledInlineButton` rather than the `StyledIconButton`.
- Maintain the styles of the current `IconButton` component by passing down the prop `iconOnly` and using this prop to switch components rather than switching based on the `display` prop.


PiyushChandra17 (approved):
Nicely Done 👍

raclim (approved):
I think this looks good to me! Thanks for working on this!

PR #2824, opened 2024-01-02
#issue-2823 changed the button colors on hovering

paulpriyanshu:
Fixes #issue-2823


lindapaiste:
This only addresses the "light" theme and not the other two ("dark" and "contrast").  I'm going to close this because we have other PRs that are closer to correct.

Issue #2846, opened 2024-01-06
The t&c portion of the SingUp page is not translated


harmanbatheja15:
### What is your operating system?

Mac OS

### Actual Behavior

The T&C is not being translated

<img width="748" alt="Screenshot 2024-01-06 at 6 44 30 AM" src="https://github.com/processing/p5.js-web-editor/assets/66913564/f00223d5-429a-48bc-bf56-5cc3b93ff481">

### Expected Behavior

The T&C should be translated.

<img width="610" alt="Screenshot 2024-01-06 at 6 45 04 AM" src="https://github.com/processing/p5.js-web-editor/assets/66913564/927dc550-0d5c-483b-b6d1-e10c7c615ecd">

### Steps:
1. Add ```"Warning": "By signing up, you agree to the p5.js Editor's <0>Terms of Use</0> and <1>Privacy Policy.</1>"``` in [_translations/locales/en-US/translations.json_](/processing/p5.js-web-editor/blob/develop/translations/locales/en-US/translations.json) file, inside the _SignupView_ section.
2. Update the import statement in the [_SignupView.jsx_](/processing/p5.js-web-editor/blob/develop/client/modules/User/pages/SignupView.jsx) file.
3. Add the updated Translation code in the [_SignupView.jsx_](/processing/p5.js-web-editor/blob/develop/client/modules/User/pages/SignupView.jsx) file.

### Snippet:

```js

"SignupView": {
    "Title": "p5.js Web Editor | Signup",
    "Description": "Sign Up",
    "Or": "Or",
    "AlreadyHave": "Already have an account?",
    "Login": "Log In",
    "Warning": "By signing up, you agree to the p5.js Editor's <0>Terms of Use</0> and <1>Privacy Policy.</1>"
  },

```

```js

import { useTranslation, Trans } from 'react-i18next';

```

```js

<Trans
    i18nKey="SignupView.Warning"
    components={[
        <Link to="/terms-of-use">Terms of use</Link>,
        <Link to="/privacy-policy">Privacy Policy</Link>
    ]}
/>

```

lindapaiste:
fixed by #2847

PR #2847, opened 2024-01-06
Fixed: Translation of T&C on SignUp page

harmanbatheja15:
Fixes issue #2846 
Fixed: Translation of T&C on the SignUp page.


client/modules/User/pages/SignupView.jsx
```diff
@@ -27,9 +27,13 @@ function SignupView() {
             <SocialAuthButton service={SocialAuthButton.services.google} />
           </div>
           <p className="form__navigation-options">
-            By signing up, you agree to the p5.js Editor&apos;s{' '}
-            <Link to="/terms-of-use">Terms of Use</Link> and{' '}
-            <Link to="/privacy-policy">Privacy Policy</Link>.
+            <Trans
+              i18nKey="SignupView.Warning"
+              components={[
+                <Link to="/terms-of-use">Terms of use</Link>,
```
lindapaiste, line 33: Thanks for catching this!

We need to look over the i18next docs to make sure that we've got the right syntax for the `components`.  Having the phrase "Terms of use" here in English makes me nervous that the anchor text of the link won't get translated.  If you want to add the Hindi translations to this PR that would be helpful because the we would have a language that we can switch to in order to double-check the behavior.

lindapaiste: Okay I think that you have it right and the anchor text of the `<Link>` doesn't matter.  Weird.  https://react.i18next.com/latest/trans-component#alternative-usage-components-array

harmanbatheja15: Added Hindi Translation also.

lindapaiste (approved):
Well now I'm realizing that maybe I shouldn't have had you include the Hindi in the same PR because I can't verify that the Hindi text is correct 🙃.  But I'm going to trust you on it and if anyone wants an alternate translation then they can open a PR to change it.

I ran your branch and switched the language to Hindi and I can see that the anchor texts of both links are getting translated so that's great ✔️.

PR #2875, opened 2024-01-09
fix signup issue

sourav-sm:
Fixes ##2871
Initilaly it was --
Screenshot 2024-01-08 204742

After fixing the issue--
Screenshot 2024-01-09 222102

Changes:
I have changes the signup issue which is not properly translated earlier,But now it fixed


translations/locales/en-US/translations.json
```diff
@@ -365,7 +365,7 @@
     "Or": "Or",
     "AlreadyHave": "Already have an account?",
     "Login": "Log In",
-    "Warning": "By signing up, you agree to the p5.js Editor's <0>Terms of Use</0> and <1>Privacy Policy.</1>"
+    "Warning": " you agree to the p5.js Editor's <0>Terms of Use</0> and <1>Privacy Policy.</1>"
```
raclim: "By signing up," should be added back in here!

raclim (changes requested):

translations/locales/en-US/translations.json
sourav-sm: oho yeah my bad ,i missed it, should i add it ?

raclim: Yes, ideally there should be no changes to this file!

sourav-sm: Hello raclim i fixed the above error .
image
image

raclim: Thanks, looks good! I also added in a minor change that updates "Signing" to be lowercase instead of capitalized.

raclim (approved):
Thanks for your contribution!

sourav-sm:
It is my pleasure to fix it as my 1st open source contribution

PR #2883, opened 2024-01-12
Feature/enhancement of hoverinig buttons

### LINKED ISSUE #2823: Solid pink buttons need a hover effect
### Increasing Access

All buttons should have a visible change when hovered so that the user knows that it's hovered.

Gray buttons go to pink when hovered but pink buttons stay pink.  We should make it a slightly darker shade of pink to show that it is hovered.

The solid pink button is the `secondary` `kind` of the [`Button`](https://github.com/processing/p5.js-web-editor/blob/e77bc6ac3e5fd499a12821fef8ecac91a172fa3a/client/common/Button.jsx) component.  It is used in the `CookieConsent` dialog and potentially in other places in the future.

In the `theme.js` file we would change the `hover` colors for the `secondary` button style to be something different.  Right now we have the same colors for `default` and `hover` but there's a `colors.p5jsActivePink` that we can use.  We don't seem to have an equivalent variation of the yellow color for the `contrast` theme so we would need to define one.

https://github.com/processing/p5.js-web-editor/blob/e77bc6ac3e5fd499a12821fef8ecac91a172fa3a/client/theme.js#L102-L112

This issue is intended for @haarsh157 but I can't assign it unless they comment 😝

haarsh157:
Fixes #2823, #2905

Changes:
1. Changed hovering color to `p5jsActivePink`.
2. Added new yellow color `p5ContrastYellow` and changed the button hovering color in the contrast theme.

* [x] is descriptively named and links to an issue number, i.e. `Fixes #2823 `


raclim (approved):
Thanks for your work on this! 

I think overall this looks good! I made a few adjustments where I removed the changes to the `RootPage.jsx` file and the change to the `foreground` property in the `theme.js` file. I also feel that the `p5jsActivePink` color is not a noticeable enough difference to identify the hover state, but think that could be reconsidered down the line!

Issue #1094, opened 2019-06-05
Privacy and exposed API keys


jeremydouglass:
#### Nature of issue?

- New feature request

#### New feature details:

Currently, people who create sketches on the web editor using API keys are automatically exposing those keys to the world, and may not be aware they are doing so. 

Right now searching the web editor for "apiKey" or similar strings
https://www.google.com/search?q="apiKey"+site%3Aeditor.p5js.org

...turns up a couple hundred examples from firebase, google maps, the new york times, giphy, mlab.com, pixabay, uifaces.co, wordnik, openweathermap.org, etc. Most of these are low security, but some are not.

Other than a robust private sketch feature, here are some ways of approaching this problem with features:

1. post a warning for the author if the editor recognizes an api string and api key. The author can ignore the warning or not.
2. advise the user to use save their sketch with the key variables empty / included. They must be re-pasted to 
3. advise the user to save their sketch with loading the key from localstorage. This is the workaround that codepen suggests: https://blog.codepen.io/2017/08/10/window-prompt-localstorage-api-keys/
4. offer an option for users to create private strings, AKA secrets. This is what glitch.com implements -- for example https://glitch.com/edit/#!/processing-js?path=.env:1:0

catarak:
thanks for reporting! given that all apps on the web editor are front end only, this is tricky to solve, and also means that (4) isn't really an option. i like (3) as a solution, but the drawback is that sketches wouldn't be able to be run by users that don't own the sketch. i'm not sure how the editor would implement (1) or (2)—it seems like there could be a lot of edge cases, but maybe that's okay. 

i also agree this is another reason to add the ability to make sketches private!

ashu8912:
For Now We can add the feature of making sketch private.@catarak i can work on this one

PR #2904, opened 2024-01-15
Removed Edit Ability for Non-owned sketches 

### LINKED ISSUE #870: Feedback for inability to use Add Folder / File Menu on non-owned shetches
#### Nature of issue?

- Web browser and version:  FF 65.0.1 (64-bit)
- Operating System:  OSX 10.13.4
- Steps to reproduce this bug: Open a sketch that is not owned by your account and attempt to add  a file in the sketch menu

When looking at a sketch that is not your own (under you logged in account) you cannot add a file / folder. This is expected behavior, but I feel that there should be some feedback there to let you know that you're attempting to work on a sketch that is not your own. Especially because you can edit the code and have it run even if it's not technically 'your' sketch yet. 

Either blocking out those menu items or giving a dialogue that they need to duplicate the sketch first might be options?

vivekbopaliya:
Fixes #870 

Changes:

https://github.com/processing/p5.js-web-editor/assets/122638553/e2a4d7b3-512f-45e7-84da-e9d1bd452866

### Mobile view:
Screenshot (232)

I was planning to introduce a 'Remix' function as it was talked about [here ](https://github.com/processing/p5.js-web-editor/issues/870#issuecomment-776913417)then I realized that there's already a feature similar to 'Clone' (_duplicate sketch_) regardless you can still edit sketches that aren't yours (you can't add files/folders, though). To make it easier for users that help them distinguish between owned and non-owned sketches, we could make it **necessary** to **duplicate** sketches before making changes if they don't own them.

### DIFF
--- a/client/modules/IDE/components/Editor/index.jsx
+++ b/client/modules/IDE/components/Editor/index.jsx
@@ -423,17 +423,6 @@ class Editor extends React.Component {
       closeOnUnfocus: false
     };
 
-    const getUserAndProject = () => {
-      const { user, project } = useSelector((state) => ({
-        user: state.user,
-        project: state.project
-      }));
-
-      return { user, project };
-    };
-
-    getUserAndProject();
-
     if (_cm.options.mode === 'javascript') {
       // JavaScript
       CodeMirror.showHint(
--- a/client/modules/IDE/components/Editor/index.jsx
+++ b/client/modules/IDE/components/Editor/index.jsx
@@ -508,7 +508,11 @@ class Editor extends React.Component {
         this.props.file.fileType === 'folder' || this.props.file.url,
       // eslint-disable-next-line no-dupe-keys
       'editor-holder--readonly':
-        this.props.project.owner?.username !== this.props.user.username
+        //  Check if there is a project owner, the user has a username,
+        // and the project owner's username is not the same as the user's username
+        this.props.project.owner && this.props.user.username
+          ? this.props.project.owner?.username !== this.props.user.username
+          : ''
     });
 
     return (
@@ -571,8 +575,10 @@ class Editor extends React.Component {
               <section>
                 <EditorHolder
                   readOnly={
-                    this.props.project.owner?.username !==
-                    this.props.user.username
+                    this.props.project.owner &&
+                    this.props.user.username &&
+                    this.props.project.owner.username !==
+                      this.props.user.username
                   }
                   ref={(element) => {
                     this.codemirrorContainer = element;
--- a/client/modules/IDE/components/Editor/index.jsx
+++ b/client/modules/IDE/components/Editor/index.jsx
@@ -41,7 +41,7 @@ import { CSSLint } from 'csslint';
 import { HTMLHint } from 'htmlhint';
 import classNames from 'classnames';
 import { debounce } from 'lodash';
-import { connect } from 'react-redux';
+import { connect, useSelector } from 'react-redux';
 import { bindActionCreators } from 'redux';
 import MediaQuery from 'react-responsive';
 import '../../../../utils/htmlmixed';
@@ -423,6 +423,17 @@ class Editor extends React.Component {
       closeOnUnfocus: false
     };
 
+    const getUserAndProject = () => {
+      const { user, project } = useSelector((state) => ({
+        user: state.user,
+        project: state.project
+      }));
+
+      return { user, project };
+    };
+
+    getUserAndProject();
+
     if (_cm.options.mode === 'javascript') {
       // JavaScript
       CodeMirror.showHint(
@@ -505,7 +516,10 @@ class Editor extends React.Component {
     const editorHolderClass = classNames({
       'editor-holder': true,
       'editor-holder--hidden':
-        this.props.file.fileType === 'folder' || this.props.file.url
+        this.props.file.fileType === 'folder' || this.props.file.url,
+      // eslint-disable-next-line no-dupe-keys
+      'editor-holder--readonly':
+        this.props.project.owner?.username !== this.props.user.username
     });
 
     return (
@@ -567,6 +581,10 @@ class Editor extends React.Component {
               </header>
               <section>
                 <EditorHolder
+                  readOnly={
+                    this.props.project.owner?.username !==
+                    this.props.user.username
+                  }
                   ref={(element) => {
                     this.codemirrorContainer = element;
                   }}
@@ -588,6 +606,10 @@ class Editor extends React.Component {
 }
 
 Editor.propTypes = {
+  // eslint-disable-next-line react/forbid-prop-types
+  user: PropTypes.object.isRequired,
+  // eslint-disable-next-line react/forbid-prop-types
+  project: PropTypes.object.isRequired,
   autocloseBracketsQuotes: PropTypes.bool.isRequired,
   autocompleteHinter: PropTypes.bool.isRequired,
   lineNumbers: PropTypes.bool.isRequired,
--- a/client/modules/IDE/components/Header/MobileNav.jsx
+++ b/client/modules/IDE/components/Header/MobileNav.jsx
@@ -263,7 +263,7 @@ const MobileNav = () => {
                 }`}
                 onClick={() => dispatch(cloneProject())}
               >
-                Duplicate to edit{' '}
+                {t('Toolbar.DuplicateToEdit')}{' '}
                 <EditIcon
                   className={`toolbar__icon ${
                     theme === 'contrast' ? 'contrast' : 'normal'
--- a/client/modules/IDE/components/Header/MobileNav.jsx
+++ b/client/modules/IDE/components/Header/MobileNav.jsx
@@ -35,6 +35,8 @@ import { setLanguage } from '../../actions/preferences';
 import Overlay from '../../../App/components/Overlay';
 import ProjectName from './ProjectName';
 import CollectionCreate from '../../../User/components/CollectionCreate';
+import { cloneProject } from '../../actions/project';
+import EditIcon from '../../../../images/pencil.svg';
 
 const Nav = styled(NavBar)`
   background: ${prop('MobilePanel.default.background')};
@@ -67,15 +69,20 @@ const LogoContainer = styled.div`
 
 const Title = styled.div`
   display: flex;
-  flex-direction: column;
-  gap: ${remSize(2)};
+  gap: ${remSize(10)};
 
   * {
     padding: 0;
     margin: 0;
   }
 
-  > h5 {
+  p {
+    margin-left: 2px;
+    padding: 3px 8px;
+  }
+
+  h5 {
+    margin-top: 2px;
     font-size: ${remSize(13)};
     font-weight: normal;
   }
@@ -205,6 +212,8 @@ const MobileNav = () => {
   const user = useSelector((state) => state.user);
 
   const { t } = useTranslation();
+  const theme = useSelector((state) => state.preferences.theme);
+  console.log(theme);
 
   const editorLink = useSelector(selectSketchPath);
   const pageName = useWhatPage();
@@ -228,7 +237,7 @@ const MobileNav = () => {
   }
 
   const title = useMemo(resolveTitle, [pageName, project.name]);
-
+  const dispatch = useDispatch();
   const Logo = AsteriskIcon;
   return (
     <Nav>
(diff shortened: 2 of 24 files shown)


client/modules/IDE/components/Editor/index.jsx
```diff
@@ -505,7 +505,14 @@ class Editor extends React.Component {
     const editorHolderClass = classNames({
       'editor-holder': true,
       'editor-holder--hidden':
-        this.props.file.fileType === 'folder' || this.props.file.url
+        this.props.file.fileType === 'folder' || this.props.file.url,
+      // eslint-disable-next-line no-dupe-keys
```
lindapaiste, line 509: I don't think there's any dupe keys?  You can probably delete this `disable`.

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -505,7 +505,14 @@ class Editor extends React.Component {
...
+      'editor-holder--readonly':
+        //  Check if there is a project owner, the user has a username,
+        // and the project owner's username is not the same as the user's username
+        this.props.project.owner && this.props.user.username
+          ? this.props.project.owner?.username !== this.props.user.username
+          : ''
```
lindapaiste, line 515: We'll want to store the `isReadOnly` (or `canEdit`) condition to a variable because it is used in multiple places.  But I don't think this logic is 100% correct because it would allow editing someone else's saved sketch if the user is not logged in.

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -588,6 +601,10 @@ class Editor extends React.Component {
 }
 
 Editor.propTypes = {
+  // eslint-disable-next-line react/forbid-prop-types
+  user: PropTypes.object.isRequired,
+  // eslint-disable-next-line react/forbid-prop-types
+  project: PropTypes.object.isRequired,
```
lindapaiste, line 607: IMO the better thing to do here would be to determine the `isReadOnly`/`canEdit` condition in a Redux selector.  For right now we would add it to `mapStateToProps` and eventually when we re-write this component we can use the same selector function in a `useSelector`.  That way we are only adding one new prop with a `boolean` value instead of passing down these complex objects.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -228,18 +237,43 @@ const MobileNav = () => {
...
+        </div>
+
+        <div>
+          {title === project.name &&
+            project.owner &&
+            user.username !== project.owner.username && (
+              // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
+              <p
+                className={`toolbar__duplicatetoedit ${
+                  theme === 'contrast' ? 'contrast' : 'normal'
```
lindapaiste, line 262: You shouldn't need to access the `theme` directly here.  You should use theme variables in the `.scss` instead.  I can explain that more later if we decide to work with this PR.

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -228,18 +237,43 @@ const MobileNav = () => {
...
+            <h5>by {project?.owner?.username}</h5>
+          )}
+        </div>
+
+        <div>
+          {title === project.name &&
+            project.owner &&
+            user.username !== project.owner.username && (
+              // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
+              <p
```
lindapaiste, line 260: The reason that you are getting all of these eslint errors is because you are using a `<p>` element instead of `<button>`.  Keyboards, screen readers, etc. will not handle it properly unless it is a `<button>`.

client/modules/IDE/components/Header/Toolbar.jsx
```diff
@@ -97,20 +100,40 @@ const Toolbar = (props) => {
...
+                  </Link>
+                </p>
+              );
+            }
+            return null;
+          })()}
+        </div>
+        {project.owner && user.username !== project.owner.username && (
+          // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
+          <p
```
lindapaiste, line 121: This needs to be a `<button>` too.

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -218,6 +219,8 @@ const IDEView = () => {
                     className="editor-preview-subpanel"
                   >
                     <Editor
+                      user={user}
+                      project={project}
```
lindapaiste, line 223: See comments re: `Editor.propTypes`.  I think that the condition which you are using currently is the equivalent of the `isUserOwner` variable here.  Could be something like `<Editor canEdit={isUserOwner} ...`

client/styles/components/_toolbar.scss
```diff
@@ -147,3 +154,36 @@
...
+
+    &.contrast {
+        background-color: #F5DC23;
+        color: black;
+    }
+
+    &.normal {
+        background-color: $p5js-pink;
+        color: white;
+    }
```
lindapaiste, line 175: See comments re `theme` in components.  We want this to look like `color: getThemifyVariable('secondary-text-color');` (with whatever the correct variable name is).

lindapaiste (changes requested):
I'm not sure about this feature because we have it tagged as a high priority and something that we definitely want to do but I personally don't really like it 🤷‍♀️.  I think it's nice to be able to edit a line or two in someone's code and see the changes without having to clone the whole thing.  @raclim what are your thoughts?

There's a lot of cleanup before this could be considered for merging.  But before you work on any of that, let's hear from @raclim about how we want this to function.

vivekbopaliya:
> I'm not sure about this feature because we have it tagged as a high priority and something that we definitely want to do but I personally don't really like it 🤷‍♀️. I think it's nice to be able to edit a line or two in someone's code and see the changes without having to clone the whole thing. @raclim what are your thoughts?
> 
> There's a lot of cleanup before this could be considered for merging. But before you work on any of that, let's hear from @raclim about how we want this to function.

Sure, i'll be on it right after we hear from @raclim, thank you for all the above quality inputs though. (always appreciate them haha). Moreover, _`"I think it's nice to be able to edit a line or two in someone's code and see the changes without having to clone the whole thing."`_  i think one of the primary reason that we are implementing this feature is to help users [distinguish ](https://github.com/processing/p5.js-web-editor/issues/870#issuecomment-470161562)between their and others sketches as it is easy to get confused here. let's see how we are tackling this one!

raclim:
> I'm not sure about this feature because we have it tagged as a high priority and something that we definitely want to do but I personally don't really like it 🤷‍♀️. I think it's nice to be able to edit a line or two in someone's code and see the changes without having to clone the whole thing. @raclim what are your thoughts?

> i think one of the primary reason that we are implementing this feature is to help users https://github.com/processing/p5.js-web-editor/issues/870#issuecomment-470161562between their and others sketches as it is easy to get confused here. let's see how we are tackling this one!

Now that I'm looking at this I'm honestly not too sure! 

I don't think I've personally had issues with the current setup, but can see how it can be hard to distinguish whether the sketch is yours or not. I'm down to implement something that signals to a user that they're working on a sketch that isn't their's more prominently, but I think I also feel hesitant about removing the edit ability altogether. 

This is straying from the original issue and is just an idea, but I think one way "removing the edit ability" could be implemented as well might be to add "Set as Read Only" or "Set as Un-copyable" options for an individual sketch, so a user can choose if they don't want their sketches to be interacted with.

vivekbopaliya:
> > I'm not sure about this feature because we have it tagged as a high priority and something that we definitely want to do but I personally don't really like it 🤷‍♀️. I think it's nice to be able to edit a line or two in someone's code and see the changes without having to clone the whole thing. @raclim what are your thoughts?
> 
> > i think one of the primary reason that we are implementing this feature is to help users [#870 (comment)](https://github.com/processing/p5.js-web-editor/issues/870#issuecomment-470161562between) their and others sketches as it is easy to get confused here. let's see how we are tackling this one!
> 
> Now that I'm looking at this I'm honestly not too sure!
> 
> I don't think I've personally had issues with the current setup, but can see how it can be hard to distinguish whether the sketch is yours or not. I'm down to implement something that signals to a user that they're working on a sketch that isn't their's more prominently, but I think I also feel hesitant about removing the edit ability altogether.
> 
> This is straying from the original issue and is just an idea, but I think one way "removing the edit ability" could be implemented as well might be to add "Set as Read Only" or "Set as Un-copyable" options for an individual sketch, so a user can choose if they don't want their sketches to be interacted with.

_`"I'm down to implement something that signals to a user that they're working on a sketch that isn't their's more prominently"`_ 
i think this can be tackled with simple "You're working on someone's sketch" the first time the user edits a sketch. 
_`might be to add "Set as Read Only" or "Set as Un-copyable" options for an individual sketch`_ 
However, this doesn't completely solve the issue, as there will be even more scenarios for users to get confused about if we don't acknowledge them that they are trying to edit a non-owned, read-only sketch.

I believe these are two different features, so there must be two different approaches for each individually:

1. End the confusion between owned and non-owned sketches (show a toast on edit or something).
2. Letting users make their sketch read-only (this time we actually remove the ability to edit and duplicate non-owned sketches).

vivekbopaliya:
> This is straying from the original issue and is just an idea, but I think one way "removing the edit ability" could be implemented as well might be to add "Set as Read Only" or "Set as Un-copyable" options for an individual sketch, so a user can choose if they don't want their sketches to be interacted with.


the idea similar to this has already been discussed [here](https://github.com/processing/p5.js-web-editor/issues/1094#issuecomment-499638522). This can address both issues: hiding the open API keys as well as removing the ability to edit sketches if they are private or set to read-only. A week ago, I wasn't in favor of letting users keep their sketches as read-only because it seemed to lead to even more confusion and lots of work that might not be worth it?! However, now it appears to be more user-friendly than simply taking away the authority to edit every sketch haha. I am down to close this pull request and create a new one!

 @raclim @catarak let me know how you want this to function.

raclim:
Sorry about that, I definitely was proposing two different features! 😅

> 1. End the confusion between owned and non-owned sketches (show a toast on edit or something).

I think I'm personally leaning towards implementing this one regardless because for me, it feels like the simplest way to address the issue of a user having difficulty distinguishing whether a sketch is theirs or not. Down to rethink some of this though if this doesn't feel like enough!

> 2. Letting users make their sketch read-only (this time we actually remove the ability to edit and duplicate non-owned sketches).

I did stray from the original issue here—I proposed this as a way to implement this feature (removing the edit ability) in a way that might feel like a choice for the user if still we did want to integrate this into the editor. 

>the idea similar to this has already been discussed https://github.com/processing/p5.js-web-editor/issues/1094#issuecomment-499638522. This can address both issues: hiding the open API keys as well as removing the ability to edit sketches if they are private or set to read-only.

I've actually been hoping to implement the privating sketches feature in the near future!

@vivekbopaliya These are all my thoughts, but I'm also wondering what makes sense to you here as well!

vivekbopaliya:
> Sorry about that, I definitely was proposing two different features! 😅
> 
> > 1. End the confusion between owned and non-owned sketches (show a toast on edit or something).
> 
> I think I'm personally leaning towards implementing this one regardless because for me, it feels like the simplest way to address the issue of a user having difficulty distinguishing whether a sketch is theirs or not. Down to rethink some of this though if this doesn't feel like enough!
> 
> > 2. Letting users make their sketch read-only (this time we actually remove the ability to edit and duplicate non-owned sketches).
> 
> I did stray from the original issue here—I proposed this as a way to implement this feature (removing the edit ability) in a way that might feel like a choice for the user if still we did want to integrate this into the editor.
> 
> > the idea similar to this has already been discussed [#1094 (comment)](https://github.com/processing/p5.js-web-editor/issues/1094#issuecomment-499638522). This can address both issues: hiding the open API keys as well as removing the ability to edit sketches if they are private or set to read-only.
> 
> I've actually been hoping to implement the privating sketches feature in the near future!
> 
> @vivekbopaliya These are all my thoughts, but I'm also wondering what makes sense to you here as well!

yes right! imo there are two ways to approach first proposal, which of one is similar to codepen's apporach.

Screenshot (234)
- In the current behavior, users can freely experiment with unowned sketches without duplication. The alternative way to duplicate a sketch is to simply save it (Ctrl + S), and it will be duplicated.

- what we can do here is, for unowned sketches, we can display a small label (kinda similar to the above PR one) but without 'to edit,' simply saying `Duplicate Sketch` or `Clone Original Sketch` This way, without restricting editing ability, we can caution the user that they are working on someone else's sketch. A label with a background color always attracts eyes' attention.

Screenshot (236)

- along with that, on the user's first edit, we can show a toast message that goes something like 'You are working on a non-owned sketch; you might want to duplicate it first.' However, personally, this feels a little spoon-feeding lol.

this was for the first proposal only. once we resolve this, afterwards we can move on to the private sketch feature and discuss how we will want it to function.

Issue #2923, opened 2024-01-19
Missing Tests for Components in User Pages Folder


gauravsingh94:
### Increasing Access

This will contribute to making the development bug-free.

### Feature enhancement details

**These files not have tests for them**:

![Screenshot from 2024-01-19 11-24-18](https://github.com/processing/p5.js-web-editor/assets/99260988/defee715-ae7f-4277-bd98-d704c837abe9)

gauravsingh94:
@lindapaiste, if tests are deemed necessary for these files, I am willing to write them all. Please assign this task to me.

lindapaiste:
@gauravsingh94 I'm generally in favor of having more tests.  Before you get writing, let's think about what specific things we want to test for and how we can implement it.

gauravsingh94:
> @gauravsingh94 I'm generally in favor of having more tests. Before you get writing, let's think about what specific things we want to test for and how we can implement it.

I would like to create tests for the following aspects:

1. Verification of the presence of required fields on the website.
2. Checking the functionality of each button to ensure they are working correctly.
3. Validating that clicking on certain elements results in appropriate changes according to specified requirements.

gauravsingh94:
> > @gauravsingh94 I'm generally in favor of having more tests. Before you get writing, let's think about what specific things we want to test for and how we can implement it.
> 
> I would like to create tests for the following aspects:
> 
> 1. Verification of the presence of required fields on the website.
> 2. Checking the functionality of each button to ensure they are working correctly.
> 3. Validating that clicking on certain elements results in appropriate changes according to specified requirements.

```ts
it('submits form with updated settings on button click', async () => {
    const dispatchMock = jest.fn(); 
    useDispatch.mockReturnValue(dispatchMock); 

    const { getByLabelText, getByText } = renderComponent();

    fireEvent.change(getByLabelText('Email'), {
      target: { value: 'newemail@example.com' }
    });
    fireEvent.change(getByLabelText('User Name'), {
      target: { value: 'newusername' }
    });
    fireEvent.change(getByLabelText('Current Password'), {
      target: { value: 'currentPassword' }
    });
    fireEvent.change(getByLabelText('New Password'), {
      target: { value: 'newPassword' }
    });

    fireEvent.click(getByText('Save All Settings'));

    await waitFor(() => {
      expect(dispatchMock).toHaveBeenCalledWith(
        updateSettings({
          email: 'newemail@example.com',
          username: 'newusername',
[...]

PR #2955, opened 2024-01-25
Issue-#2923 added test for the Account Form.

gauravsingh94:
Ref #2923

Changes:
Added the test for AccountForm.


gauravsingh94:
@lindapaiste 

I have added the test for Account from please review it. if these test feel good to you then i started to write for the other component as well.

In the `AccountForm` test  i have added the test for :

- The form component renders correctly with the initial user values.
- clicking on `save all settings` will call the dispatch the `updateSettings`.
- when out data is submitting in this period our `save all settings ` button must be `disabled`.



image

lindapaiste (dismissed):
These are very good tests! 👍 

I'm often not a fan of mocking because it's testing a fake behavior and not the real behavior.  But this is a case where it makes sense because I don't think that our dev/test setup has actual user accounts associated with it.  Maybe it does?  We should look into that.  When it comes to testing the login screen it would be cool if we could try submitting a wrong password and verify that it fails vs a correct password and verify that it succeeds.

gauravsingh94:
@lindapaiste 
Added the tests for the LoginForm.

gauravsingh94:
@lindapaiste 
Added the test for the `NewPasswordForm`.

image

khanniie:
thanks, these are great tests! : ) lgtm to me overall but you will probably have to move the test locations as it looks like the files have moved locations a little bit

raclim (approved):

PR #2970, opened 2024-01-27
fixes/ Semantic Html Issues

### LINKED ISSUE #2967: Semantic HTML issues in header section
### Increasing Access

Using proper semantic HTML allows screen readers to understand our content.

I found two issues relating to the header section while running an accessibility check:

1. We have a `<header>` element inside another `<header>` element, and multiple `<header>` elements on mobile.  We need to get this down to one.  We can change incorrect ones to either a React Fragment or a `<div>` or potentially remove them from the tree entirely.
2. We have a `<span>` as a direct child of a `<ul>` in the "login or signup" part of the nav bar.  Probably we change the `<span>or</span>` to `<li>or</li>`.  It would also be acceptable to have the entire "login or signup" as one list item, but making the "or" into its own list item is probably easier to implement.

We want to fix these HTML structures without changing the style.

Mubashirshariq:
Fixes #2967 

Changes:

### DIFF
--- a/client/modules/IDE/components/Header/index.jsx
+++ b/client/modules/IDE/components/Header/index.jsx
@@ -9,7 +9,7 @@ const Header = (props) => {
   const project = useSelector((state) => state.project);
 
   return (
-    <header>
+    <>
       <Nav />
       <MediaQuery minWidth={770}>
         {(matches) => {
@@ -23,7 +23,7 @@ const Header = (props) => {
           return null;
         }}
       </MediaQuery>
-    </header>
+    </>
   );
 };
 
--- a/client/modules/IDE/components/Editor/MobileEditor.jsx
+++ b/client/modules/IDE/components/Editor/MobileEditor.jsx
@@ -9,7 +9,7 @@ export const EditorContainer = styled.div`
   transform: ${(props) =>
     props.expanded ? 'translateX(50%)' : 'translateX(0)'};
 
-  > header {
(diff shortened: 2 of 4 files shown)


Mubashirshariq:
@lindapaiste  plz review

client/modules/IDE/components/Header/index.jsx
```diff
@@ -9,7 +9,7 @@ const Header = (props) => {
   const project = useSelector((state) => state.project);
 
   return (
-    <header>
+    <>
```
lindapaiste, line 12: I wonder if it makes sense to keep the `<header>` here and remove the one in in the `NavBar` component instead?  It's subjective.  It depends on whether we consider the toolbar with the play button etc. to be part of the header or not.  I can see the arguments for both ways.

Mubashirshariq: In my perspective, centralizing the header in the NavBar component aligns with a cleaner and more cohesive design, emphasizing a unified user interface.

Mubashirshariq: But still if you want i will remove the one in the Navbar  one and keep this one.

raclim:
Closing due to addressed by https://github.com/processing/p5.js-web-editor/pull/3130. I'm sorry that we couldn't get this in, but please feel free to check out our other issues!

PR #2376, opened 2023-08-13
Common Table component (part 1)

lindapaiste:
Fixes #1442
Fixes #2375 
Progress on #2358 

This is part 1 because I have cleaned up the __Table components, but not the __TableRow components.

Code cleanup:
- Create a `TableBase` component to handle common logic of tables:
  - Show `<Loader/>` when loading.
  - Show "no results" message when empty.
  - Render column headers with the name of each column.
  - Highlight the active column with underline and triangle up/down icon.
  - Sort the items in the table when clicking on the column header.
- Write tests for this core logic of the `TableBase` and `TableHeaderCell`.
- Use the `TableBase` in existing components:
  - `AssetList` - converted to function component
  - `CollectionList` - converted to function component
  - `SketchList` - converted to function component
  - `Collection` - still a class component because I have other open PRs that touch this (#2253)
- Sorting is handled internally by the `TableBase`, so remove `sorting` from Redux.
   - Delete the `sorting` reducer, actions, and selectors.
   - Rework one test of `SketchList` which checked redux actions.
- Use class `sketches-table` on all tables (can add additional classes alongside it), and delete SCSS properties on `asset-table` which are duplicates.
- Migrate to `styled-components`:
  - All table header styles.
  - Empty message style.

Bug Fixes:
- Sketches table in `Collection` component can be sorted (#2375)
- `SketchList` column header "Sketch" now shows the correct `aria-label`.  It was "Sort by Sketch descending." but should be "ascending".

New Features:
- `AssetList` can be sorted.
- Added ARIA attributes [`aria-sort`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort) and `aria-pressed` to column headers.

I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] has no test errors (`npm run test`)
* [x] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #2971, opened 2024-01-28
Refactored CollectionList and CollectionListRow

vivekbopaliya:
Related to #2179 and #824 

Changes:
- Migrated CollectionList into functional component
- Shifted them from props drilling to redux hooks

### DIFF
--- a/client/modules/IDE/components/CollectionList/CollectionList.jsx
+++ b/client/modules/IDE/components/CollectionList/CollectionList.jsx
@@ -1,4 +1,4 @@
-/* eslint-disable react/react-in-jsx-scope */
+import React from 'react';
 import PropTypes from 'prop-types';
 import { Helmet } from 'react-helmet';
 import { useDispatch, useSelector } from 'react-redux';
@@ -25,12 +25,10 @@ const CollectionList = (props) => {
   const { project_id: projectId } = useParams() || { project_id: null };
   const [hasLoadedData, setHasLoadedData] = useState(false);
 
-  const { user, sorting, loading, collections } = useSelector((state) => ({
-    user: state.user,
-    sorting: state.sorting,
-    loading: state.loading,
-    collections: getSortedCollections(state)
-  }));
+  const user = useSelector((state) => state.user);
+  const sorting = useSelector((state) => state.sorting);
+  const loading = useSelector((state) => state.loading);
+  const collections = useSelector((state) => getSortedCollections(state));
 
   const [
     addingSketchesToCollectionId,
@@ -175,7 +173,6 @@ const CollectionList = (props) => {
                 user={user}
                 username={props.username}
                 onAddSketches={() => showAddSketches(collection.id)}
-                t={t}
               />
             ))}
           </tbody>
--- a/client/modules/IDE/components/CollectionList/CollectionList.jsx
+++ b/client/modules/IDE/components/CollectionList/CollectionList.jsx
@@ -1,313 +1,206 @@
+/* eslint-disable react/react-in-jsx-scope */
 import PropTypes from 'prop-types';
-import React from 'react';
 import { Helmet } from 'react-helmet';
-import { withTranslation } from 'react-i18next';
-import { connect } from 'react-redux';
-import { bindActionCreators } from 'redux';
+import { useDispatch, useSelector } from 'react-redux';
+import { useState, useEffect } from 'react';
+import { useTranslation } from 'react-i18next';
 import classNames from 'classnames';
+import { useParams } from 'react-router-dom';
 import { find } from 'lodash';
-import * as ProjectActions from '../../actions/project';
-import * as ProjectsActions from '../../actions/projects';
-import * as CollectionsActions from '../../actions/collections';
-import * as ToastActions from '../../actions/toast';
-import * as SortingActions from '../../actions/sorting';
-import getSortedCollections from '../../selectors/collections';
 import Loader from '../../../App/components/loader';
 import Overlay from '../../../App/components/Overlay';
 import AddToCollectionSketchList from '../AddToCollectionSketchList';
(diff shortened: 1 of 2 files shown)


client/modules/IDE/components/CollectionList/CollectionListRow.jsx
```diff
@@ -1,21 +1,17 @@
...
-import { bindActionCreators } from 'redux';
-import { withTranslation } from 'react-i18next';
 import MenuItem from '../../../../components/Dropdown/MenuItem';
 import TableDropdown from '../../../../components/Dropdown/TableDropdown';
-import * as ProjectActions from '../../actions/project';
-import * as CollectionsActions from '../../actions/collections';
-import * as IdeActions from '../../actions/ide';
-import * as ToastActions from '../../actions/toast';
 import dates from '../../../../utils/formatDate';
+import { deleteCollection, editCollection } from '../../actions/collections';
```
lindapaiste, line 9: Crazy how many of the previous `import * as` actions were not actually used!  This is why I prefer importing the individual action creators like you are doing here. ✔️

client/modules/IDE/components/CollectionList/CollectionList.jsx
```diff
@@ -1,313 +1,206 @@
...
+          </thead>
+          <tbody>
+            {collections.map((collection) => (
+              <CollectionListRow
+                mobile={props.mobile}
+                collection={collection}
+                user={user}
+                username={props.username}
+                onAddSketches={() => showAddSketches(collection.id)}
+                t={t}
```
lindapaiste: Personally I would use another `useTranslation` for this component rather than passing down `t` from the parent.  Though I don't actually have any docs to back up if one way is better than the other.  I tried a quick search to find some authoritative reference but all I got was [one person on StackOverflow.](https://stackoverflow.com/questions/65422274/react-i18next-passing-t-function-to-all-children-or-usetranslation-in-every-chi)

client/modules/IDE/components/CollectionList/CollectionList.jsx
```diff
@@ -1,313 +1,206 @@
+/* eslint-disable react/react-in-jsx-scope */
```
lindapaiste: You should import `React` rather than disabling the warning.  `import React, { useState, useEffect } from 'react';`

client/modules/IDE/components/CollectionList/CollectionList.jsx
```diff
@@ -1,313 +1,206 @@
...
+  const [hasLoadedData, setHasLoadedData] = useState(false);
 
-    this.props.getCollections(this.props.username);
-    this.props.resetSorting();
+  const { user, sorting, loading, collections } = useSelector((state) => ({
+    user: state.user,
+    sorting: state.sorting,
+    loading: state.loading,
+    collections: getSortedCollections(state)
+  }));
```
lindapaiste: It is better to use individual selectors for each variable because it reduces unnecessary re-renders.  More info at https://react-redux.js.org/api/hooks#equality-comparisons-and-updates

lindapaiste (changes requested):
This looks pretty good for the most part, but we have some other proposed changes to the `CollectionList` component which will definitely conflict so I don't know what the plan is. #2376

vivekbopaliya:
> This looks pretty good for the most part, but we have some other proposed changes to the `CollectionList` component which will definitely conflict so I don't know what the plan is. #2376

Ahhhhh, it appears that several components were optimized along with two new Table and TableRow have been added. i didn't check the PR. my bad. although as it is pretty huge PR and i see it is in still development (as some of test cases got failed ;) ) hence, we are not sure when will it get merged as yet. in meantime, i am moving forward and fix the changes that you requested.

client/modules/IDE/components/CollectionList/CollectionList.jsx
vivekbopaliya: i have seen a lot of times `t` is being passed to child components throughout the codebase.

vivekbopaliya:
Additionally, we might want to seperate `SketchListRow` from `SketchList` and place it in its own file. I have worked on it and almost did wrap up migrating them into functional components and importing individual actions rather than prop drilling but eventually, it boiled down to failing one of 20 test errors even tho both components are completely working. I think while refactoring them, we will have to rewrite `SketchList.unit.test` component as well. 


Although I have never worked on testing before 😬, let me know if you want me to raise a PR, and we can further discuss what changes are required in the unit test component. (could be a mini collab?!)

client/modules/IDE/components/CollectionList/CollectionList.jsx
lindapaiste: Yeah there is a lot of that in the old/existing code.  Usually when I rewrite components I change it.  But like I said I'm not certain it's actually better, it might just be my personal preference.

lindapaiste:
> > This looks pretty good for the most part, but we have some other proposed changes to the `CollectionList` component which will definitely conflict so I don't know what the plan is. #2376
> 
> Ahhhhh, it appears that several components were optimized along with two new Table and TableRow have been added. i didn't check the PR. my bad. although as it is pretty huge PR and i see it is in still development (as some of test cases got failed ;) ) hence, we are not sure when will it get merged as yet. in meantime, i am moving forward and fix the changes that you requested.

Yeah everything regarding the list/table components has been kind of up in the air due to this major rewrite which has been in the works for a long time.

raclim:
Thanks so much for submitting this and I'm sorry it took a while to get to this! 

Although it's a bigger PR, it does address a lot of the refactoring needs so I think we ultimately might want to go with https://github.com/processing/p5.js-web-editor/pull/2376, since it was raised first! In that vein, I'm going to close this one for now, but maybe some changes here could also be referenced in the other one as well?

PR #2982, opened 2024-02-03
Zod validationn added for client side

### LINKED ISSUE #2964: Using Zod Library to validate inputs entered by the users in signup form and login forms.
### Increasing Access

With this feature I will add input validations on every input of login and signup form so that the data entered by the user is correct when enters the database and do verification and other things. This validation will be done in both frontend and backend. Instead of creating some functions for the input validations we will use this library , this will make our code neat and clean as well.  

### Feature request details

 Will use Zod Library to validate inputs entered by the users in signup form and login forms. I will add this library and make validations part in common folder so that it could be used in both frontend and backend. 

Hi @lindapaiste please assign this to me I would like to work on this issue

rahulrana701:
Fixes #2964

Changes:

* [X] is descriptively named and links to an issue number, i.e. `Fixes #2964 `


rahulrana701:
hi @lindapaiste I have added the zod validation for login and signup form for the client side which cleaned up a lot of code , I didn't added this validation for backend as you were not sure whether there was a validation logic for backend already , if yes I will update this pull request as per recommendations.

client/common/zod.js
```diff
@@ -0,0 +1,29 @@
...
+const EMAIL_REGEX = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/i;
+/* eslint-enable */
+
+const SignupFormInput = z.object({
+  username: z
+    .string()
+    .min(1)
+    .max(20)
+    .regex(/^[a-zA-Z0-9._-]{1,20}$/),
+  email: z.string().email().regex(EMAIL_REGEX),
```
lindapaiste: I wonder if the Zod .email() function checks for the same stuff as our EMAIL_REGEX?  It might be that we can delete the regex check — but I need to check the Zod docs and/or source code to know.

client/common/zod.js
```diff
@@ -0,0 +1,29 @@
...
+/* eslint-disable */
+const EMAIL_REGEX = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/i;
+/* eslint-enable */
+
+const SignupFormInput = z.object({
+  username: z
+    .string()
+    .min(1)
+    .max(20)
+    .regex(/^[a-zA-Z0-9._-]{1,20}$/),
```
lindapaiste: The existing regex says that it must be between 1 and 20 characters and can only use certain characters: letters, numbers, dots, underscores and dashes.

The way that Zod breaks it up into different checks is very helpful because it allows us to show more specific messages, like “username cannot be longer than 20 characters” or “username contains invalid characters” (would be nice if we could say specifically which character).  But I don’t think we have translations for all of those, so this might be something where we make some improvements now but can continue to refine it over time.

client/common/zod.js
```diff
@@ -0,0 +1,29 @@
...
+  email: z.string().email().regex(EMAIL_REGEX),
+  password: z.string().min(6),
+  confirmPassword: z
+    .string()
+    .min(6)
+    .refine((confirmPassword, data) => confirmPassword === data.password, {
+      message: "Passwords don't match",
```
lindapaiste: We need to tie this sort of message into our translation system.

client/modules/User/components/SignupForm.jsx
```diff
@@ -43,16 +35,12 @@ function SignupForm() {
...
       onSubmit={onSubmit}
     >
       {({ handleSubmit, pristine, submitting, invalid }) => (
         <form className="form" onSubmit={handleSubmit}>
-          <Field
-            name="username"
-            validate={validateUsername}
-            validateFields={[]}
-          >
+          <Field name="username" validateFields={[]}>
```
lindapaiste: The signup form has an asynchronous check where we ping the API to make sure that the username and email address are not already in use.

We definitely don’t want to lose that check.  We would want to either migrate it to Zod, or run it as an independent function on those form fields.

I think that keeping it independent might make sense because as it is right now we are pinging to API more than is necessary.  So it’s been on my list of low-priority things that I want to refactor.

That is, unless Zod has some sort of utility that allows us to call the API only when the username has changed.  That would be great if such a thing exists.  Otherwise I need to write one.

client/utils/reduxFormUtils.js
```diff
@@ -54,50 +54,6 @@ export function validateSettings(formProps) {
...
-  return errors;
-}
-
-function validatePasswords(formProps, errors) {
-  if (!formProps.password) {
-    errors.password = i18n.t('ReduxFormUtils.errorEmptyPassword');
-  }
-  if (formProps.password && formProps.password.length < 6) {
-    errors.password = i18n.t('ReduxFormUtils.errorShortPassword');
-  }
```
lindapaiste, line 74: You can see how the deleted code has specific translated messages for various failures.  We need to integrate those into the Zod code.  `.min(6, { message: i18n.t('ReduxFormUtils.errorShortPassword') })`.

client/common/zod.js
lindapaiste: I confirmed that Zod has their own email regex: https://github.com/colinhacks/zod/blob/a5a9d31018f9c27000461529c582c50ade2d3937/src/types.ts#L567-L568

client/common/zod.js
```diff
@@ -0,0 +1,29 @@
...
+      path: ['confirmPassword']
+    })
+});
+
+const LoginFormInput = z.object({
+  email: z.string().email(),
+  password: z.string().min(6)
+});
+
+export default { SignupFormInput, LoginFormInput };
```
lindapaiste: I'm getting errors when running your branch due to the way that you are exporting and importing.  You want to export the individual validators separately.  Not `export default`.

client/modules/User/components/SignupForm.jsx
```diff
@@ -43,16 +35,12 @@ function SignupForm() {
   return (
     <Form
       fields={['username', 'email', 'password', 'confirmPassword']}
-      validate={validateSignup}
+      validate={(values) => SignupFormInput.safeParse(values)}
```
lindapaiste, line 46: The `validate` function here needs to return an object of errors, where the keys are the fields and the values are the error messages.

I'm playing around and I've got this
```
validate={(values) => {
  const parsed = SignupFormInput.safeParse(values);
  console.log(parsed);
  return parsed.error.formErrors.fieldErrors;
}}
```
but it's just showing "Required" and not the custom messages (which I put in)

lindapaiste: Ah ok this format is working:
```
username: z
    .string({
      required_error: i18n.t('ReduxFormUtils.errorEmptyUsername')
    })
    .max(20, { message: i18n.t('ReduxFormUtils.errorLongUsername') })
    .regex(/^[a-zA-Z0-9._-]{1,20}$/, {
      message: i18n.t('ReduxFormUtils.errorValidUsername')
    }),
  email: z
    .string({
      required_error: i18n.t('ReduxFormUtils.errorEmptyEmail')
    })
    .email({ message: i18n.t('ReduxFormUtils.errorInvalidEmail') }),
```

Putting an error on the `min(1)` didn't do it.  Seems like we need to use `required_error` to handle empty field errors.

When playing around it is helpful to set your language to something other than English so you can see that the messages are getting translated.

rahulrana701:
@lindapaiste I will solve everything by today.

client/common/zod.js
rahulrana701: @lindapaiste so should I use the zod email regex or this one is fine

rahulrana701:
hi @lindapaiste I have made some changes as you suggested.

rahulrana701:
@lindapaiste should I add this validation logic in the server side as well ??

lindapaiste:
> @lindapaiste should I add this validation logic in the server side as well ??

Focus on getting the front-end to work.  Are you seeing the correct errors when running your branch?  You haven't addressed some of the things that I pointed out so probably not.

To be honest I'm not sure where we stand on Zod right now.  It can be very clean when you're just doing simple stuff like `.min(1).max(20)`, but now that we put in the proper translations and everything it's not obvious that this is an improvement over the existing logic.

rahulrana701:
@lindapaiste, you are correct that adding translations in Zod won't be as effective since the code hasn't been cleaned up as much as we wanted. However, it still cleans up a bunch of code. This is my first time working with translations, so we might encounter some errors. I apologize for any inconvenience caused, but I'll definitely make it work for the frontend.

PR #2382, opened 2023-08-16
[WIP] Common Table component (part 2)

lindapaiste:
Fixes #1442

Merge after #2376 and #2379 

Changes:
- Create a common table component for `SketchList` and `CollectionList` which allow renaming.  This is a little bit tricky as the columns cannot be completely independent, since clicking the "Rename" option in the dropdown column causes changes in the "name" column.

I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] has no test errors (`npm run test`)
* [x] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #2988, opened 2024-02-04
Fix : [#2986] Add ref callback

### LINKED ISSUE #2986: Collections Rename Input Fails to Autofocus.
### Actual Behavior

When attempting to rename a collection, the input field does not automatically focus, leading to a slight inconvenience in the user experience.

### Expected Behavior

Upon pressing the rename button, the input field should autofocus, similar to the behavior observed when renaming a sketch

### Steps to reproduce:
1. Navigate to a collections page.
2.  Select the rename option from the dropdown of any collection.
(You have to manually click on the input field or click on the rename option again to focus.)

Ri-Sharma:
Fixes #2986

Changes:
1.  Added ref callback function to resolve the issue

* [x] is descriptively named and links to an issue number.


client/modules/IDE/components/CollectionList/CollectionListRow.jsx
```diff
@@ -116,7 +119,10 @@ const CollectionListRowBase = (props) => {
             onKeyUp={handleRenameEnter}
             onBlur={handleRenameBlur}
             onClick={(e) => e.stopPropagation()}
-            ref={renameInput}
+            ref={(node) => {
+              renameInput.current = node;
+              handleRenameFocus();
+            }}
```
lindapaiste, line 200: Does the ref function get executed every time that the element is mounted?  Like if you open it and close it and open it again, will it get focused the second time?  I think so but I'm not certain off the top of my head.

lindapaiste (commented):
Overall this looks good, but I'm not sure what to do with it because we've been working on some larger changes to the tables which would overwrite this.  Specifically, the [`RenameInput`](https://github.com/processing/p5.js-web-editor/blob/88a49610b3a97010f2c078abbf702458e7c68301/client/common/Table/TableWithRename.jsx#L6-L48) and [`TableWithRename`](https://github.com/processing/p5.js-web-editor/blob/88a49610b3a97010f2c078abbf702458e7c68301/client/common/Table/TableWithRename.jsx#L68-L86) in #2382.  (It's been so long that I don't even remember which parts were and were not finished).

If we have this problem in the `CollectionListRow` then odds are that we also have it in the `SketchListRow`.  Those components are very similar and I don't like the current system where we have to make analogous changes in both places when we see an issue like this.

client/modules/IDE/components/CollectionList/CollectionListRow.jsx
Ri-Sharma: > will it get focused the second time?

Yes, it will. The way it works is when the component gets mounted the ref callback is called with the input field as an parameter. And when the component gets unmounted the callback is called again with null as an parameter.

Ri-Sharma:
> If we have this problem in the `CollectionListRow` then odds are that we also have it in the `SketchListRow`. Those components are very similar and I don't like the current system where we have to make analogous changes in both places when we see an issue like this.

The `SketchListRow` is working fine and the problem only exists with the collections. I think this is because the `SketchListRow` uses class components where as the `CollectionListRow` uses funtional component and for some reason same logic is working for the classes but not for the functions.

raclim (approved):
Thanks for your work on this!

PR #3008, opened 2024-02-08
Drag n drop

### LINKED ISSUE #1417: Drag and Drop Files into Folder
#### Nature of issue?

- New feature request

#### New feature details:

So I noticed how when you create a folder in p5.js, you can't drag and drop existing files into that folder. Instead you have to manually import files into that specific folder. So I'm wondering, could it be a possibility that you could drag and drop existing files into a folder? Thanks

mhsh312:
Fixes #1417

Changes: 
- Added Drap and Drop functionality to transfer files among folders in the Sidebar
- Also supports transferring one folder into another folder

Ui Changes:
- Used `HTML Drag and Drop API `
- Could not use `React DnD` as initially planned because **[it does not support class components](https://stackoverflow.com/questions/61645035/how-to-implement-drag-and-drop-in-class-based-component-in-react-dnd)** (FileNode.jsx)

Redux Changes:
- Made a new reducer `changeParent` 
- Added actions and constants to facilitate said reducer
- Made it so that the changes done in the UI reflects in the `file state`

Saving the sketch reflects the state changes in the database as well.

https://github.com/processing/p5.js-web-editor/assets/135870090/e25e6779-a484-457e-a00a-d25bb924bdd1


lindapaiste:
> * Could not use `React DnD` as initially planned because **[it does not support class components](https://stackoverflow.com/questions/61645035/how-to-implement-drag-and-drop-in-class-based-component-in-react-dnd)** (FileNode.jsx)

This is a bummer and I wonder if we want to convert the FileNode first?  There’s definitely an open PR for that, though I recall that it had a minor issue with the dropdown.

I have adjustments that I would want to make here but I’m on vacation this weekend so I’ll get back to you with more details later.

client/modules/IDE/components/FileNode.jsx
```diff
@@ -230,6 +230,45 @@ class FileNode extends React.Component {
...
+        '1px solid #ed225d';
+    }
+  };
+
+  handleDragLeave = (e) => {
+    e.preventDefault();
+    if (this.props.fileType === 'folder') {
+      e.target.parentNode.parentNode.style.border = null;
+    } else if (this.props.fileType === 'file') {
+      e.target.parentNode.parentNode.parentNode.parentNode.parentNode.style.border = null;
```
lindapaiste, line 252: We want to avoid direct manipulation of styles and instead set some state somewhere which controls the style.  I’m also concerned that the number of parent levels that we need to go up by might depend on where this specific file I in the tree, and could be incorrect if it’s a few levels deep.

mhsh312:
> I wonder if we want to convert the FileNode first? 

That is definitely the way to go. This way we can take advantage of all the little developer experience features of react dnd as well. This will also fix the issue you talked about in the commit comment as well, since all the styling will be handled differently in react-dnd. It'll also make it eaiser for everyone if in the future the dnd implementation has to be improved upon or modified. 

And I'd be down to work on this feature with react-dnd and other modifications after the FileNode has been updated. Till then enjoy your vacation :)

RAHULDAS6009:
@lindapaiste can i take the issue of converting to functional components(covert to FileNode)

> > * Could not use `React DnD` as initially planned because **[it does not support class components](https://stackoverflow.com/questions/61645035/how-to-implement-drag-and-drop-in-class-based-component-in-react-dnd)** (FileNode.jsx)
> 
> This is a bummer and I wonder if we want to convert the FileNode first? There’s definitely an open PR for that, though I recall that it had a minor issue with the dropdown.
> 
> I have adjustments that I would want to make here but I’m on vacation this weekend so I’ll get back to you with more details later.

raclim:
Since it's been a while since this PR has been last active, I'm going to close this for now. I'm sorry that we couldn't get this in, but please feel free to revisit this issue!

PR #3013, opened 2024-02-11
error added that the new password entered should be different from the current password

### LINKED ISSUE #3012: No Confirmation that the new password that is entered should be different from current password.
### Increasing Access

Earlier, what we were doing was simply updating the account settings even though the new password entered by the user was the same as the previous password. Therefore, we need to avoid this unnecessary action of calling our database and updating the settings. Because it doesn't makes sense to update the same password
Screenshot 2024-02-11 113044

### Feature request details

With this feature we are addressing a warning that the new password entered should be different from current password because then only it makes sense to update the account settings or password like this 

Screenshot 2024-02-11 131810

rahulrana701:
Fixes #3012

Changes:

* [X] is descriptively named and links to an issue number, i.e. `Fixes #3012 `


rahulrana701:
@lindapaiste @raclim  I have added the error message that the new password entered should be different from current password when updating the account settings.

raclim (approved):
Thanks for your work on this! I also made a few translations changes and added in filler ones to the other languages, but overall this looks good!

Issue #3004, opened 2024-02-08
Migrating codebase from `javascript` to `typescript`


swrno:
### Increasing Access

By migrating the codebase to TypeScript, we aim to increase accessibility to developers who are familiar with statically typed languages and who prefer the robust type checking provided by TypeScript. This change will facilitate a smoother onboarding process for new contributors and help improve collaboration within the open-source community.

### Feature enhancement details

TypeScript offers static type checking, which helps catch potential errors during development, leading to more reliable and maintainable code.

TypeScript provides enhanced IDE support with features like code completion, refactoring tools, and better documentation integration, improving the development experience.

With TypeScript, navigating through the codebase becomes more efficient due to its support for defining explicit types and interfaces, enabling developers to quickly understand the structure of the application.

But, first we have to define the steps to migrate to Typescript.

rahulrana701:
Migrating only client side part to typescript is not a feasible approach in my opinion  , as we are building the whole project 
together via webpack and if we convert only the client side into typescript  we may get a lot of errors on the go  , so it is  better to migrate the whole project from Javascript to typescript but in this approach of migrating the whole project , we have to make some changes in the webpack files so that it could convert the ts files into js files and we have to make  lot of other changes in the backend code and client code as well , ideal and best approach is to use monorepos or turborepos but in this approach we may have to bootstrap the project from start in some cases , @lindapaiste @raclim  what are your views on this , if you want to convert this project from javascript to typescript I can give it a try you , I will open a new issue for that and you could assign me that and after that I will start working on it.

swrno:
Well I think, this is not a issue which can be solved by a single person, the codebase is so large, generally when we convert any project from `.js` to `.ts` for type safety and for less errors in long run. 

Your enthusiasm is quite appreciated @rahulrana701, but as I opened this issue, let me work on this, if at any point of time I need your help, definitely I will ask you. 

@raclim @lindapaiste can you assign this issue to me, I will not delete the `.js` or `.jsx` files, instead of that I will try do go with the basic configuration and convert the client side code to typescript. after that step by step, We can convert the whole codebase to `typescript`.

rahulrana701:
@Swarnendu0123, you can absolutely work on this issue if you want, brother. It's just that you didn't mention it initially, so I thought you didn't want to work on it. But it's okay. In my opinion, it's not so large that we need more than one developer to convert it to TypeScript; it can be done alone. However, you can continue

swrno:
[...]

PR #3015, opened 2024-02-12
Basic Config for `typescript` and changed user model to test the system.

### LINKED ISSUE #3007: Migrating the whole codebase from `javascript` to `typescript`
### Increasing Access

By migrating the codebase to TypeScript, we aim to increase accessibility to developers who are familiar with statically typed languages and who prefer the robust type checking provided by TypeScript. This change will facilitate a smoother onboarding process for new contributors and help improve collaboration within the open-source community.

On a very high level we have 3 main folders, which we can convert into typescript:

- [ ] client
- [ ] server
- [ ] webpack

swrno:
Ref #3004

Changes:
Hello @lindapaiste as you mentioned, I have created the Basic configurations for `typescript` and changed the `server\models\user.js` to `server\models\user.ts` to test the system. But still I am getting some error in `server\models\user.ts`

```javascript
const callback = typeof options === 'function' ? options : cb;
  if (isEmail) {
    return this.findByEmail(value, callback);
  }
  return this.findByUsername(value, callback);
};
```
In the above snipped, its looks like there are some problem in the definition of `findByEmail` and `findByUsername`.

```javascript
userSchema.statics.EmailConfirmation = EmailConfirmationStates;
```
in the above code, the `userSchema.statics.EmailConfirmation` is throwing some error.
can you please suggest how to came out this error?


lindapaiste:
My recollection of working with Mongoose and TypeScript is that it's not as automatic as you would hope.  We probably need to annotate our extra methods somewhere.  I'll take a closer look later.  Also there are some packages for using Mongoose + TS together which may or may not be worth using.  If the main goal of this PR is to establish a configuration, then maybe we choose a simpler test file.  Like something in `/server/utils`.

lindapaiste:
Look at that, we found an actual error!  Thank you TypeScript!
image

I've got to look into where we are using that `findByEmailOrUsername` function and why we haven't noticed that problem before.

---

I got it (mostly) working but it's extremely repetitive.  There's got to be a better way.

```
type FindUserCallback = (err: Error | null, doc?: User) => void;

interface UserStatics {
  findByEmail: (
    email: string | string[],
    cb?: FindUserCallback
  ) => void;
  findByUsername: (
    username: string,
    options?: { caseInsensitive?: boolean },
    cb?: FindUserCallback
  ) => void;
  findByEmailOrUsername: (
    value: string,
    options?: { caseInsensitive?: boolean; valueType?: 'email' | 'username' },
    cb?: FindUserCallback
  ) => void;
  findByEmailAndUsername: (
    email: string,
    username: string,
    cb?: FindUserCallback
  ) => void;
  EmailConfirmation: typeof EmailConfirmationStates;
}

interface UserModel extends Model<User>, UserStatics {
  statics: UserStatics;
}

const userSchema = new Schema<User, UserModel>(
```

Then use `UserModel` where you have `Model<User>`.

swrno:
@lindapaiste I tried. but unfortunately, I am getting a lots of error. Can you please go through this PR, and and push a commit over this PR?

lindapaiste:
Looks like we need to make some changes to the Jest setup so that Jest can read the `.ts` file.

swrno:
But when I tried to set up the configurations, the main error was some eslint version error, maybe that is the root of all of those errors.

I'm quite proficient in writing tests using Jest or Enzyme, both of which we already use in our current repository. I can assist in creating additional test files and enhancing them by incorporating type safety.

Let me give it a try, again.

lindapaiste:
> But when I tried to set up the configurations, the main error was some eslint version error, maybe that is the root of all of those errors.

Yes omg I have had an open PR to fix that for ages.  Every time you install a package you have to use `--legacy-peer-deps` or else it won't work.  If that's the issue.  Maybe not.  I was getting some sort of error with using the `"@typescript-eslint"` eslint plugin.  So I removed that but then I was getting errors about missing file extensions from the "import" plugin.  I can try to help out but I'm not sure when I'll have time to take a deeper look.

swrno:
> Every time you install a package you have to use --legacy-peer-deps or else it won't work. If that's the issue.

🥲 Unfortunately yes! Additionally "can't use `import` statement outside of a module"

server/models/user.ts
```diff
@@ -302,20 +296,11 @@ userSchema.statics.findByEmailOrUsername = function findByEmailOrUsername(
...
-/**
- *
- * Queries User collection, performing a MongoDB logical or with the email
- * and username (i.e. if either one matches, will return the first document).
- *
- * @param {string} email
- * @param {string} username
- * @callback [cb] - Optional error-first callback that passes User document
- * @return {Promise<Object>} - Returns Promise fulfilled by User document
- */
```
lindapaiste: We do want to keep a JSDoc block to describe the function, if one exists already.  In this case the description of the function is helpful because it describes what the function does.

I'm not certain of the best way to handle the descriptions of the parameters.  I think the most common is to keep the description text but remove the types.  We can move those comments inline between the parameters but I'm not sure if that gets the same intellisense support?  I have to play.  It's also possible that we don't really need them in the majority of cases because they don't add any information that's not already in your new types.  Like "Optional error-first callback that passes User document" doesn't tell us anything that `cb?: (err: Error | null, doc?: User) => void` doesn't.  The types indicate that `cb` is optional, the error is the first argument, the value is a `User`, and more.

lindapaiste:
BTW the other TypeScript error that I was seeing has to do with setting `userSchema.statics.EmailConfirmation = EmailConfirmationStates;` as the types for `statics` indicate that is should be used for functions, and we are using it to store a dictionary object.  TBH I'm more of a frontend person than a backend person so I'm not sure if there's a better or different place to store those values.  Possibly that's an error that we would be okay with ignoring, at least in the short term.

swrno:
> BTW the other TypeScript error that I was seeing has to do with setting `userSchema.statics.EmailConfirmation = EmailConfirmationStates;` as the types for `statics` indicate that is should be used for functions,

Yes, I have already mentioned that in the PR description.

> and we are using it to store a dictionary object.  TBH I'm more of a frontend person than a backend person so I'm not sure if there's a better or different place to store those values.  Possibly that's an error that we would be okay with ignoring, at least in the short term.

Let me do some sort of research on it. Most probably I will come up with a solution.

>Maybe not. I was getting some sort of error with using the "@typescript-eslint" eslint plugin. So I removed that but then I was getting errors about missing file extensions from the "import" plugin.

But I need your help in this.

Swarnendu0123:
> BTW the other TypeScript error that I was seeing has to do with setting `userSchema.statics.EmailConfirmation = EmailConfirmationStates;` as the types for `statics` indicate that is should be used for functions,

Yes, I have already mentioned that in the PR description.

> and we are using it to store a dictionary object.  TBH I'm more of a frontend person than a backend person so I'm not sure if there's a better or different place to store those values.  Possibly that's an error that we would be okay with ignoring, at least in the short term.

Let me do some sort of research on it. Most probably I will come up with the solution.

>Maybe not. I was getting some sort of error with using the "@typescript-eslint" eslint plugin. So I removed that but then I was getting errors about missing file extensions from the "import" plugin.

But I need your help in this.

lindapaiste:
FYI there's also some trickiness with the Mongoose `exec()` function because their types declare that it returns `void` if a callback is provided or a `Promise` if there is no callback.

This is what's in the types for `mongoose` version "5.13.20" which we are using:

```
    /** Executes the query */
    exec(): Promise<ResultType>;
    exec(callback?: Callback<ResultType>): void;
    // @todo: this doesn't seem right
    exec(callback?: Callback<ResultType>): Promise<ResultType> | any;
```

In the long term we can avoid that issue entirely if we modernize the codebase and stop using callbacks, which are largely an outdated legacy syntax.  But I don't know if that's a priority to @raclim.  The current version of `mongoose` does not even allow passing a callback to `exec()`.  So the type is a lot simpler:
```
    /** Executes the query */
    exec(): Promise<ResultType>;
```
https://github.com/Automattic/mongoose/blob/1f340d4e3a846abe520beb9b2f659b2eb3fe1a35/types/query.d.ts#L213-L214

lindapaiste:
I fixed some of the issues and pushed a commit. ✅

swrno:
Still I am getting some error, is it running for you?
If yes, probably, I have to setup the branch again in my lical system. And we can start working on it. In my opinion, the frontend part will be easy to convert, 

can I convert any small component for testing purposes?

raclim:
Thanks for your work on this! After giving it some thought I don't think implementing TypeScript will be on the roadmap in the near future in favor of addressing some of the other technical debt in this repository first. Once that's done though, I do think this can be revisited around later in the year or early next year. I'm going to add this to a [Board for Saved PRs to Revisit](https://github.com/processing/p5.js-web-editor/projects/11#card-92688730) to look at until we're ready to come back to this!

PR #3022, opened 2024-02-14
Delete unused callback on `isSlugUnique` function

lindapaiste:
Ref #521
This is part of a larger effort to move away from callback-based code.

Quick & easy cleanup.  `projectSchema.methods.isSlugUnique` supports an optional callback `cb` argument.  But the one place in the code where we use this function is already using it as an `async`.  So we can safely delete all logic related to the callback.


raclim (approved):
Thanks for catching this, it's so much easier to read!

PR #2906, opened 2024-01-16
Added folder icons

Anubhav-Vashishtha:
Fixes #2881 

Changes:

image

- I have changed scss files to make options icon more visible initially it was smaller in size

Replace Directional arrows and Added folder icon to the side bar to make user experiences better

I have verified that this pull request:

* [V] has no linting errors (`npm run lint`)
* [V] has no test errors (`npm run test`)
* [V] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* [V] is descriptively named and links to an issue number, i.e. `Fixes #123`

Issue #1987, opened 2021-12-29
Feature Request: Private Option for Sketches


golanlevin:
- New feature request

#### New feature details:

Hi friends, 
I'm writing to re-open issues #91, #1604, and #1832 — a feature request for a "private" option for sketches. Sketches that have been set to Private would not be visible, unless their author was logged into the Editor in their browser.  

Some reasons that an artist might want to set a sketch to Private could be that: 
* It is a work-in-progress, and they're not ready to 'publish' it
* It is a solution for a homework that they're giving to their students
* It's a scrap sketch or piece of test code that they don't want shown in their public gallery
* The project deals with controversial themes, etc.
* (**NEW!!**) The creator is concerned about theft by NFT copyminters

Like it or not, the NFT phenomenon has radically transformed the landscape of generative and algorithmic software art. P5.js is now widely used as a tool for making NFTs. While this has led to some income for our community, the monetization of generative art has also led to a lot of scammers, fake accounts, 'copyminting', and other foul play. Unfortunately, some unscrupulous people are trawling through various Editor accounts, gleaning projects and then selling them as NFTs. The requested feature would allow an artist to work on a sketch away from the prying eyes of NFT thieves. 

I think the logical place to set a sketch to "Private" is in this menu:

<img width="399" alt="Screen Shot 2021-12-29 at 5 14 10 AM" src="https://user-images.githubusercontent.com/290053/147653784-53de976a-67b6-41b6-820f-bc2be9bf4b5b.png">

Thanks for your consideration!
<3 Golan

CarsonKompon:
Bumping in hopes that there is a reconsideration

CarsonKompon:
Bumping in hopes that there is a reconsideration

dhowe:
This is really _really_ needed

vivekbopaliya:
@raclim how are you planning to tackle this?---simple if-else or even post request and complete server actions can be involved for more privacy imo

golanlevin:
Hi friends — if there's any chance this is being considered, I'd be so happy.
One UX implementation suggestion – I'm a fan of how YouTube has three privacy categories: 

- 🟢 **Public**: publicly listed on the creator's page, and viewable to all.
- 🟡 **Unlisted**: not publicly listed on the creator's page, but viewable if the author has shared the URL with you.
- 🔴 **Private**: not publicly listed on the creator's page, and only viewable to the creator when they're logged in. 

Best,
Golan

[...]

PR #3034, opened 2024-02-18
Private Sketch feat.

vivekbopaliya:
Changes:

https://github.com/processing/p5.js-web-editor/assets/122638553/40cf3f6f-15af-43b6-9226-de5783b43ff0

### WORK FLOW: 
- Every sketch is default public, user can change to Private.
- Private sketch will be visible to owner of project only.
- If someone who is not owner tries to access Private project through URL, they will be redirected to /
- Filtering is done on backend.


lindapaiste:
This is awesome.  I can tell that you have put a lot of thought into it and done some very good work here!  I'm sure @raclim will have some thoughts on the UI.  I'm going to see if I can figure out why you're getting errors in the `getProjectsForUser` handler.  I'm not sure what steps we need to take as far as backwards compatibility, as existing projects won't have a `visibility` property.  Possibly it's enough just to set `default: 'Public'` on the Schema?

server/controllers/project.controller/getProjectsForUser.js
```diff
@@ -10,25 +10,44 @@ import { toApi as toApiProjectObject } from '../../domain-objects/Project';
...
+      !currentUser.equals(user._id) ? publicProjectsOnly : projects
+    );
+
     res.json(response);
-  } catch (e) {
-    res.status(500).json({ message: 'Error fetching projects' });
+  } catch (error) {
+    res
+      .status(500)
+      .json({ success: false, message: 'Error fetching projects' });
```
lindapaiste: > I'm going to see if I can figure out why you're getting errors in the `getProjectsForUser` handler. 

The test failure is something really simple.  You added `{ success: false }` to the 500 error response so the `function expectations()` in the test file needs to be updated to include `{ success: false }`.  We are not very consistent with the formatting of error responses and that is something that I would like to address at some point.

server/controllers/project.controller/getProjectsForUser.js
```diff
@@ -10,25 +10,44 @@ import { toApi as toApiProjectObject } from '../../domain-objects/Project';
...
     const projects = await Project.find({ user: user._id })
       .sort('-createdAt')
-      .select('name files id createdAt updatedAt')
+      .select('name files id createdAt updatedAt visibility')
       .exec();
-    const response = mapProjectsToResponse(projects);
+
+    const publicProjectsOnly = projects.filter(
+      (project) => project.visibility === 'Public'
+    );
```
lindapaiste: Ideally we're going to want to do this filtering as part of the Mongoose query, rather than returning all and then filtering the array.

server/controllers/project.controller/getProjectsForUser.js
```diff
@@ -10,25 +10,44 @@ import { toApi as toApiProjectObject } from '../../domain-objects/Project';
...
       res
         .status(404)
         .json({ message: 'User with that username does not exist.' });
       return;
     }
+
     const projects = await Project.find({ user: user._id })
       .sort('-createdAt')
-      .select('name files id createdAt updatedAt')
+      .select('name files id createdAt updatedAt visibility')
```
lindapaiste: Now that I'm looking at this, the way that we have it currently set up is kind of silly because in the `apiGetProjectsForUser` case we could just have `'name id'` here, rather than returning the whole object from the DB and mapping it 🤔

server/controllers/project.controller.js
```diff
@@ -270,3 +270,40 @@ export function downloadProjectAsZip(req, res) {
...
+      {
+        visibility: newVisibility
+      },
+      {
+        new: true
+      }
+    );
+    const updatedProjectWithoutFiles = await Project.findById(
+      updatedProject._id
+    ).select('-files');
```
lindapaiste: I didn't know that you could do a negative like this!  Cool.

server/controllers/collection.controller/updateCollection.js
```diff
@@ -43,7 +43,7 @@ export default function createCollection(req, res) {
         { path: 'owner', select: ['id', 'username'] },
         {
           path: 'items.project',
-          select: ['id', 'name', 'slug'],
+          select: ['id', 'name', 'slug', 'visibility'],
```
lindapaiste, line 46: Annoying that you had to update this same block of code in 5 places.  That's definitely a sign of something that should be refactored. (In a future PR, not here.)

client/modules/IDE/components/SketchList.jsx
```diff
@@ -179,6 +208,56 @@ class SketchListRowBase extends React.Component {
...
           />
         )}
+
+        {this.state.visibleDialogOpen && (
+          <Overlay
+            title={title}
+            closeOverlay={() => this.setState({ visibleDialogOpen: false })}
+          >
+            {/* TODO: these <li> should come from translate.json */}
+            <div className="sketch-visibility">
```
lindapaiste: I'm wondering if we should move the `VisibilityDialog` to its own component.  I'm thinking that we probably want to use this for confirmation when clicking the lock icon on the sketch page.  Otherwise it might be too easy to make something public by mistake?

server/controllers/project.controller/getProjectsForUser.js
lindapaiste: Ok so filtering on `visibility: { $eq: 'Public' }` does not seem to give any results, even with `default: 'Public'`.  But we can use a ["not equals" `$ne`](https://www.mongodb.com/docs/manual/reference/operator/query/ne/) to filter for projects which are not private:  `{ $ne: 'Private' }`.

Something like this:
```
const canViewPrivate = req.user && req.user._id === user._id;

const filter = { user: user._id };
if (!canViewPrivate) {
  filter.visibility = { $ne: 'Private' };
}

const projects = await Project.find(filter)
  .sort('-createdAt')
  .select('name files id createdAt updatedAt visibility')
  .exec();
```

server/models/project.js
```diff
@@ -38,6 +38,10 @@ const projectSchema = new Schema(
     serveSecure: { type: Boolean, default: false },
     files: { type: [fileSchema] },
     _id: { type: String, default: shortid.generate },
+    visibility: {
+      type: String,
+      enum: ['Private', 'Public']
```
lindapaiste: We want to add `{ default: 'Public' }`.  That won't update existing database records, but it does seem to add the property to the objects when they are returned from a query.

server/routes/project.routes.js
```diff
@@ -26,4 +26,6 @@ router.get('/:username/projects', ProjectController.getProjectsForUser);
 
 router.get('/projects/:project_id/zip', ProjectController.downloadProjectAsZip);
 
+router.patch('/project/visibility', ProjectController.changeProjectVisibility);
```
lindapaiste, line 29: Not sure if we need a separate endpoint?  Can we use the "update" endpoint?

vivekbopaliya:
> I'm not sure what steps we need to take as far as backwards compatibility, as existing projects won't have a visibility property. Possibly it's enough just to set default: 'Public' on the Schema?

Yeah, that will work. will do that. However, I have set up the state like this in `Toolbar.jsx`: 

```javascript
const [visibility, setVisibility] = React.useState(
    project.visibility || 'Public'
);
```

So for old sketches, the default value "Undefined" will be replaced by "Public" in the state. This way, the user should be able to change the visibility of old sketches as well.

server/controllers/project.controller/getProjectsForUser.js
vivekbopaliya: Oh, wowwww! Although I know a good practice to set up default formatting for error responses (as well as success responses) rather than changing everywhere independently, we can create a custom class for both types of responses and use them all over the backend, overwriting values accordingly for errors. this way it will be a consistent responses?!

```
class ApiError  {
    constructor(
        statusCode,
        message= "Something went wrong",
        errors = [],
        stack = ""
    ){
        super(message)
        this.statusCode = statusCode
        this.data = null
        this.message = message
        this.success = false;
        this.errors = errors
    }
}
```

server/controllers/project.controller/getProjectsForUser.js
vivekbopaliya: This is super great, man! The function got reduced by at least 10 lines and it took away the burden from the database as we now don't need to do extra queries for filtering.

server/controllers/project.controller/getProjectsForUser.js
vivekbopaliya: I believe we're using each property within `SketchList.jsx` to render the data for each row? 🤔

client/modules/IDE/components/SketchList.jsx
vivekbopaliya: Yes, you're right. I've also considered that. However, since this is part of a complete private sketch feature, I believe there will be quite a discussion about the UI. Later on, I will move it to its own component once the UI is finalized. I spent more than an hour trying to find two symmetric good icons and align it perfectly inside the row of `SketchList.jsx` (yet it seems like I've left some missing pieces). I didn't realize how messy it would be to manage SVG directly. Perhaps we might want to use React-Icons?

server/routes/project.routes.js
vivekbopaliya: I guess it will be better to have its separate endpoint. I'm not sure where we want to put that code without making another function messy. There's already a lot going on inside the `updateProject` function.

client/modules/IDE/components/SketchList.jsx
vivekbopaliya: Additionally, I don't know if you have noticed this, but the `<Overlay />` component overwrites any toast message.

lindapaiste: > Additionally, I don't know if you have noticed this, but the `<Overlay />` component overwrites any toast message.

Hmm.  I have not noticed it.  That's concerning.  Feel free to open a new issue explaining in more detail.

lindapaiste: > I spent more than an hour trying to find two symmetric good icons and align it perfectly inside the row of `SketchList.jsx` (yet it seems like I've left some missing pieces). I didn't realize how messy it would be to manage SVG directly. Perhaps we might want to use React-Icons?

Yessss dealing with the SVG files is really burdensome and unnecessary.  I would love to use `react-icons`!  See the discussion here: https://github.com/processing/p5.js-web-editor/pull/2906

server/controllers/project.controller/getProjectsForUser.js
lindapaiste: Yes.  I'm thinking to myself about how to manage the two versions of the function.  Currently the difference comes from the `mapProjectsToResponse` argument.  But now my gears are turning and I'm realizing that instead we could have a `fields` argument where we pass either `'name id'` or `'name files id createdAt updatedAt visibility'` as the argument.  Just something I'm thinking about.  FYI before I worked on this file we had two completely separate functions which were *almost* identical but I loathe repetitive code so I made it a higher order function #2320.

server/controllers/project.controller/getProjectsForUser.js
lindapaiste: 🤗 Also this sort of setup should work a lot better if we ever get around to implementing pagination.

vivekbopaliya:
@lindapaiste  @raclim  still waiting for UI to be approved so that i can go ahead

raclim:
Thanks so much for working on this and I'm sorry for the delay in getting to this! 

Honestly, this looks really awesome and I don't think I have much else to add so far on the UI! I know there were some [previous designs done for a potential Privacy feature](https://www.figma.com/file/IAqdYZhfODO0FvDmvL2pI2/p5js?type=design&node-id=88-205&mode=design&t=0a2BDjr1I9eBwIQI-0). It's a bit late, but you can probably still choose to incorporate them if you'd like, but I think what's currently here is great as well. 

I think one potential next step for this after some further development is having it tested out with a small group of users who can probably give further feedback from there on the design/functionality as well. I can also try to start tagging and sharing it with some folks if it makes more sense to do it sooner rather than later down the line!

vivekbopaliya:
> I know there were some [previous designs done for a potential Privacy feature](https://www.figma.com/file/IAqdYZhfODO0FvDmvL2pI2/p5js?type=design&node-id=88-205&mode=design&t=0a2BDjr1I9eBwIQI-0).

This whole look is quite so cool and refreshing actually. I'm excited about adding it to my design. It's got a premium vibe that I think will work really well. However, it seems like it will require a significant amount of effort on its own. I'm currently tied up with some other tasks for the time being, but I'll try to dedicate some time to it within the next month and raise a PR. In the meantime, It would be best to focus on its functionality and gathering feedback related to it. it's better to move forward by sharing it with some individuals to gather feedback on its functionality.

vivekbopaliya:
> It's a bit late, but you can probably still choose to incorporate them if you'd like, but I think what's currently here is great as well.

@raclim Just a quick update: I've created a new [PR ](https://github.com/processing/p5.js-web-editor/pull/3108)with changes inspired by the design. let me know your thoughts!

raclim:
@vivekbopaliya Sorry that it took a while, just want to update that I created a [shareable environment](https://app-privating-sketches.release.editor.p5js.org/) for this PR! Please feel free to share this around and get some feedback on it! 

I feel like it looks pretty solid so far though, and super excited to see this out soon!

vivekbopaliya:
@raclim finally it's hereeee and it is looking GOOD, thank you very much!

vivekbopaliya:
> @vivekbopaliya Sorry that it took a while, just want to update that I created a [shareable environment](https://app-privating-sketches.release.editor.p5js.org/) for this PR! Please feel free to share this around and get some feedback on it!
> 
> I feel like it looks pretty solid so far though, and super excited to see this out soon!

Finally, it is here! Thank you very much. Let's see how this works out. If it does, we might want to design an implementation for mobile view as well.

raclim:
Sounds good, I'll also do some further testing on my end as well!

Felixpaulsen:
Nice work! I hope we'll see this public soon as it will also fix issues with private api keys being publicly exposed.
[](https://github.com/processing/p5.js-web-editor/issues/1094)

vivekbopaliya:
@raclim it's been a quite time. What's status of this feature?

raclim:
Hi @vivekbopaliya, thanks so much for checking back in on this. I genuinely appreciate your patience and I'm really sorry about not following through with an update sooner. I got caught up with a few tasks that came up and wasn't able to stay on top of this, but will be more responsive here moving forward.

To give more context on what's happening here, over the past few months we've been handling a wave of privacy-related requests from schools. We decided to prioritize receiving legal consultation before moving forward with any larger features, which included this one, to ensure we were adhering to best practices and regulations. It's taken longer than expected, but we're aiming to wrap up the consultation by the end of this month. 

Once that's done, I think it makes sense to make this feature a priority. If you're still on board, we could aim to roll this out by the end of February or mid-March? Thank you so much again for your patience and for sticking around with this!

vivekbopaliya:
No worries at all! Thanks for the update. happy to help whenever I can.

raclim:
Hi @vivekbopaliya! I'm sorry that it took a little longer, but we were finally able to finish our legal consultation, and confirmed that we're able to move forward with this feature as planned. Although it's been a while, would you still be open to completing it?

raclim:
Hi @vivekbopaliya! Just wanted to check in again if you'd be interested in furthering development on this. If not, I'd love to continue the development on this if it's alright with you.

vivekbopaliya:
Hey @raclim thank you for getting back to me. Yes I'm still on this, and I'll try to finish the pending work in a few days.

Anything additional changes you need me to include?

raclim:
Thanks so much again for your patience and for checking back in on this after quite some time! Just wanted to note that there's no rush on trying to wrap it up, I just wanted to check in and see if you'd still be down!

I think the only change that seemed to be left was the one pointed out in this comment (https://github.com/processing/p5.js-web-editor/issues/1987#issuecomment-2166794463). I was able to recreate it by: 

1. Saving a new sketch. 
Screenshot 2025-04-10 at 4 36 53 PM

2. Making the sketch private by toggling the "private" checkbox. 
Screenshot 2025-04-10 at 4 36 59 PM

3. Setting the sketch back to public by toggling the setting in **My Sketches**. 
Screenshot 2025-04-10 at 4 37 12 PM

4. Navigate back to the sketch, make some changes, and then attempt to save. On save, I get the following error. 
Screenshot 2025-04-10 at 4 37 29 PM



I think besides this, this feature might overall be good to try to release for now! I think I would just double check that private sketches are only visible to the project owners and other users who attempt to do so are redirected to `/`. I'm also just wondering, would this also guarantee that it's not searchable by the browser as well?

vivekbopaliya:
hey @raclim, I've pushed new changes, also incorporated the old bug mentioned here: https://github.com/processing/p5.js-web-editor/pull/3108#issuecomment-2137933375. do check out!

to clarify, sketches are only visible to their project owners. If another user tries to access a private sketch, they’ll be redirected to /. However, it's worth noting that private sketches may still be discoverable via browser search, but surely any attempt to open them will result in a redirect to /.

raclim:
Just wanted to update that I recreated the[ testing environment for this PR](https://app-privating-sketches.release.editor.p5js.org/)! I'll look through it over the next few days, as well as with the PF team, and should have an update by the middle of next week!

SableRaf:
image

I’d suggest renaming the column to "Visibility" instead of "Make Private". That label reads more clearly as a status rather than an action, and it would also align better with the other column headers. 

I’d also recommend using a dropdown with options like "Private" and "Public" instead of a toggle (even though it looks really nice). This would make the options more legible and give us more flexibility in the future if we want to support additional visibility states like "Unlisted."

.env.example
```diff
@@ -19,7 +19,7 @@ MAILGUN_KEY=<your-mailgun-api-key>
 ML5_LIBRARY_USERNAME=ml5
 ML5_LIBRARY_EMAIL=examples@ml5js.org
 ML5_LIBRARY_PASS=helloml5
-MONGO_URL=mongodb://localhost:27017/p5js-web-editor
+MONGO_URL=mongodb://localhost:27017/local
```
raclim: Would you be able to delve into further detail about the change here? If it's not pivotal to this feature, I think we might not need to add this here!

client/modules/User/components/Collection.jsx
```diff
@@ -95,6 +177,7 @@ const Collection = ({ collectionId, username }) => {
             ) : (
               <ArrowDownIcon
                 role="img"
+                IST
```
raclim: Could you clarify what `IST` is?

client/modules/User/components/Collection.jsx
```diff
@@ -11,7 +12,88 @@ import Loader from '../../App/components/loader';
 import ArrowUpIcon from '../../../images/sort-arrow-up.svg';
 import ArrowDownIcon from '../../../images/sort-arrow-down.svg';
 import CollectionMetadata from './CollectionMetadata';
-import CollectionItemRow from './CollectionItemRow';
+import dates from '../../../utils/formatDate';
+import RemoveIcon from '../../../images/close.svg';
+
+const CollectionItemRow = ({ item, isOwner, collection, user }) => {
```
raclim: In the original file, it seems like `CollectionItemRow` was being imported. Would it be possible to add these updates to `CollectionItemRow.jsx` instead?

client/modules/IDE/pages/IDEView.jsx
```diff
@@ -105,7 +105,7 @@ const IDEView = () => {
   const [sidebarSize, setSidebarSize] = useState(160);
   const [isOverlayVisible, setIsOverlayVisible] = useState(false);
   const [MaxSize, setMaxSize] = useState(window.innerWidth);
-
+  const history = useHistory();
```
raclim: Let's address the lint warning here!

client/modules/IDE/components/SketchListRowBase.jsx
```diff
@@ -162,7 +226,7 @@ SketchListRowBase.defaultProps = {
 
 function mapDispatchToPropsSketchListRow(dispatch) {
   return bindActionCreators(
-    Object.assign({}, ProjectActions, IdeActions), // Binding both ProjectActions and IdeActions
+    Object.assign({}, ProjectActions, IdeActions),
```
raclim: Please add the comment back in!

client/modules/IDE/components/SketchListRowBase.jsx
```diff
@@ -75,12 +77,67 @@ const SketchListRowBase = ({
   };
 
   const handleSketchDuplicate = () => cloneProject(sketch);
+
+  // const handleSketchShare = () => {
```
raclim, line 82: Is this isn't being utilized, let's remove the commented out functions!

client/modules/IDE/components/SketchListRowBase.jsx
```diff
@@ -75,12 +77,67 @@ const SketchListRowBase = ({
...
+        <svg
+          width="8"
+          height="11"
+          viewBox="0 0 8 11"
+          fill="none"
+          xmlns="http://www.w3.org/2000/svg"
+          className="lock"
+        >
+          <path
+            d="M8 5.68627V10.0784C8 10.5882 7.54067 11 7.00478 11H0.995215C0.440191 11 0 10.5686 0 10.0784V5.68627C0 5.17647 0.440191 4.7451 0.995215 4.7451C1.09035 4.7451 1.16746 4.66798 1.16746 4.57285V2.90196C1.16746 1.29412 2.43062 0 3.98086 0C5.55024 0 6.8134 1.29412 6.8134 2.90196V4.55371C6.8134 4.65941 6.89908 4.7451 7.00478 4.7451C7.54067 4.7451 8 5.15686 8 5.68627ZM2.33716 3.11732C2.34653 4.01904 3.08017 4.7451 3.98194 4.7451C4.89037 4.7451 5.62679 4.00867 5.62679 3.10024V2.90196C5.62679 1.96078 4.89952 1.21569 3.98086 1.21569C3.10048 1.21569 2.33493 1.96078 2.33493 2.90196L2.33716 3.11732Z"
```
raclim: Let's address the lint warning here as well! I think we can add the icon directly into `client/images` and import the icon image. There should be examples of this in other components within the repo!

client/modules/IDE/components/SketchListRowBase.jsx
raclim: Ohh, was this added in case we want to add it to the menu dropdown as well?

client/modules/IDE/components/SketchList.jsx
```diff
@@ -60,7 +61,9 @@ const SketchList = ({
   const renderEmptyTable = () => {
     if (!isLoading() && sketches.length === 0) {
       return (
-        <p className="sketches-table__empty">{t('SketchList.NoSketches')}</p>
+        <p className="sketches-table curative__empty">
```
raclim: Sorry if I might've missed this elsewhere, but could you identify where `curative__empty` is from?

client/modules/IDE/components/Header/Toolbar.jsx
```diff
@@ -101,9 +118,22 @@ const Toolbar = (props) => {
       <div className="toolbar__project-name-container">
         <ProjectName />
         {(() => {
-          if (project.owner) {
+          if (project?.owner && userIsOwner) {
```
raclim: Similar to the comment in `MobileNav.jsx`, I think maybe we can reformat the conditional rendering logic that's happening here!

client/modules/IDE/components/Header/MobileNav.jsx
```diff
@@ -237,21 +246,53 @@ const MobileNav = () => {
...
       </LogoContainer>
       <Title>
-        <h1>{title === project.name ? <ProjectName /> : title}</h1>
-        {project?.owner && title === project.name && (
-          <Link to={`/${project.owner.username}/sketches`}>
-            by {project?.owner?.username}
-          </Link>
-        )}
+        <h1>{title === project?.name ? <ProjectName /> : title}</h1>
+        {(() => {
```
raclim: Would it be possible to handle the conditional rendering logic handled outside of JSX for readability? 

For example, you could define a variable above the `return` block like: 
```
const showPrivacyToggle = project?.owner && title === project.name && userIsOwner;
``` 

Then in the JSX: 

```
{showPrivacyToggle && (
  <main className="toolbar__makeprivate">
    ...
  </main>
)}
```

Let me know what you think!

raclim (changes requested):
Thanks so much for the continued wrap-up on this! 

I think this looks pretty solid so far and really excited to see it near it's completion! I added in a few notes—let me know if you have any further questions or thoughts on them! Otherwise, I'll be following up soon with some of the testing feedback from other users as well.

raclim:
> I’d also recommend using a dropdown with options like "Private" and "Public" instead of a toggle (even though it looks really nice). This would make the options more legible and give us more flexibility in the future if we want to support additional visibility states like "Unlisted."

Thanks so much @SableRaf for the suggestion! Would it look something like this? I can do some quick A/B Testing with Amy to see what folks think!

**Current View with Toggle**: 
Private_Feature_A

**Proposed View with Dropdown**:
Private_Feature_B

Private_Feature_B_2

client/modules/IDE/components/SketchListRowBase.jsx
vivekbopaliya: Yes, that’s right — it was added earlier

.env.example
vivekbopaliya: Correct! i will change it back -- it was probably for local testing!

client/modules/User/components/Collection.jsx
vivekbopaliya: Oops! a typo

client/modules/IDE/components/Header/MobileNav.jsx
vivekbopaliya: Great point! Moving the conditional logic out of JSX would definitely improve **readability**.  Thanks for the suggestion!

client/modules/IDE/components/SketchList.jsx
vivekbopaliya: Probably, we won't need it! My bad

client/modules/IDE/components/SketchList.jsx
```diff
@@ -59,9 +60,7 @@ const SketchList = ({
 
   const renderEmptyTable = () => {
     if (!isLoading() && sketches.length === 0) {
-      return (
-        <p className="sketches-table__empty">{t('SketchList.NoSketches')}</p>
-      );
+      return <p className="sketches-table">{t('SketchList.NoSketches')}</p>;
```
raclim: I think the className here should be `sketches-table__empty`!

raclim (changes requested):
Thanks so much for updating the requested changes, this looks great so far! 

I think this was mentioned in another comment in the original issue, but when I tried playing around with the feature I ran into an error with saving the project once the "Private" flag is on. 

Screenshot 2025-06-30 at 4 19 11 PM


I produced this error by: 
1. Opening a new sketch. 
2. Saving the sketch. 
3. Toggling the "Private" checkbox in the toolbar above the sketch. 
4. Adding in a file change to `sketch.js`. 
5. Saving my new changes, which triggers the error pictured above. 

I'll try to look into what might be going on here as well, but wanted to update that this has been happening for me! Please let me know if you have any difficulties with reproducing it!

vivekbopaliya:
@raclim Sure, I’ve pushed the fix. The issue was happening because the server was updating the updatedAt field, but the frontend (Redux store) didn’t have that updated value. So when we tried to update the sketch, the server threw a 409 Conflict error since it thought the data was outdated.

Also, I wanted to quickly bring up something for consistency:

Currently, on the collection page, we show non-owners the message: "Sketch is private" for private sketches.
But on the sketch list, we don’t show private sketches at all to non-owners.

Should we align this behavior?
Either:

hide private sketches everywhere for non-owners, 
or
show them with a "Sketch is private" placeholder consistently.

Let me know your thoughts!
Screenshot from 2025-07-05 13-48-42

SableRaf:
> Thanks so much @SableRaf for the suggestion! Would it look something like this? I can do some quick A/B Testing with Amy to see what folks think!

Thanks! The dropdown looks good, but I think it should also indicate the current state (whether the sketch is Public or Private) without needing to click it. That way it’s immediately clear what the current visibility setting is.

Here's how YT does it for example:

image

raclim:
Hi @vivekbopaliya! 

I wanted to check-in to see if it felt okay time-wise to transition from the current design of the privacy toggle to a dropdown one, similar to the one on Youtube proposed by @SableRaf. If it does feel okay, would it also be possible to also have the toolbar utilize the same dropdown design as well, replacing the checkbox (pictured below)? 

Screenshot 2025-07-15 at 4 39 34 PM

Let me know your thoughts!

vivekbopaliya:
> Hi @vivekbopaliya!
> 
> I wanted to check-in to see if it felt okay time-wise to transition from the current design of the privacy toggle to a dropdown one, similar to the one on Youtube proposed by @SableRaf. If it does feel okay, would it also be possible to also have the toolbar utilize the same dropdown design as well, replacing the checkbox (pictured below)?
> 
> Screenshot 2025-07-15 at 4 39 34 PM
> Let me know your thoughts!

Sure! I'm almost done with it. just wrapping up a few things. I'll push the changes in the next couple of days.

raclim:
Thanks so much @vivekbopaliya for your updates on this! 

Just wanted to note that after taking a look, I think the PR overall feels functionally good to me! We just merged in a few changes to update the `mongoose` and `node` version which seems like it's causing a minor merge conflict, but overall I think this feels great!

raclim (approved):
Thank you so much again for your work on this, this is really awesome and it's so amazing to see how this feature shaped out! :) 

As we discussed, I'm going to merge this in a temporary feature branch ([`feature-develop`](https://github.com/processing/p5.js-web-editor/tree/feature-develop)) until the planned release date on August 25th. I'll also be adding in some minor tweaks to the Toolbar UI within this branch—feel free to keep track of the changes here and share thoughts if anything comes up!

PR #3045, opened 2024-02-22
Udpated : Passport handles expected errors as exceptions

### LINKED ISSUE #3044: Passport handles expected errors as exceptions
### Increasing Access

We should follow best practices for error handling.

Our sign in with Google and Github features are not handling errors in the proper way.  We are treating login failures due to invalid credentials as if they were server exceptions.  These error messages should be in the third argument of the `done` callback instead of the first argument.

This line is incorrect:
https://github.com/processing/p5.js-web-editor/blob/a9e518c48ce657a7713cd73b99655946cd3b1881/server/config/passport.js#L126-L128

This line is correct:
https://github.com/processing/p5.js-web-editor/blob/a9e518c48ce657a7713cd73b99655946cd3b1881/server/config/passport.js#L53

Explanation from [passport documentation](https://www.passportjs.org/concepts/authentication/strategies/): (read the last paragraph)

> A `verify` function yields under one of three conditions: success, failure, or an error.
>
> If the `verify` function finds a user to which the credential belongs, and that credential is valid, it calls the callback with the authenticating user:
> ```return cb(null, user);```
>
> If the credential does not belong to a known user, or is not valid, the `verify` function calls the callback with `false` to indicate an authentication failure:
> ```return cb(null, false);```
>
> If an error occurs, such as the database not being available, the callback is called with an error, in idiomatic Node.js style:
> ```return cb(err);```
>
> It is important to distinguish between the two failure cases that can occur. Authentication failures are expected conditions, in which the server is operating normally, even though invalid credentials are being received from the user (or a malicious adversary attempting to authenticate as the user). Only when the server is operating abnormally should `err` be set, to indicate an internal error.

SjxSubham:
Fixes #3044 

Changes: Line 126 inside done()

* [x] is descriptively named and links to an issue number, i.e. `Fixes #3044 `

### DIFF
--- a/server/config/passport.js
+++ b/server/config/passport.js
@@ -123,9 +123,7 @@ passport.use(
       User.findOne({ github: profile.id }, (findByGithubErr, existingUser) => {
         if (existingUser) {
           if (req.user && req.user.email !== existingUser.email) {
-            done(
-              new Error('GitHub account is already linked to another account.')
-            );
+           done(null, false, { msg: 'Invalid email or password.' }); 
             return;
           } else if (existingUser.banned) {
             done(new Error(accountSuspensionMessage));


lindapaiste:
This is good by my intention was to change all similar lines too, not just the one line that I pointed out.  I'm closing this in favor of #3049

PR #3049, opened 2024-02-23
Updated passport handles expected errors as exceptions

Kuldeep246:
Added error messages in the third argument of the done callback instead of the first argument in Google and Github sign in features

Fixes #issue-number

Changes:

### DIFF
--- a/server/config/passport.js
+++ b/server/config/passport.js
@@ -13,6 +13,9 @@ import User from '../models/user';
 const accountSuspensionMessage =
   'Account has been suspended. Please contact privacy@p5js.org if you believe this is an error.';
 
+const accountLinkingErrorMessage =
+  'Account is already linked to another account.';
+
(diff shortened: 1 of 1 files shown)


lindapaiste (changes requested):
This has some formatting errors which can be fixed by running the `lint` command.

server/config/passport.js
```diff
@@ -13,6 +13,9 @@ import User from '../models/user';
 const accountSuspensionMessage =
   'Account has been suspended. Please contact privacy@p5js.org if you believe this is an error.';
 
+const accountLinkingErrorMessage =
+  'Account is already linked to another account.';
+
```
lindapaiste, line 18: Thanks for your work on this.  I'm going to close this PR and go with #3050 instead because I don't like this particular change.  IMO we want to keep the name of the service in there.

PR #3054, opened 2024-02-26
Fixed/ Logged out error modal.

### LINKED ISSUE #3053: Logged out error modal pops up when downloading another user's sketch
### p5.js version

_No response_

None

### Web browser and version

_No response_

### Actual Behavior

image

### Expected Behavior

I should be able to download p5's sketch without any error.

The sketch actually does export successfully, using the saved version of the project.

We are getting this message because we make a PUT request to the API before exporting a sketch.  That API request will fail if the user is not the owner of the sketch.  The easiest solution would be to only make that API call if there is a logged in user and that user is the owner of the sketch.

The relevant code is here:
https://github.com/processing/p5.js-web-editor/blob/64c36cae65541783e848eb5c0e11eb255ec6c940/client/modules/IDE/hooks/useSketchActions.js#L42-L45

There is some nuance here with regards to how we would handle local/unsaved changes.  Right now those changes would not be applied if it is not your sketch.  This is part of a larger discussion happening in #2904.

### Steps:
1. Make sure that you are logged out
2. Go to a sketch which is not your own, such as https://editor.p5js.org/p5/sketches/Data:_True_and_False
3. Click the "File" menu
4. Click "Download"

letscodedanish:
### Changes:
With this update, the downloadSketch function now ensures that the project is only exported when the conditions regarding the logged-in user and ownership of the project are met. This helps prevent unnecessary API requests and improves the user experience by handling the download functionality more efficiently.

* [x] is descriptively named and links to an issue number, i.e. `Fixes #[issue-number]`

### DIFF
--- a/client/modules/IDE/hooks/useSketchActions.js
+++ b/client/modules/IDE/hooks/useSketchActions.js
@@ -41,7 +41,10 @@ const useSketchActions = () => {
 
   function downloadSketch() {
     dispatch(autosaveProject());
-    exportProjectAsZip(project.id);
+    // Only export the project if there is a logged-in user and they are the owner
+    if (authenticated && project.owner === authenticated.userId) {
+      exportProjectAsZip(project.id);
+    }
   }
 
   function shareSketch() {
--- a/client/modules/IDE/hooks/useSketchActions.js
+++ b/client/modules/IDE/hooks/useSketchActions.js
@@ -40,8 +40,10 @@ const useSketchActions = () => {
   }
 
   function downloadSketch() {
-    dispatch(autosaveProject());
     if (authenticated && project.owner === authenticated.userId) {
+      dispatch(autosaveProject());
+      exportProjectAsZip(project.id);
+    } else {
       exportProjectAsZip(project.id);
     }
   }
--- a/client/modules/IDE/hooks/useSketchActions.js
+++ b/client/modules/IDE/hooks/useSketchActions.js
@@ -41,7 +41,6 @@ const useSketchActions = () => {
 
   function downloadSketch() {
     dispatch(autosaveProject());
-    // Only export the project if there is a logged-in user and they are the owner
     if (authenticated && project.owner === authenticated.userId) {
       exportProjectAsZip(project.id);
     }
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "LGPL-2.1",
       "dependencies": {
         "@auth0/s3": "^1.0.0",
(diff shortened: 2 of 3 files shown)


client/modules/IDE/hooks/useSketchActions.js
```diff
@@ -41,7 +41,9 @@ const useSketchActions = () => {
 
   function downloadSketch() {
     dispatch(autosaveProject());
-    exportProjectAsZip(project.id);
+    if (authenticated && project.owner === authenticated.userId) {
+      exportProjectAsZip(project.id);
+    }
```
lindapaiste: I think you've got the `if` checking a little bit backwards here.  We want to allow downloading of other people's projects.  But we do not want to call save functions on other people's projects.

Should be more like
```
 if (authenticated && project.owner === authenticated.userId) {
  dispatch(autosaveProject());
 }   
 exportProjectAsZip(project.id);
```

lindapaiste (changes requested):

client/modules/IDE/hooks/useSketchActions.js
letscodedanish: @lindapaiste ,I've revised the downloadSketch function to ensure that we only trigger the autosave functionality if the user is authenticated and they are the owner of the sketch. Here's the updated implementation:
 
 function downloadSketch() {
    if (authenticated && project.owner === authenticated.userId) {
        dispatch(autosaveProject());
    }
    exportProjectAsZip(project.id);
}

letscodedanish: @lindapaiste , And I think this could also be done , correct me if I'm wrong

if (authenticated && project.owner === authenticated.userId) {
  dispatch(autosaveProject());
  exportProjectAsZip(project.id);
} else {
  exportProjectAsZip(project.id);
}

This code block first checks if the user is authenticated and if the current user is the owner of the project. If both conditions are met, it triggers the autosaveProject function before exporting the project as a zip file. If the conditions are not met, it directly exports the project as a zip file without triggering the autosave function.

lindapaiste: You should run your code locally and make sure that it works before submitting a PR.  `authenticated.userId` will not work because `authenticated` is a `boolean` value -- it is not the user object.  `project.owner` is not the id -- it's the entire user object for the owner.  We need a different conditional check here.  Probably `authenticated && canEditProjectName` would do it?  But we need to try it out and check the various situations, like if this is a new sketch that has not been saved before.

PR #3127, opened 2024-05-17
refactor user reducers and actions using redux toolkit

PiyushChandra17:
Associated issue: #2042 

Changes:
+ Deleted all the corresponding `constants`
+ Used `createSlice` to rewrite user reducers
+ Exported the automatically generated action creator functions
+ Deleted the existing action creator functions


lindapaiste (changes requested):
Lots of thoughts here:

You need to be very cognizant of what the payloads are.   The existing code has `authenticateUser(response.data)` and `export function authenticateUser(user)` so we can infer that `response.data` is the user object.  That means your `authUser` reducer isn't right because you are expecting `user` to be a property of the `action.payload` when in fact the entire `action.payload` is the `user`.

You don't need to do `...state` stuff when using `createSlice`.  Just update the property that you want to update and don't `return` anything.  But I think you get the point by now if you see this after my other reviews.

This is a place where [`createAsyncThunk`](https://redux-toolkit.js.org/api/createAsyncThunk) would be very helpful because we currently have actions for the various stages of the process (verify/verified/invalid).  We would call `createAsyncThunk` once and that would create a `verifyEmail` action.  In the reducer we would respond independently (using `extraReducers`) to `verifyEmail.pending`, `verifyEmail.fulfilled`, and `verifyEmail.rejected`.

There are things in the existing code that are silly and/or unnecessary. The reducer always sets the status to `'checking'` when encountering the `EMAIL_VERIFICATION_VERIFY` action type.  So we don't need to include `state: 'checking'` in the action.  Likewise, `EMAIL_VERIFICATION_VERIFIED` and `EMAIL_VERIFICATION_INVALID` both include a `message` that is not used, though I can see how we might want to use that for a toast or something in the future.

Every property other than `authenticated` is optional in the current code.  The initial state right now is just `{ authenticated: false }`.  I do like that you include all of the properties in your version of the `initialState` though because it makes the structure much clearer.  But in that case I would change your `unauthUser` and `authError` actions to return ` => initialState` instead of `=> ({ authenticated: false })` so that we keep those properties present.  I did check the codebase to make sure that changing `emailVerificationTokenState` from not present (`undefined`) to `null` won't break anything and that seems fine.

PR #3130, opened 2024-05-22
fix semantic HTML

### LINKED ISSUE #2967: Semantic HTML issues in header section

PiyushChandra17:
Fixes #2967 

Changes:


raclim (approved):
This overall looks good to me! Thanks for your work on this!

PiyushChandra17:
> This overall looks good to me! Thanks for your work on this!

On to the next one

PR #2993, opened 2024-02-05
Removed redundant redirects between client and server

Mubashirshariq:
Fixes #2992 

Changes:
Server Side:
Made changes on the server: Redirected "/sketches" and "/assets" to URLs with the username. Also, redirected "/reset-password" and "/reset-password/:reset_password_token" to "/account" if the user is logged in.
client :
Made changes on the client: Removed all Higher-Order Component (HOC) auth checks from routes.jsx, deleted createRedirectWithUsername component, and removed auth.js file.

I have verified that this pull request:

* [ ] has no linting errors (`npm run lint`)
* [ ] has no test errors (`npm run test`)
* [ ] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* [ ] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #3148, opened 2024-06-06
Fix redundant redirects between client and server

### LINKED ISSUE #2992: Evaluate redundant redirects between client and server
### Increasing Access

It is confusing to work with code where redirections can be be initiated from multiple places.

Ref: https://github.com/processing/p5.js-web-editor/pull/2987#issuecomment-1925915203

We should look through the https://github.com/processing/p5.js-web-editor/blob/develop/client/routes.jsx file and see which redirects are not actually needed because they are already handled by the server in https://github.com/processing/p5.js-web-editor/blob/develop/server/routes/server.routes.js.

Specifically, we should check over our usages of `userIsNotAuthenticated`, `userIsAuthenticated`, `userIsAuthorized`, and `createRedirectWithUsername`.  I have a hunch that we can actually delete all four of those functions.  Let's make a list of all routes which use some form of redirection and whether each route's redirect logic duplicates what is already implemented on the server.

PiyushChandra17:
Fixes #2992 

Changes to the client side:
+ Removed all auth checks ie. `userIsAuthenticated`, `userIsNotAuthenticated`,  `userIsAuthorized` from `routes.jsx` file
+ Deleted `utils/auth.js` file
+ Deleted `components/createRedirectWithUsername` file

Changes to the server side:
+ redirect `'/sketches'` and `'/assets'` to the url with the username
+ if the user is logged in redirect `'/reset-password'` and `'reset-password/:reset_password_token'` to `'/account'`


raclim:
Thanks for your work on this! This issue ultimately got addressed by an earlier PR (https://github.com/processing/p5.js-web-editor/pull/2993). I'm sorry that we couldn't get this one in, but will continue to check out your other PRs!

PR #3149, opened 2024-06-07
Fix searchbox fit overlays on mobile

### LINKED ISSUE #2369: Search box doesn't fit in overlays on mobile
### Increasing Access

It will look a lot better on mobile devices.

image

@dewanshDT I came across a place where we could really use a `<MediaQuery>` in the JS.  It should be quick if you want to do it.   In the `Overlay` component we sometimes pass in a search box through an `actions` prop.  It doesn't fit on small screens. We should make it so that the `actions` is only rendered in its current position on large screens.  Then on small screens we would place the `{actions}` below the header, between lines 91 `</header>` and 92 `{children}`.

https://github.com/processing/p5.js-web-editor/blob/b5ecd605e8dab40e13d3df0d12491e0388510682/client/modules/App/components/Overlay.jsx#L79-L93

Just change that one part of it.  Making the tables fit will be fixed by #2368 and converting `Overlay` to a function component will be fixed by #2309.

It may need a wrapper div with some margin or padding, which we would render only if `actions` is present.  But that's less important.  You could do it or not bother.  It will be a huge improvement just to make everything fit!

PiyushChandra17:
Fixes #2369 

Changes:
+ Added `MediaQuery` component from 'react-responsive'


raclim:
Thanks for your work on this! Would you be able to provide screenshots showcasing your changes? I think I'm personally having some difficulty recreating the issue! 

Screenshot 2024-06-14 at 9 30 23 AM

PiyushChandra17:
@raclim @lindapaiste I think on the latest `develop` this issue is not fully reproducible, not sure though. Perhaps searchbox partially overlays on small screens but adding `padding-left` does look good on mobile. Here's screenshots, 

before: 
before

After:
after

raclim (approved):
After testing this out I think this looks okay to me so far! Any other needed changes can be implemented down the line. Thanks again for your work on this!

PiyushChandra17:
> After testing this out I think this looks okay to me so far! Any other needed changes can be implemented down the line. Thanks again for your work on this!

@raclim Cool :)

PR #3152, opened 2024-06-08
Added more robust way to handle error

### LINKED ISSUE #3023: `deleteProject` handler doesn't wait for S3 deletion to complete
### Increasing Access

It seems like we could end up with files left behind on S3, which use up resources.

The `deleteProject` handler fires off an asynchronous function `deleteObjectsFromS3` (via `deleteFilesFromS3`) but it does not wait for the S3 deletion to continue before sending an API response.

I'm not sure what actually happens here -- do those files continue to get deleted in the background or not?  I'm concerned that we could wind up with "leftover" files on S3 that are no longer linked to any project.

There is no waiting here:
https://github.com/processing/p5.js-web-editor/blob/6cac275429c3911d87afdcdbbdbf185092aeac29/server/controllers/project.controller/deleteProject.js#L61-L72

I noticed this thanks to a warning in my IDE 
image

PiyushChandra17:
Fixes #3023 

Changes:
+ Added `sendProjectNotFound` function
+ Added more robust way to handle error in `catch` block if the promise is not resolved


raclim (approved):
Thanks for catching and fixing this, this is great!

PR #3161, opened 2024-06-14
FIX:Inputboxes are cut off on Login and Signup forms on smaller screens

### LINKED ISSUE #3158: Signup & login pages are cropped on both side
iOS

### Web browser and version
Chrome 125.0.6422.80

### Actual Behavior

Input fields are being cropped on both sides
| Login page | Signup page |
|--------|--------|
|  | |

### Expected Behavior
login

### Steps to reproduce
Go to signup or login page from mobile phone

rushikeshg25:
Fixes [#3158](https://github.com/processing/p5.js-web-editor/issues/3158)

Changes:

https://github.com/processing/p5.js-web-editor/assets/45139653/e9ccfdd0-e1bc-4270-bf47-0316283a6fe7

Added some CSS styles in media queries to target smaller screens

### DIFF
--- a/client/styles/components/_forms.scss
+++ b/client/styles/components/_forms.scss
@@ -71,6 +71,11 @@
     max-width: 100%;
     width: 100%;
   }
+
+  @media (max-width: 360px) {
+    min-width: 100%;
+    width: 100%;
+  }
 }
 
 .form__eye__icon {
@@ -79,6 +84,10 @@
   top: 7px;
   right: -355px;
   margin-left: -40px;
+
+  @media (max-width: 360px) {
+    right: -200px;
+  }
 }
 
 


client/styles/components/_forms.scss
```diff
@@ -79,6 +84,10 @@
   top: 7px;
   right: -355px;
```
lindapaiste, line 85: My comment is related moreso to the existing code than to your PR.  This sort of positioning based on hard-coded pixel values is extremely fragile.  355 pixels away from what?  @raclim we should revisit this.  Put both the eye `button` and the `input` inside of a `div` with `position: relative`.  Then use `position: absolute; right: 0` on the `button` to have it be aligned to the right side of the input.  If we do that then we shouldn't need this new media query to adjust the `right` position because it would always be `0` regardless of screen size.

raclim: I think that completely works. Maybe we could make those adjustments within this PR if @rushikeshg25 is down?

raclim:
I'm going to close this, since this has been addressed by https://github.com/processing/p5.js-web-editor/pull/3176.

PR #3192, opened 2024-07-18
feat: enable `network policy` and disable legacy `metadata` endpoints in main.tf

Abhinavcode13:
This feature is added to make the `GKE cluster` more secure by controlling `pod` traffic and preventing unauthorized access to metadata.

cc: @raclim

### DIFF
--- a/infrastructure/terraform/main.tf
+++ b/infrastructure/terraform/main.tf
@@ -19,6 +19,10 @@ resource "google_container_cluster" "primary" {
   # node pool and immediately delete it.
   remove_default_node_pool = true
   initial_node_count       = 1
+
+  network_policy {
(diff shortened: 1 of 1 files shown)


raclim:
Thanks for raising this! I'll try to get a deeper look into this by early next week!

Abhinavcode13:
sure!

raclim:
Thanks so much again for your work and patience on this! 

After some review, we've decided not to go through with these changes for now due to other changes we would like to prioritize in our GKE Setup. However, we are open to revisiting this again once that work has been completed, so I've added this to a board to note that. 

I'm sorry that we couldn't get this in, but please feel free to check out our other issues!

PR #3198, opened 2024-07-25
fixed the issue 1852 

### LINKED ISSUE #1852: Set default iframe background color to white
#### Nature of issue?

- Found a bug

#### Details about the bug: 

Hi @catarak I have run into this as well, here is a screenshot of how text looks in the Preview window in Dark mode, but in High Contrast mode it is similar, defaulting to black text on a very dark background:
image
Here's an example of a sketch that displays this issue: https://editor.p5js.org/ml5/sketches/XCFPKqIKrt

_Originally posted by @leemark in https://github.com/processing/p5.js-web-editor/issues/1400#issuecomment-803752647_

@catarak: I think the issue here is that the iframe should have a default background color of white, as is the norm for websites.

TUSHAR1651:
Fixes #issue-number 1852

Changes:

* has no linting errors (`npm run lint`)
* has no test errors (`npm run test`)
* is from a uniquely-named feature branch and is up to date with the  `Issue1852` branch.
* [ ] is descriptively named and links to an issue number, i.e. `Fixes #1852`


TUSHAR1651:
I have the fixed the issue in this PR on using the dark mode the text colour was not looking good so i trued to change it

raclim:

Although this PR does accomplish the issue of darker text not appearing on dark mode, it also seems to affect the background of the other themes as well. Ideally, this should be isolated to just the Dark and High Contrast themes. Due to this, I also feel that these changes should also be more dependent on when the theme is changed. There also seems to be an additional change to the Dockerfile, which I feel might not be related to this issue—ideally we would like to keep these types of changes to a minimum! 

Overall though, I know that what the solution for this issue should look like isn't very clear yet at the moment 😅—I've been hoping to find one that could elegantly implement this without it looking too visually out of place. Due to some of the issues listed above I don't think we'll be merging this in for now—I'm sorry that we couldn't get this in, but please feel free to continue brainstorming on what the solution for this could look like!

Issue #2080, opened 2022-11-04
Error message claiming size limit of 250MB reached


garystil:
- p5.js version: web editor
- Web browser and version: Safari Version 15.6.1
- Operating System: MacOSX 10.15.7

When I attempt to upload a file to a new or existing sketch, I receive the following response:

“Error: You cannot upload any more files. You have reached the total size limit of 250 MB. If you would like to upload more, please remove the ones you aren’t using anymore by in your [assets](https://editor.p5js.org/assets).”

My account shows a total asset size of 344.86 MB, exceeding the 250 MB limit. However, after deleting a number of files, my calculations result in a total asset size of only 143 MB. I can't figure out this discrepancy that has persisted for several months. Thanks in advance for your help.

dewanshDT:
well I found an opposite issue, even after exceeding the limit I can upload assets
![assetsduplicated](https://user-images.githubusercontent.com/71703033/219931395-17996b87-4cbc-42c4-b050-25c6ae083ff5.png)

Nishthajain7:
Can i work on this issue?

PR #3219, opened 2024-08-12
The upload file button will be disabled after reaching the limit 250MB

Noorain464:
Fixes #issue-number: Error message claiming size limit of 250MB reached #2080
Changes:


client/modules/IDE/components/FileNode.jsx
```diff
@@ -11,6 +11,7 @@ import DownArrowIcon from '../../../images/down-filled-triangle.svg';
 import FolderRightIcon from '../../../images/triangle-arrow-right.svg';
 import FolderDownIcon from '../../../images/triangle-arrow-down.svg';
 import FileTypeIcon from './FileTypeIcon';
+import { sizeLimit, currentSize } from './AssetSize';
```
raclim, line 14: This doesn't seem to import the correct constants you've created in `AssetSize.jsx`!

client/modules/IDE/components/FileNode.jsx
```diff
@@ -373,14 +379,24 @@ class FileNode extends React.Component {
...
                       <li>
-                        <button
-                          aria-label={t('FileNode.UploadFileARIA')}
-                          onClick={this.handleClickUploadFile}
-                          onBlur={this.onBlurComponent}
-                          onFocus={this.onFocusComponent}
-                        >
-                          {t('FileNode.UploadFile')}
-                        </button>
+                        {currentSize < sizeLimit ? (
```
raclim, line 382: Maybe this could be handled by changing the css property depending on `currentSize < sizeLimit` rather than the whole `button`?

raclim:

After reviewing this, it seems like you're trying to add an alert message once a user has reached the 250MB limit for storing assets. We do have error handling for this implemented already in the File Upload Modal ([`UploadFileModal.jsx`](https://github.com/processing/p5.js-web-editor/blob/41d0303384b4e72887224eff37af99574c813ee0/client/modules/IDE/components/UploadFileModal.jsx)), although I do think it could be interesting to explore using an alert instead or doing this in the sidebar navigation!

I don't think the overall changes here are addressing the original issue, but I've left a few notes on what's here so far! I feel that the original issue seems to involve how the asset sizes are being determined, so it might be better to try working from there instead. I'm going to close this pull request for now, but please feel free to revisit this with any updates!

PR #3221, opened 2024-08-16
fix (some) test suite warnings

khanniie:
Fixes the test suite warnings that are unrelated to the React lifecycle, which I assume will be resolved as a side effect of the ongoing class -> functional conversion.

Changes:
- Provides the props that were missing but marked as required in the EditorAccessibility tests and ErrorModal tests
- creates a fake version of window.confirm which is missing from the jsdom environment but used in the FileNode tests
- hides the extra log in the createProject test and changes it from a log to a warning type in the controller itself

### DIFF
--- a/client/modules/IDE/components/ErrorModal.unit.test.jsx
+++ b/client/modules/IDE/components/ErrorModal.unit.test.jsx
@@ -8,20 +8,20 @@ jest.mock('../../../i18n');
 
 describe('<ErrorModal />', () => {
   it('renders type forceAuthentication', () => {
-    render(<ErrorModal type="forceAuthentication" closeModal={jest.fn()} />);
+    render(<ErrorModal type="forceAuthentication" closeModal={jest.fn()} service="google" />);
 
     expect(screen.getByText('Login')).toBeVisible();
     expect(screen.getByText('Sign Up')).toBeVisible();
   });
 
   it('renders type staleSession', () => {
-    render(<ErrorModal type="staleSession" closeModal={jest.fn()} />);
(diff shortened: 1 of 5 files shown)


raclim (approved):
This is awesome! Thanks so much for working on this! 

Pinging at @nahbee10 to see if she has any additional thoughts as well—otherwise I think I'm going go ahead with merging this in tomorrow!

raclim:
I'm going to go ahead with merging this in, but please feel free to note if any additional thoughts comes up!

khanniie:
Yay! thanks rachel!! : )

client/modules/IDE/components/ErrorModal.unit.test.jsx
```diff
@@ -8,20 +8,30 @@ jest.mock('../../../i18n');
...
 
     expect(screen.getByText('Login')).toBeVisible();
     expect(screen.getByText('Sign Up')).toBeVisible();
   });
 
   it('renders type staleSession', () => {
-    render(<ErrorModal type="staleSession" closeModal={jest.fn()} />);
+    render(
+      <ErrorModal type="staleSession" closeModal={jest.fn()} service="google" />
+    );
```
lindapaiste, line 26: Personally I would say that the more logical fix for the PropTypes error would be to make it so that the `service` is not a required prop.  It's only needed when `type="oauthError"` and it's not used in any other circumstance so it's kind of weird that it's always required.  (I'm not sure if PropTypes has a way to require it depending on the value of the `type`.  The "correct" thing would be to require the `service` if and only if the `type` is "oauthError" but that might be getting too complicated.)

PR #3231, opened 2024-09-11
React class to functional conversion - final

### LINKED ISSUE #2709: Changing class component to Functional Component.
### Increasing Access

It is important to update components according to new Functional components.

The component uses useRef for setting a reference to the Codemirror instance which can be used in any useEffect and function. I am referencing a pull request for these issue. I have done a pr before and changed my commit but it was unable to reach to maintainers.

nahbee10:
Fixes https://github.com/processing/p5.js-web-editor/issues/2709

Changes:
applied react functional component conversion to CollectionList, SketchList, FileNode, ConsoleInput, AssetList. This would be the last conversion ticket


client/modules/IDE/components/SketchListRowBase.jsx
```diff
@@ -0,0 +1,167 @@
...
+
+  const name = (
+    <>
+      <Link to={url}>{renameOpen ? '' : sketch.name}</Link>
+      {renameOpen && (
+        <input
+          value={renameValue}
+          onChange={handleRenameChange}
+          onKeyDown={handleRenameEnter}
+          onBlur={handleRenameBlur}
```
khanniie, line 94: was comparing this to the original component and noticed there was an event.stopPropagation on that one -- do we need this here?

client/modules/IDE/components/SketchListRowBase.jsx
```diff
@@ -0,0 +1,167 @@
...
+
+  const closeRename = () => setRenameOpen(false);
+
+  const updateName = useCallback(() => {
+    if (renameValue.trim().length > 0) {
+      changeProjectName(sketch.id, renameValue.trim());
+    }
+  }, [renameValue, sketch.id, changeProjectName]);
+
+  const handleRenameChange = (e) => setRenameValue(e.target.value);
```
khanniie, line 48: so i was looking into useCallback since that function is new to me and I read [in the docs](https://react.dev/reference/react/useCallback#should-you-add-usecallback-everywhere) that it's typically used when functions are passed as props to prevent unnecessary rerenders -- I was wondering if we want to apply it to all of these functions that are being passed as onClick props? Or what's the logic for only using it for a few of them?

khanniie (changes requested):
not done reviewing yet but figured i should submit the first wave of comments so you can get started!

client/modules/IDE/components/SketchListRowBase.jsx
nahbee10: Good point! Added to the component :)

client/modules/IDE/components/SketchListRowBase.jsx
nahbee10: yeah @khanniie some of them don't have to be rendered using useCallback since:
- When to use useCallback:
1. Passing callbacks to child components: If a callback is passed as a prop to a child component and that child uses React.memo to prevent re-renders, using useCallback ensures that the callback doesn't get recreated on each parent re-render, preventing unnecessary child re-renders.
2. Expensive computations: If the function involves expensive computations (e.g., heavy calculations or side effects), wrapping it in useCallback can avoid recreating the function unnecessarily.
- Don't have to use useCallback:
1. No performance gain: For simple state updates or inline event handlers (e.g., handleChange, handleClick), memoization won't provide much benefit and adds unnecessary complexity.
2. Not passed as props or dependencies: If the function isn’t passed as a prop or used in a useEffect dependency array, memoizing it is unnecessary.

raclim (commented):
Thanks so much for working on this! It's really awesome that all of these files are getting revamped! 

I'm doing some quick manual testing here, and I couldn't find any major issues in most of the changed components except for `FileNode.jsx`, which seems to cause some crashes when opening certain example sketches (i.e 3D: shader as a texture). I wanted to also start pointing this out now, but will continue to review more closely!

client/modules/User/components/Collection.jsx
raclim, line 1: This file is so much neater!

client/modules/IDE/components/AssetList.jsx
```diff
@@ -1,187 +1,67 @@
...
-import { Link } from 'react-router-dom';
+import React, { useEffect } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
 import { Helmet } from 'react-helmet';
-import prettyBytes from 'pretty-bytes';
-import { useTranslation, withTranslation } from 'react-i18next';
-import MenuItem from '../../../components/Dropdown/MenuItem';
-import TableDropdown from '../../../components/Dropdown/TableDropdown';
-
+import { withTranslation, useTranslation } from 'react-i18next';
```
raclim: I think this component can use the `useTranslation` hook instead of `withTranslation`! This probably also applies to the `Filenode` and  `CollectionList` components in this PR too!

client/modules/IDE/components/CollectionList/CollectionList.jsx
```diff
@@ -22,133 +22,135 @@ import CollectionListRow from './CollectionListRow';
...
+    if (!loading) {
+      setHasLoadedData(true);
     }
-  }
+  }, [loading]);
 
-  getTitle() {
-    if (this.props.username === this.props.user.username) {
-      return this.props.t('CollectionList.Title');
+  const getTitle = useCallback(() => {
```
raclim: Would maybe `useMemo()` be preferable to `useCallback()` here?

client/modules/IDE/components/Console.jsx
```diff
@@ -227,6 +227,10 @@ const Console = () => {
     };
   });
 
+  useEffect(() => {
```
raclim: Just wondering if you intended on keeping this for this PR!

client/modules/IDE/components/SketchList.jsx
```diff
@@ -1,433 +1,177 @@
...
-import MenuItem from '../../../components/Dropdown/MenuItem';
-import TableDropdown from '../../../components/Dropdown/TableDropdown';
-import dates from '../../../utils/formatDate';
-import * as ProjectActions from '../actions/project';
-import * as ProjectsActions from '../actions/projects';
-import * as CollectionsActions from '../actions/collections';
-import * as ToastActions from '../actions/toast';
-import * as SortingActions from '../actions/sorting';
-import * as IdeActions from '../actions/ide';
+import * as ProjectsActions from '../actions/projects'; // Added Projects actions
```
raclim: I think some of these comments can be removed!

client/modules/IDE/components/SketchList.jsx
```diff
@@ -1,433 +1,177 @@
...
-  mobile: false
-};
-
-function mapDispatchToPropsSketchListRow(dispatch) {
-  return bindActionCreators(
-    Object.assign({}, ProjectActions, IdeActions),
-    dispatch
+  }, [sketches]);
+
+  const getSketchesTitle = useCallback(
```
raclim: Could this also potentially be `useMemo()`?

client/modules/IDE/components/SketchList.jsx
```diff
@@ -1,433 +1,177 @@
...
+      return direction === SortingActions.DIRECTION.ASC
+        ? t('SketchList.ButtonLabelDescendingARIA', { displayName })
+        : t('SketchList.ButtonLabelAscendingARIA', { displayName });
+    },
+    [sorting, t]
+  );
 
-  _renderFieldHeader = (fieldName, displayName) => {
-    const { field, direction } = this.props.sorting;
+  const renderFieldHeader = (fieldName, displayName) => {
```
raclim: Do you feel that maybe using `useCallback()` might make sense here?

client/modules/IDE/components/SketchListRowBase.jsx
```diff
@@ -0,0 +1,168 @@
+import PropTypes from 'prop-types';
+import slugify from 'slugify';
+import React, { useState, useRef, useCallback } from 'react';
+import { Link } from 'react-router-dom';
+import { bindActionCreators } from 'redux';
+import { connect } from 'react-redux';
+import * as ProjectActions from '../actions/project';
+import * as IdeActions from '../actions/ide'; // Adding the missing Ide actions
```
raclim: I feel that this comment could be removed as well!

client/modules/IDE/components/FileNode.jsx
```diff
@@ -190,234 +184,185 @@ class FileNode extends React.Component {
...
+                focusable="false"
+                aria-hidden="true"
+              />
+            </span>
+          )}
+          {isFolder && (
+            <div className="sidebar__file-item--folder">
+              <button
+                className="sidebar__file-item-closed"
+                onClick={showFolderChildren(id)}
```
raclim: I think the issue with this file is stemming from lines 248 and 260! I think they should've been passing a function instead of an object (i.e `onClick={() => showFolderChildren(id)}` vs. `onClick={showfolderChildren(id)}`.

raclim (changes requested):
Just took a closer look through some of the files! Added a few more thoughts, let me know if you have any questions about any of them, thanks!

nahbee10:
Thanks for the review and fix @raclim and @khanniie !! Just updated code and requested review again :)

Seems like the error from filenode is solved after fixing the button onclick function @raclim!!

client/modules/IDE/components/AssetList.jsx
```diff
@@ -1,187 +1,63 @@
...
-    super(props);
-    this.props.getAssets();
-  }
-
-  getAssetsTitle() {
-    return this.props.t('AssetList.Title');
-  }
+  useEffect(() => {
+    dispatch(AssetActions.getAssets());
+  }, [dispatch]);
```
khanniie: i feel like since this is from the constructor we probably don't need anything in the dependency array?

client/modules/IDE/components/CollectionList/CollectionList.jsx
```diff
@@ -22,133 +22,135 @@ import CollectionListRow from './CollectionListRow';
...
-    if (props.projectId) {
-      props.getProject(props.projectId);
+  useEffect(() => {
+    if (projectId) {
+      getProject(projectId);
     }
+    getCollections(propsUsername || user.username);
+    resetSorting();
+  }, [
+    projectId,
```
khanniie: same here, since its from the constructor do we need the dependency array?

khanniie (dismissed):
thank you nahee!!!!

raclim:
Thanks so much for your work on this @nahbee10! 

I think maybe one additional change that would be great is to remove the `console.log` in the `client/modules/IDE/components/Console.jsx` file if that wasn't needed for a specific purpose! Otherwise, I think this is good to go!

raclim (approved):
Thanks so much again @nahbee10 and @khanniie! I'm going to merge this in! Feel free to make any additional changes if anything comes up!

client/modules/IDE/components/AssetList.jsx
lindapaiste: `dispatch` never changes its value so it doesn't make a difference whether it's in the dependency array or not.  The effect will run one time on mount either way.  Personally I'm in the habit of including it to appease linting rules such as `react-hooks/exhaustive-deps`.  But we don't have that rule enabled on this repo (at least not yet) because we violate it all the time.

client/modules/IDE/components/CollectionList/CollectionList.jsx
lindapaiste: If the value of `propsUsername` were to change then we would want to redo the fetch with the new username.  So you need that as a dependency.

If the `projectId` were to change then you would want to redo the project fetch, but not the collection fetch.  So I would have multiple `useEffect` hooks with separate dependencies.

```
  const username = propsUsername || user.username;
  
  useEffect(() => {
    if (projectId) {
      getProject(projectId);
    }
  }, [projectId]);
  
  useEffect(() => {  
    getCollections(username);
  }, [username]);
  
 useEffect(() => {
   resetSorting();
 }, []);
```

I'm not sure what you mean by "from the constructor"?  In practice, with the way that the app is set up, the value of `propsUsername` probably doesn't change without a total unmount and remount because we never link from one user profile to another.  Maybe if you've viewing another user's collection page and you click the collections link from your user dropdown?  That's the sort of situation where there could be bugs if we don't have the right dependencies.

client/modules/IDE/components/CollectionList/CollectionList.jsx
```diff
@@ -22,133 +22,125 @@ import CollectionListRow from './CollectionListRow';
...
+    setAddingSketchesToCollectionId
+  ] = useState(null);
 
-    if (props.projectId) {
-      props.getProject(props.projectId);
+  useEffect(() => {
+    if (projectId) {
+      getProject(projectId);
     }
+    getCollections(propsUsername || user.username);
```
lindapaiste, line 50: It seems like the `username` should be required and it's a mistake that we don't have it marked as such.  I'm basing that off of two observations:

1. The current code does not have any sort of fallback for if the username is not provided in the `getCollections()` call.
2. The current condition in `getTitle` for checking if this is your own collection or someone else's is `if (this.props.username === this.props.user.username) {`.  So that definitely implies that when viewing your own collection we would expect that ` username` prop is explicitly provided and that it matches the username of the current logged-in user.

client/modules/IDE/components/AssetList.jsx
```diff
@@ -1,187 +1,63 @@
...
-const AssetMenu = ({ item: asset }) => {
+const AssetList = () => {
   const { t } = useTranslation();
-
   const dispatch = useDispatch();
+  const { username, assetList, loading } = useSelector((state) => ({
+    username: state.user.username,
+    assetList: state.assets.list,
+    loading: state.loading
+  }));
```
lindapaiste, line 16: Not a big deal but FYI it's better to have separate `useSelector` calls for each piece of state that you read:

> ### Call useSelector Multiple Times in Function Components
>
> When retrieving data using the useSelector hook, prefer calling useSelector many times and retrieving smaller amounts of data, instead of having a single larger useSelector call that returns multiple results in an object. Unlike mapState, useSelector is not required to return an object, and having selectors read smaller values means it is less likely that a given state change will cause this component to render.
>
> However, try to find an appropriate balance of granularity. If a single component does need all fields in a slice of the state , just write one useSelector that returns that whole slice instead of separate selectors for each individual field.

https://redux.js.org/style-guide/#call-useselector-multiple-times-in-function-components

PR #3236, opened 2024-09-21
fix: DEBOUNCING ISSUES

sohelhussain:
#3237


raclim:
Thank you so much for submitting this change and raising this issue!

Although this is a good start, looking through this I think there are a few issues with this PR where the debounce logic doesn't seem to be handled properly. We'd also ideally like to apply this to both the username and the email fields as well. 

Since we have a few PRs for this issue, I'm going to close this one for now but please feel free to make any updates and reopen this one!

PR #3259, opened 2024-11-01
resolved the repeated code issue between two files

### LINKED ISSUE #2805: Repeated code between server `previewGeneration.js` and client `EmbedFrame.jsx`

harshit10541:
Fixes #2805 

Changes:


raclim:
Thanks for taking a look at this! I think there were a few tasks and questions in the associated issue that would be ideal for folks to address before making changes to the code! Would you be able to add your thoughts or answers to some of them in the description to help get a sense of your changes?

raclim:
I'm going to close this PR for now due to inactivity—I'm sorry that we couldn't get this one in, but please feel free to reopen this with any updated changes or to check out our other issues!

PR #2366, opened 2023-08-08
Convert `Collection` to a function component

lindapaiste:
Progress on #2358 

Merge after #2253 

Changes:
- Convert `Collection` to a function component.
- Use Redux hooks.
- Don't fetch the collection from the API if it's already loaded.

I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] has no test errors (`npm run test`)
* [x] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #3270, opened 2024-11-07
Update Collection View

### LINKED ISSUE #3269: Collections Not Visible to Non-Owners
### p5.js version

_No response_

Mac OS

### Web browser and version

All Browsers

### Actual Behavior

Collections is not visible to users who do not own the collection. 

Screenshot 2024-11-07 at 2 01 00 PM

### Expected Behavior

Collection to be visible (pictured below). 
Screenshot 2024-11-07 at 12 35 56 PM

### Steps:
1. Find a URL to a collection that is not created under your own account. i.e `/:username/collections/:collection-id`
2. Afterwards, a blank page should appear.

raclim:
Fixes #3269 

Changes:
- passes `username` as a prop to `Collection` component, which was previously not accessible 
- updates conditions to determine `isOwner`, which was previously returning `undefined` instead of a boolean
- imports `removeFromCollection` action directly into `CollectionItemRow`

### DIFF
--- a/client/modules/User/components/Collection.jsx
+++ b/client/modules/User/components/Collection.jsx
@@ -13,19 +13,16 @@ import ArrowDownIcon from '../../../images/sort-arrow-down.svg';
 import CollectionMetadata from './CollectionMetadata';
 import CollectionItemRow from './CollectionItemRow';
 
-const Collection = ({ collectionId }) => {
+const Collection = ({ collectionId, username }) => {
   const { t } = useTranslation();
   const dispatch = useDispatch();
 
-  const { user, collection, sorting, loading, username } = useSelector(
-    (state) => ({
-      user: state.user,
-      collection: getCollection(state, collectionId),
-      sorting: state.sorting,
-      loading: state.loading,
-      username: state.user.username
-    })
-  );
+  const { user, collection, sorting, loading } = useSelector((state) => ({
+    user: state.user,
+    collection: getCollection(state, collectionId),
+    sorting: state.sorting,
+    loading: state.loading
+  }));
 
   useEffect(() => {
     dispatch(CollectionsActions.getCollections(username));
@@ -34,7 +31,7 @@ const Collection = ({ collectionId }) => {
 
   const isOwner = () =>
     user != null &&
-    user.username &&
+    typeof user.username !== 'undefined' &&
     collection?.owner?.username === user.username;
 
   const hasCollection = () => !!collection;
@@ -160,7 +157,8 @@ const Collection = ({ collectionId }) => {
 };
 
 Collection.propTypes = {
-  collectionId: PropTypes.string.isRequired
+  collectionId: PropTypes.string.isRequired,
+  username: PropTypes.string.isRequired
 };
 
 export default Collection;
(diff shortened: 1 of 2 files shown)


client/modules/User/components/Collection.jsx
```diff
@@ -34,7 +29,7 @@ const Collection = ({ collectionId }) => {
 
   const isOwner = () =>
     user != null &&
-    user.username &&
+    typeof user.username !== 'undefined' &&
     collection?.owner?.username === user.username;
```
lindapaiste, line 33: There is definitely room to clean this up further.

The `user != null` check is unnecessary because `user` is `state.user` which is [always an object](https://github.com/processing/p5.js-web-editor/blob/develop/client/modules/User/reducers.js) even when there is no one logged in.

You are now passing `username` as a prop, which is the username of the collection owner, so you can use that instead of `collection?.owner?.username`.  Using `username` has a few advantages, including that it won't be `undefined` while the collection is fetching so you get an accurate boolean during loading.

Additionally, `username` is never `undefined`.  So there is no possibility that `username === user.username` when there is no logged in user (and `user.username` is `undefined`).  So you don't need to explicitly check that there is a `user.username`.  If `username === user.username` is `true` then there is a user logged in who is the the owner of this collection.  If it is `false` then either there is no user logged in or the logged in user is not the owner.

So you can write `const isOwner = () => username === user.username` and it will have the same effect.

[When I converted this component](https://github.com/processing/p5.js-web-editor/pull/2366/files) I wrote it as:
```
const currentUsername = useSelector((state) => state.user.username);
const isOwner = username === currentUsername;
```

lindapaiste: IMO it's important to use descriptive variable names like `currentUsername` and `ownerUsername` to avoid bugs like the one that you fixed here.

I know that I can be rude sometimes but I think you can understand my frustrations.  Because I already converted this component over a year ago and it was cleanly written and did not have that sort of mistake.

raclim: Thanks for catching these, there's probably still a lot of other areas this component could be improved here! 

Since I wasn't able to get through all of your work and made some changes to these files that differed from yours, I do get your frustrations on how it probably feels like this could've been preventable. Although it'll be a lot of work since I didn't keep up with the PRs from back then, if you'd like to, we can get still try to get them in. I could be wrong, but I think most of them could probably be merged with some minimal changes except for a few (like the Common Table Components and the CodeMirror one, esp since it's part of the fellowship). I was also planning to [upgrade the Node.js minor version](https://github.com/processing/p5.js-web-editor/compare/develop...update-node-version) in the near future, which I think might conflict with [this one](https://github.com/processing/p5.js-web-editor/pull/2220). 

This is a bit of a tangent and definitely depends on the context, but I also personally feel that it's okay for code to end up a little messy and have mistakes sometimes! While I do think it's important to try not to make them, I don't want to view them negatively because I feel like they're a really normal part of any process, and in most cases, can usually be fixed.

raclim: Wanted to add that I meant the latter more as a point of conversation, rather than disagreeing!

Issue #3268, opened 2024-11-07
Javascript error in running this sketch


jht9629-nyu:
### p5.js version

Web Editor: v2.12.9 , p5.js: v1.9.2

### What is your operating system?

Mac OS

### Web browser and version

Version 124.0.6367.62 (Official Build) (arm64)

### Actual Behavior

Run this sketch
https://editor.p5js.org/sairamved/sketches/kHR1qrQC5

Get this error

const { _report } = window.p5;

TypeError: Cannot read properties of undefined (reading '_report')
    at undefined:2:36978

### Expected Behavior

No error

### Steps:
open editor 
https://editor.p5js.org/sairamved/sketches/kHR1qrQC5
run

1. 2. 3.
### Snippet:
```js

// Paste your code here :)

```

raclim:
Thanks for raising this @jht9629-nyu, will look into it!

raclim:
Looking into it, I think the error was triggered by lines 4 and 6 in the `index.html` file. Once I removed them, they were able to disappear for me. Since the console shouldn't be returning these types of messages, I'll try to make some adjustments in the [previewEntry.js file](https://github.com/processing/p5.js-web-editor/blob/5e5ef5f19b70fc3cd7a955d7b07fdfa6443b9321/client/utils/previewEntry.js#L151), which I think is where it can be resolved, and push it out within the next few days!

jht9629-nyu:
Thanks for the response.
Best to be safe - all kinds of extra characters can end up in index.html
I think this student was following chat-GPT

Be well,
-{ john-henry }-[ thompson ]

lindapaiste:
I'm seeing an additional error which may or may not be related
> GET https://editor.p5js.org/sairamved/sketches/undefined 404 (Not Found)

We should look into where that is coming from, and if we are loading files properly.

lindapaiste:
> Looking into it, I think the error was triggered by lines 4 and 6 in the `index.html` file. Once I removed them, they were able to disappear for me. Since the console shouldn't be returning these types of messages, I'll try to make some adjustments in the [previewEntry.js file](https://github.com/processing/p5.js-web-editor/blob/5e5ef5f19b70fc3cd7a955d7b07fdfa6443b9321/client/utils/previewEntry.js#L151), which I think is where it can be resolved, and push it out within the next few days!

[...]

PR #3276, opened 2024-11-13
Update Console Components

### LINKED ISSUE #3271: Console history not accessible with up arrow
### p5.js version

1.11.1

Linux

### Web browser and version

Firefox Version: 131.0.3 / 20241014102019

### Actual Behavior

Pressing the up arrow key does not bring old code into the console's input field.

### Expected Behavior

Pressing the up and down arrow keys cycles through history, so that we don't have to retype any of the code we've just entered
This used to be the case until recently; not sure at what point the regression happened, however.

### Steps:
1. Enter some code into the console's input field, e.g. `1`. Press enter to execute it.
2. Press the up arrow key. 
3. Observe that nothing happens.
4. Input some more text, without pressing enter.
5. Press up arrow again.
6. Observe that your just-entered code gets deleted.

### Snippet:
N/A (Default sketch works just fine for reproducing.)

raclim:
Related Issues: 
- Fixes #3271 
- Progress towards #3268 
- (Not an issue, also just makes additional changes to `Console.jsx`) 

Changes:
- moves the styling for the Console into a separate utility file, `consoleStyles.js`
- separates hook in `ConsoleInput.jsx` into individual ones to handle keyboard inputs for Enter, ArrowUp, and ArrowDown
- removes named import in `previewEntry.js` to prevent "_report is undefined" warnings

### DIFF
--- a/client/modules/IDE/components/Console.jsx
+++ b/client/modules/IDE/components/Console.jsx
@@ -38,8 +38,11 @@ const Console = () => {
   const handleMessageEvent = useHandleMessageEvent();
 
   useEffect(() => {
-    listen(handleMessageEvent);
-  }, [handleMessageEvent]);
+    const unsubscribe = listen(handleMessageEvent);
+    return function cleanup() {
+      unsubscribe();
+    };
+  });
 
   const handleClearConsole = () => dispatch(ConsoleActions.clearConsole());
   const handleCollapseConsole = () => dispatch(IDEActions.collapseConsole());
@@ -92,13 +95,7 @@ const Console = () => {
           />
         </div>
         {isExpanded && isPlaying && (
-          <ConsoleInput
-            theme={theme}
-            dispatchConsoleEvent={(event) =>
-              dispatch(ConsoleActions.dispatchConsoleEvent(event))
-            }
-            fontSize={fontSize}
-          />
+          <ConsoleInput theme={theme} fontSize={fontSize} />
         )}
       </div>
     </section>
--- a/client/modules/IDE/components/Console.jsx
+++ b/client/modules/IDE/components/Console.jsx
@@ -1,31 +1,10 @@
-import React, { useRef, useEffect } from 'react';
+import React, { useRef, useEffect, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
-
-import { bindActionCreators } from 'redux';
 
 import { useSelector, useDispatch } from 'react-redux';
 import classNames from 'classnames';
 import { Console as ConsoleFeed } from 'console-feed';
-import warnLightUrl from '../../../images/console-warn-light.svg?byUrl';
-import warnDarkUrl from '../../../images/console-warn-dark.svg?byUrl';
-import warnContrastUrl from '../../../images/console-warn-contrast.svg?byUrl';
-import errorLightUrl from '../../../images/console-error-light.svg?byUrl';
-import errorDarkUrl from '../../../images/console-error-dark.svg?byUrl';
-import errorContrastUrl from '../../../images/console-error-contrast.svg?byUrl';
-import debugLightUrl from '../../../images/console-debug-light.svg?byUrl';
-import debugDarkUrl from '../../../images/console-debug-dark.svg?byUrl';
-import debugContrastUrl from '../../../images/console-debug-contrast.svg?byUrl';
-import infoLightUrl from '../../../images/console-info-light.svg?byUrl';
-import infoDarkUrl from '../../../images/console-info-dark.svg?byUrl';
-import infoContrastUrl from '../../../images/console-info-contrast.svg?byUrl';
 import ConsoleInput from './ConsoleInput';
-
-import commandLightUrl from '../../../images/console-command-light.svg?byUrl';
-import resultLightUrl from '../../../images/console-result-light.svg?byUrl';
-import commandDarkUrl from '../../../images/console-command-dark.svg?byUrl';
-import resultDarkUrl from '../../../images/console-result-dark.svg?byUrl';
-import commandContrastUrl from '../../../images/console-command-contrast.svg?byUrl';
-import resultContrastUrl from '../../../images/console-result-contrast.svg?byUrl';
 
 import UpArrowIcon from '../../../images/up-arrow.svg';
 import DownArrowIcon from '../../../images/down-arrow.svg';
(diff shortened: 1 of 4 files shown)


client/utils/previewEntry.js
```diff
@@ -175,6 +176,6 @@ if (_report) {
     paths.forEach((path) => {
       newMessage = newMessage.replaceAll(path, window.objectPaths[path]);
     });
-    _report.apply(window.p5, [newMessage, method, color]);
+    p5Call.apply(window.p5, [newMessage, method, color]);
```
lindapaiste: Previously the `apply` was on `window.p5._report` (a function).  Now it is on `window.p5` (an object).  So that's a mistake.  I would keep a `_report` variable but define it inside the `if (p5Call)` branch.

client/modules/IDE/components/Console.jsx
```diff
@@ -35,197 +14,36 @@ import * as ConsoleActions from '../../IDE/actions/console';
...
+
   useEffect(() => {
-    const unsubscribe = listen(handleMessageEvent);
-    return function cleanup() {
-      unsubscribe();
-    };
-  });
+    listen(handleMessageEvent);
+  }, [handleMessageEvent]);
+
```
lindapaiste: I would keep the unsubscribe on cleanup behavior.

client/modules/IDE/components/Console.jsx
```diff
@@ -268,15 +86,17 @@ const Console = () => {
...
             key={`${theme}-${fontSize}`}
           />
         </div>
         {isExpanded && isPlaying && (
           <ConsoleInput
             theme={theme}
-            dispatchConsoleEvent={dispatchConsoleEvent}
+            dispatchConsoleEvent={(event) =>
+              dispatch(ConsoleActions.dispatchConsoleEvent(event))
+            }
```
lindapaiste: IMO it's better to do the dispatching directly from `ConsoleInput`.

client/modules/IDE/components/ConsoleInput.jsx
```diff
@@ -24,75 +24,116 @@ function ConsoleInput({ theme, dispatchConsoleEvent, fontSize }) {
...
     });
 
-    cmInstance.current.on('keydown', (cm, e) => {
+    cmInstance.current.getWrapperElement().style['font-size'] = `${fontSize}px`;
+
+    return () => {
+      if (cmInstance.current) {
+        cmInstance.current = null;
+      }
+    };
```
lindapaiste: Probably don't need a cleanup here if all that it's doing is clearing a local variable.  If there's a destroy or cleanup function on the CM instance we would want to call that though.

client/modules/IDE/components/ConsoleInput.jsx
```diff
@@ -24,75 +24,116 @@ function ConsoleInput({ theme, dispatchConsoleEvent, fontSize }) {
...
+    if (cmInstance.current) {
+      cmInstance.current.on('keydown', handleEnterKey);
+    }
+
+    return () => {
+      if (cmInstance.current) {
+        cmInstance.current.off('keydown', handleEnterKey);
+      }
+    };
+  }, [commandHistory, dispatchConsoleEvent]);
```
lindapaiste: The `off` and `on` is good.  That will prevent us having multiple event handlers.

Having `dispatchConsoleEvent` as a dependency means that this effect will re-run on every render of `Console` because the function is declared inline and is not memoized.  You could memoize it (with `useCallback`) but I would instead handle the dispatching here instead of passing a function prop.

client/modules/IDE/components/ConsoleInput.jsx
```diff
@@ -24,75 +24,116 @@ function ConsoleInput({ theme, dispatchConsoleEvent, fontSize }) {
...
         const newCursor = Math.min(
           commandCursor + 1,
           commandHistory.length - 1
         );
-        cmInstance.current.getDoc().setValue(commandHistory[newCursor] || '');
-        const cursorPos = cmInstance.current.getDoc().getLine(0).length - 1;
-        cmInstance.current.getDoc().setCursor({ line: 0, ch: cursorPos });
+        cm.setValue(commandHistory[newCursor] || '');
+        const cursorPos = cm.getDoc().getLine(0).length - 1;
+        cm.getDoc().setCursor({ line: 0, ch: cursorPos });
```
lindapaiste, line 78: Good change ✔️

client/modules/IDE/components/ConsoleInput.jsx
```diff
@@ -24,75 +24,116 @@ function ConsoleInput({ theme, dispatchConsoleEvent, fontSize }) {
...
+      }
     };
-  }, []);
-
-  useEffect(() => {
-    cmInstance.current.setOption('theme', `p5-${theme}`);
-    cmInstance.current.getWrapperElement().style['font-size'] = `${fontSize}px`;
-    cmInstance.current.refresh();
-  }, [theme, fontSize]);
+  }, [commandCursor, commandHistory]);
```
lindapaiste, line 119: We def need these as dependencies if we want to use those variables, but this does mean that we are changing the handler very frequently.  The goal with having both values in the same state is that you can access both values from within a `prevState` callback.  So the the effect doesn't need to re-run on every change of the variables.

lindapaiste (changes requested):
`previewEntry.js` needs to be fixed.

`ConsoleInput.jsx` looks pretty solid but could be cleaner.  I'll see if I can write a PR with what I have in mind.

client/utils/previewEntry.js
lindapaiste: Circling back to this, perhaps the only change that we need in this file is from `const { _report } = window.p5;` to `const _report = window.p5?._report;`.

However I do have questions as to why `window.p5` is `undefined` in the first place.  I'll leave some more info on the issue.

client/modules/IDE/components/ConsoleInput.jsx
raclim: Yeah I think this was what I started to become unsure about as I was working on it 😅

client/utils/previewEntry.js
raclim: Thanks for looking further into this!! I actually didn't realize there were so many issues about this already 😂 

I think for now, only changing `window.p5` to `window.p5?.report` makes sense to me. 

I'll change this PR as progress towards the issue, since it seems like we need to change where `previewScripts` is loading, but I'm not sure where the best placement might be for it yet.

client/modules/IDE/utils/consoleStyles.js
```diff
@@ -0,0 +1,189 @@
...
+      );
+    case 'contrast':
+      return Object.assign(
+        CONSOLE_FEED_CONTRAST_STYLES || {},
+        CONSOLE_FEED_CONTRAST_ICONS,
+        CONSOLE_FEED_SIZES,
+        style
+      );
+    default:
+      return '';
```
lindapaiste: This is supposed to return an object, right?  So `return '';` doesn't make sense.

client/modules/IDE/components/ConsoleInput.jsx
```diff
@@ -117,7 +153,6 @@ function ConsoleInput({ theme, dispatchConsoleEvent, fontSize }) {
 
 ConsoleInput.propTypes = {
   theme: PropTypes.string.isRequired,
-  dispatchConsoleEvent: PropTypes.func.isRequired,
   fontSize: PropTypes.number.isRequired
 };
```
lindapaiste, line 149: `theme` and `fontSize` are both selected from Redux.  so those can be selected inside the `ConsoleInput` rather than selected by the parent and passed down.

lindapaiste: Although now I see that the parent still needs to select them regardless because it uses them for the `ConsoleFeed` style.

client/modules/IDE/components/ConsoleInput.jsx
```diff
@@ -24,75 +25,110 @@ function ConsoleInput({ theme, dispatchConsoleEvent, fontSize }) {
...
-    cmInstance.current.on('keydown', (cm, e) => {
+    cmInstance.current.getWrapperElement().style['font-size'] = `${fontSize}px`;
+  }, []);
+
+  useEffect(() => {
+    if (cmInstance.current) {
+      cmInstance.current.setOption('theme', `p5-${theme}`);
+      cmInstance.current.getWrapperElement().style[
+        'font-size'
+      ] = `${fontSize}px`;
```
lindapaiste: I think we can add a `style` property to one of the `div`s that we control instead of manipulating the element.

```
<div ref={codemirrorContainer} className="console__editor" style={{ fontSize: `${fontSize}px` }} />
```

Where it is being set currently is a direct child of that div but it has the same effect.

client/modules/IDE/utils/consoleStyles.js
raclim: Yes!! I'm thinking of setting the default to the light theme instead?

client/modules/IDE/components/ConsoleInput.jsx
raclim: Just tried it out!

raclim:
I'm going to merge this in since I noted in the original issue that I would get something out for it soon, but am definitely open to following up with further changes here!

client/modules/IDE/components/ConsoleInput.jsx
```diff
@@ -23,76 +24,99 @@ function ConsoleInput({ theme, dispatchConsoleEvent, fontSize }) {
...
         dispatchMessage({
           type: MessageTypes.EXECUTE,
-          payload: {
-            source: 'console',
-            messages
-          }
+          payload: { source: 'console', messages }
         });
+
         dispatchConsoleEvent(consoleEvent);
```
bojidar-bg, line 48: Wouldn't you need to wrap this in `dispatch = useDispatch()`, `dispatch(...)`, like the reducers used in `Console.jsx`? Asking since, currently, while Up arrow does work, the console no longer echoes the inputted commands. (:

raclim: Yes!! Thank you so much for catching that 😂 I thought I had it in at some point! I'll try to have the update out for this at some point early next week.

bojidar-bg: The update works, thanks a ton! :tada:

PR #3282, opened 2024-11-25
how to render coordinates

catarak:
this isn't a real PR but you know that


raclim:
I'm going to close this since I was able to proceed with implementing everything thanks to the annotations here! Thanks so much again!!!!! :)

PR #3314, opened 2024-12-31
Jatin issue#3303

### LINKED ISSUE #3303: ProjectOptions closes before option can activate.
I'm working on a solve now, so there's no real need to read this issue carefully

### p5.js version

_No response_

Windows

### Web browser and version

Firefox 133.0.3 (64-bit)

### Actual Behavior

When I open the ProjectOptions by clicking the `+` and try to select options such as "Upload File" by tapping my trackpad (on my Lenovo ideapad Slim 5 Pro) the ProjectOptions UI closes but the upload file menu doesn't open. If I hold the option for more than a second it also closes the UI even while I am still holding the cursor down.
354814298-3c327be5-9e52-41a3-bf02-8022aed49db4

### Expected Behavior

I expect the upload file menu to open, which is what happens with I use my touchscreen, and when I use medium length full clicks. 

### Steps to reproduce

1) Click on `+` icon which in response opens ProjectOptions.
2) Depending on your hardware, try different methods of clicking on the `Upload file` option. Clicking and holding the mouse for a few seconds should show the issue for most hardware, but tapping the trackpad may be difficult to reproduce without one.

Jatin24062005:
Fixes #3303 

Changes:
Removed  of the setTimeout(() => dispatch(closeProjectOptions()), 0); allows the project options to remain open without prematurely closing when interacting with the menu, especially when using trackpads.


Jatin24062005:
@raclim, I fixed the test issues. Sorry for these— it is  my old pull request, and I didn't focus on checking the tests earlier.

raclim (changes requested):

I think the fix here might not be the best approach, as we probably want to keep `setTimeout()` to prevent the menu options from closing too soon and to ensure that the project options close after the action has been complete. 

In the original issue, it links to another PR that probably was the cause, so I think trying to reset some of its changes and investing from there could be a start! I'm thinking that one approach could be to implement a `useEffect()` to handle open and closing the dropdown when the state of `projectOptionsVisible` changes, rather than using `onBlur` and `onFocus` to resolve both this issue and the one where the project options dropdown doesn't close when a user clicks outside of it.

Jatin24062005:
@raclim  i tried you Approach and it is actually much better than the OnBlur and Onfocus method Please review these again... ugh : )

raclim (approved):
Thanks for the update on this, I'm glad that it seems to work! I removed the extra comment and added back in the `onContextMenu` prop just to try to keep the same behavior, but feel free to provide any additional comments on it if you feel otherwise!

PR #3320, opened 2025-01-17
[Menubar] Adds keyboard behaviors to the menubar, improves accessibility

### LINKED ISSUE #2933: Up/Down arrow keys not working on dropdown
### p5.js version

_No response_

Mac OS

### Web browser and version

_No response_

### Actual Behavior

Up/Down arrow keys are not working on dropdown items.

### Expected Behavior

We should be able to navigate through dropdown items using up/down arrow keys.

### Steps:
1. Go to https://editor.p5js.org
2. Click on File from menu
3. Press tab to move focus to the dropdown.
4. Use up/down arrow keys to navigate the dropdown.

tespin:
Fixes #2933 

This PR focuses on adding proper keyboard behaviors to the recently refactored Menubar.

## Update 
The Menubar component now has keyboard navigation and enhanced accessibility support. There are some other improvements to make which are listed at the end, but we can work on them in future PRs seeing as this one has gotten rather large. 

**Before:**
Navigation happened purely through Tab / Shift + Tab. Escape functionality worked but conventional Menubar behaviors were not accessible. Focus management was not connected to state.

https://github.com/user-attachments/assets/1ff66f9e-2843-4de2-9409-2a2df5c9eb19

**After:**
Menubar behaviors from the [WAI ARIA authoring guide](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/#keyboardinteraction) implemented. Tab behavior works as expected and initial focus styles are available. 

https://github.com/user-attachments/assets/a167b0c6-b21c-4f5c-819e-00dc94be79f6

## Key Changes
- Implemented roving tab index in `Menubar.jsx` for focus management
- Updated aria roles, such as `aria-orientation='horizontal'`on `Menubar` and aria-labels where applicable
- Navigation support, including:
  - Horizontal and vertical navigation
  - Home/end, enter/space support
- Improved HTML structure
  - wrapped `Menubar` in `nav` element
  - moved from `div`-based menubar to `ul/li`
  - the `Logo` is moved outside of the `nav`
  - left and right menu items now wrapped in `ul` elements with `role=group`
- added `SubmenuContext` for managing submenu state
- `MenubarSubmenu` supports `listbox` aria roles, such as with the options in `LanguageMenu`
- basic tests for checking component renders, keyboard behavior, and aria roles
- added keyboard styles
- added basic JSDoc comments for components
- `MobileNav` uses a simplified version of a menuitem 
- menuitems which used to be hidden are now visible in submenus but disabled until their conditions are met. disabled menuitems remain focusable per [wcag guidelines](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols).

## To Do
- testing:
  - improve `Menubar` tests
  - test with screen readers
  - verify aria structure and role handling
  - test functionality on mobile
- errors:
  - focus is not properly entering certain modals
  - some timing issues with useEffects and blur/focus handlers
  - editor component rerenders when LanguageMenu is mounted
- optimizations:
  - consider refactoring context structure
  - memoize index and menuItem calculations
  - `menuItems` and `menuItemToId` could probably be combined into a better data structure
  - consolidate the more complex state updates with `useReducer`
- other features/enhancements:
  - add support for matching focused menu item to pressed character key
  - improve focus styles 
  - explore options for handling simultaneous styles when hovering over a menuitem and navigating via keyboard
  - tweak hover styles on the "Back to Editor" link (documented below)
  - extract some of the more common logic such as navigation into hooks
  - extract `MenubarSubmenu` components into separate files 
  - extend certain Menubar components to work on mobile (such as `MenubarItem`)
  - improve JSDocs


client/modules/IDE/components/Header/Nav.jsx
tespin, line 1: - Extract logo to header level for better semantic structure 
- Ensure arrow keys only cycle through menu items, excluding the logo

client/components/Menubar/Menubar.jsx
tespin, line 1: Issues:
- originally we were using `useKeyDownHandlers` to handle the `Menubar` keyboard events. This hook captured events globally, which conflicted with keyboard behaviors for the rest of the page, including the editor itself.

Changes:
- Added `e.stopPropagation()` to all handlers
- Attached handlers directly to the `Menubar` container with `role='menubar'`, `aria-orientation='horizontal'`, and `tabIndex='0'`
- removed use of `useKeyDownHandlers`
- added checks for menu and focus state

New issues:
- `role='menubar'` is now in two places: within the `div` in `Menubar.jsx` and in a `ul` within `Nav.jsx`

These changes should make it so that `Menubar` keyboard events are only handled when the actual component or related elements are focused, while preserving the keyboard behaviors for the rest of the page. Instead of using the `useKeyDownHandlers` hook, events are now handled directly by the `Menubar` element with checks for focus and menu state.

We might need to take a second look at the way we compose the `Menubar` component in `Nav.jsx` to resolve the new issues.

client/components/Menubar/Menubar.jsx
```diff
@@ -200,7 +205,20 @@ function Menubar({ children, className }) {
 
   return (
     <MenubarContext.Provider value={contextValue}>
-      <div className={className} ref={nodeRef} onFocus={handleFocus}>
+      <div
```
tespin: Issues:
- originally we were using `useKeyDownHandlers` to handle the `Menubar` keyboard events. This hook captured events globally, which conflicted with keyboard behaviors for the rest of the page, including the editor itself.

Changes:
- Added `e.stopPropagation()` to all handlers
- Attached handlers directly to the `Menubar` container with `role='menubar'`, `aria-orientation='horizontal'`, and `tabIndex='0'`
- removed use of `useKeyDownHandlers`
- added checks for menu and focus state

New issues:
- `role='menubar'` is now in two places: within the div in `Menubar.jsx` and in a `ul` within `Nav.jsx`

These changes should make it so that `Menubar` keyboard events are only handled when the actual component or related elements are focused, while preserving the keyboard behaviors for the rest of the page. Instead of using the `useKeyDownHandlers` hook, events are now handled directly by the `Menubar` element with checks for focus and menu state.

We might need to take a second look at the way we compose the `Menubar` component in `Nav.jsx` to resolve the new issues.

client/components/Menubar/Menubar.jsx
```diff
@@ -1,29 +1,42 @@
...
-  const [activeIndex, setActiveIndex] = useState(-1);
-  const [menuItems, setMenuItems] = useState([]);
-  const [submenuActiveIndex, setSubmenuActiveIndex] = useState(-1);
-  const [submenuItems, setSubmenuItems] = useState([]);
+
+  const menuItems = useRef(new Set()).current;
+  const [activeIndex, setActiveIndex] = useState(0);
+  const prevIndex = usePrevious(activeIndex);
+
+  // old state variables
```
tespin: The previous version of the `Menubar` did not keep track of its menu items. However, we needed a way to manage this collection to implement keyboard navigation. Arrays were useful for basic keyboard features but sets allow for more predictable registration / cleanup operations and guarantee uniqueness. 

Changes:
- array-based management of state renamed to maintain functionality during migration
- added new state variables for managing collections using sets

Keyboard navigation is eventually reimplemented with the new set data structure. The old state variables are cleaned up in a future commit.

tespin:
Adding a video here to see the changes so far --- I'm still testing focus management and evaluating state variables against certain race conditions, but most keyboard behaviors have been implemented. Will slowly work on adding tests 

[Menubar Keyboard Patterns](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/#keyboardinteraction)


https://github.com/user-attachments/assets/48197f7d-40b4-4ce9-8cd4-17d7b77bb29b

client/components/Menubar/Menubar.jsx
```diff
@@ -1,37 +1,213 @@
...
+  useRef,
+  useState,
+  useEffect
+} from 'react';
 import useModalClose from '../../common/useModalClose';
 import { MenuOpenContext, MenubarContext } from './contexts';
+import usePrevious from '../../common/usePrevious';
+
+/**
+ * @component
```
raclim: I love the addition of documentation for the components!

raclim (commented):
Thanks so much for continuing the work on the navigation menu section—I really love the transformation you've been giving it! 

I think overall this PR looks great so far! I feel like there's also a few great practices that are being introduced here (such as the formatting of the PR, component documentation, more detailed commenting) that could be adopted throughout the codebase as well! I also agree that this is a good point to stop at for this PR, and the rest of the tasks outlined in the description could be addressed in future ones. 

I think the only issue I might've found is that I noticed that when using the arrow keys to navigate specifically the **File SubMenu**, it ends up skipping over "Open". 

https://github.com/user-attachments/assets/00839ae8-6bc5-4d21-b58c-0f2fb6d8ee7f

--------------------------------------------------------------------------------------
Besides this, I mostly had a few questions while testing it out and looking through the code, which I'll list out below!

1. When navigating the SubMenus, I noticed that both the mouse hover and arrow key inputs are visible, which I tried to depict in the video. I feel like this could get a bit confusing, but unsure if it might've been intentional!

https://github.com/user-attachments/assets/ec06c327-4c2d-4803-b800-d70c54d20176

2. I also noticed that the focus state immediately appears on the Menubar as soon as I landed on the editor page. I was wondering if this might've been intended as well! 
Screenshot 2025-02-17 at 10 08 18 PM

3. I know there's a few different opinions around adding comments, I was curious about why you decided to add more explanatory comments throughout the components! 

4. While looking through `MenubarSubmenu.jsx`, I feel like it could be broken down into maybe a few files, but was wondering if you felt that way as well!

tespin:
> I think the only issue I might've found is that I noticed that when using the arrow keys to navigate specifically the **File SubMenu**, it ends up skipping over "Open".
>
ooo interesting ... for some reason this didn't happen to me earlier but now it's occurring. it also gets very weird when I save the sketch and new menu items appear? seems like it has something to do with `hideIf` prop. the way I'm tracking the menu items probably isn't accounting for the conditionally rendered options. definitely needs another look.

separately, i also run into an error when the user is unauthenticated and attemps to click "Login" or "Signup", as it's looking for a menuitem element to focus but there isn't one in this case. 
> 
>     1. When navigating the SubMenus, I noticed that both the mouse hover and arrow key inputs are visible, which I tried to depict in the video. I feel like this could get a bit confusing, but unsure if it might've been intentional!
>
definitely not intentional, should probably prioritize one interaction over the other! 
> 
>     2. I also noticed that the focus state immediately appears on the Menubar as soon as I landed on the editor page. I was wondering if this might've been intended as well!
> 
i should fix this as well, the auto-focusing feels disruptive and likely isn't wcag-oriented. 
>  
>     3. I know there's a few different opinions around adding comments, I was curious about why you decided to add more explanatory comments throughout the components!
>
thanks for this question! most comments I added for myself just to keep track of what was happening as it was getting a little complex ... i also felt that it wasn't always clear what some code was doing. i can probably remove some of the comments for code that is more self-evident.
>
>     4. While looking through `MenubarSubmenu.jsx`, I feel like it could be broken down into maybe a few files, but was wondering if you felt that way as well!
yup, agreed! this is a change that makes sense to me.
>

i know it's a lot to review so thanks for taking the time! let me go back in and address some of the more urgent bugs regarding the auto-focusing and weird navigation. as for the other issues (mouse and key inputs being visible, breaking down the submenu into smaller files, etc ...) do you think it's ok if i add those to this PR or can I address them in future ones?

raclim:
> i know it's a lot to review so thanks for taking the time! let me go back in and address some of the more urgent bugs regarding the auto-focusing and weird navigation. as for the other issues (mouse and key inputs being visible, breaking down the submenu into smaller files, etc ...) do you think it's ok if i add those to this PR or can I address them in future ones?

That sounds good to me!! I think the other issues can probably be addressed in future ones, but please feel free to go ahead with whichever feels best to you! Thanks so much again for your efforts on this! :)

tespin:
update: part of the navigation bug looks related to the early return in `MenubarItem` where we return `null` if `hideIf` is true. i moved the registration logic above that and added a check to only register if the `menuitem` is visible. 

this seems to work but breaks again if the user refreshes the page. might need to dig a little deeper into this ...

dipamsen:
> When navigating the SubMenus, I noticed that both the mouse hover and arrow key inputs are visible, which I tried to depict in the video. I feel like this could get a bit confusing, but unsure if it might've been intentional!

Here's one way it could work: https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-editor/

It only ever highlights a single item at a time, which follows the keyboard navigation, but jumps to a different element on mouse hover. Maybe this is not very easy to implement and requires additional logic.

Another option could be to have different styles on hover and focus, since they have different semantics (eg. pressing enter will select the focused option, not the hovered one)

tespin:
@dipamsen thanks for linking the example! that's been the model for this menubar's implementation and is aligned with other examples i've seen -- i'll be using the same approach. 

@raclim spent a little more time with the navigation issue. i think the problem is in the `menuitems` which get conditionally rendered; there seems to be a conflict between when the menuitems are registered and when the conditions for `hideIf` change. `hideIf` tends to change after the component mounts, so when the submenu item eventually registers, it gets added to the end of the `submenuItems` set. even if they are rendered a different way they still get navigated to as if they were the last items. e.x. if a user signs in, saves the sketch, then refreshes the page, the items which depend on `isUnsaved` (`Duplicate`, `Share`, etc) are added at the end of the navigation order. 

i'm curious how you'd feel if, instead of conditionally rendering the items, maybe keeping them visible but disabled instead? that way the ordering of the menuitems can be stable regardless of visibility. i know that the example menubars we looked at above do something similar as well, depending on the context, so it looks like an established pattern. it will probably be an adjustment though, so i just wanted to check in, in case there was an argument for keeping the conditional rendering.

tespin:
here's how things look with the disabled but visible approach. 

**unsaved, unauthenticated**
Screenshot 2025-03-26 at 12 14 00 PM

**unsaved, authenticated**
Screenshot 2025-03-26 at 12 14 48 PM

**saved, authenticated**
Screenshot 2025-03-26 at 12 15 06 PM

i used some existing inactive styles. if we go this route we should probably push it more since it's still hard to tell if an item is disabled or not. navigation works as expected too. we can add tooltips or modals to clarify why an item is disabled.

raclim:
> spent a little more time with the navigation issue. i think the problem is in the menuitems which get conditionally rendered; there seems to be a conflict between when the menuitems are registered and when the conditions for hideIf change. hideIf tends to change after the component mounts, so when the submenu item eventually registers, it gets added to the end of the submenuItems set. even if they are rendered a different way they still get navigated to as if they were the last items. e.x. if a user signs in, saves the sketch, then refreshes the page, the items which depend on isUnsaved (Duplicate, Share, etc) are added at the end of the navigation order.

Thanks so much for catching this—that explains a lot! It makes so much sense when I interact with the menubar now!!! 😂

> i'm curious how you'd feel if, instead of conditionally rendering the items, maybe keeping them visible but disabled instead? that way the ordering of the menuitems can be stable regardless of visibility. i know that the example menubars we looked at above do something similar as well, depending on the context, so it looks like an established pattern. it will probably be an adjustment though, so i just wanted to check in, in case there was an argument for keeping the conditional rendering.
  
Hmm, yeah since disabling the menuitems seems like an established pattern, I'm totally on board with going that route here! I think the conditional rendering was originally used because the component was first built using `toggleDropdown` to control when the menu items would render. 

> i used some existing inactive styles. if we go this route we should probably push it more since it's still hard to tell if an item is disabled or not. navigation works as expected too. we can add tooltips or modals to clarify why an item is disabled.

Thanks so much for linking the screenshots and confirming that it seems to work! I'm wondering do you think it makes sense to break this out into a separate PR, or is it better to keep it within this one?

tespin:
> Thanks so much for linking the screenshots and confirming that it seems to work! I'm wondering do you think it makes sense to break this out into a separate PR, or is it better to keep it within this one?
>
i thought i'd want to break it out, but i think it makes more sense to keep in the current PR since it feels necessary for the component to work. i'll keep the changes small we can open up another PR in the future in case we need to make improvements.

tespin:
@raclim these last changes should put the menubar in better condition! all the menuitems are now visible but ones that require certain conditions to be met are disabled. disabled options are still focusable but cannot be activated as described in the [menubar pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/) and this [guide](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols) on keyboard interfaces. 

i had some state updating issues which are hopefully resolved now but will probably need to be monitored. i didn't address the simultaneous mouse hover / key inputs being visible since that doesn't necessarily impact the component on a functional level; it also appears like there aren't any specific wcag guidelines so we can maybe discuss what feels ideal for the editor environment (the [menubar example](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/) does show both indicators simultaneously though).

otherwise would love for you to check this out when you can! there's a little more clean up left to do and then i just want to finish documenting any other changes / improvements that can be made in future PRs.

raclim:
> all the menuitems are now visible but ones that require certain conditions to be met are disabled. disabled options are still focusable but cannot be activated as described in the [menubar pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/) and this [guide](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#focusabilityofdisabledcontrols) on keyboard interfaces.
> 
> i had some state updating issues which are hopefully resolved now but will probably need to be monitored. i didn't address the simultaneous mouse hover / key inputs being visible since that doesn't necessarily impact the component on a functional level; it also appears like there aren't any specific wcag guidelines so we can maybe discuss what feels ideal for the editor environment (the [menubar example](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/) does show both indicators simultaneously though).

Thanks so much for touching this up, I think this is really awesome and feels much smoother to me! 

I think the only minor tweaks that I'm seeing are styling related, mostly regarding the background when a user hovers over the "Back to Editor" link (pictured below), which I think appears a bit cut off! I'm also wondering this extends to the Upper-right Hand Dropdown menu as well (aslo pictured below), but I think it could also be fine as is!

**"Back to Editor" on Hover**
Screenshot 2025-04-26 at 10 24 39 AM

**Upper-right Hand Dropdown Menu on Hover**
Screenshot 2025-04-26 at 10 24 46 AM

Besides this, I feel overall that it's in a pretty good place so far! Thanks so much again for continuing to work on this and bringing some really exciting changes to the navigation!!

tespin:
ah yeah that is a bit awkward. i'll add to the list for future tasks -- thanks for spotting it! 

doing a bit more clean up and i'll ping again when everything feels ready.

tespin:
@raclim did another pass to clean up what i could ... also, the initial comment here has been updated with recent changes and possible tasks for the future. there are things i may not have caught, but hopefully they'll surface when the changes roll out and we can address them then. 

let me know what you think!

raclim (approved):
Thanks so much for the updates on this!!

I think this overall looks great to me! I think we can probably roll this out if it feels good to you as well!

raclim:
I'm going to merge this in since we're going to start performing some screen reader testing at the end of this week! Please feel free to follow up on anything here before then!

Thanks so much again for all of your patience, time and dedication to revamping the menubar—this is really amazing and I'm really excited to see how folks will engage with it!

tespin:
wooo thanks! can't believe it's finally out there. i'll move on to opening up issues for some of the improvements we brought up. excited for the screen reader tests and will keep an eye out on that as well.

thanks again!

PR #3328, opened 2025-01-28
Fix editor accessibility test

khanniie:
Fixes the editor accessibility issue that appears to happen on the develop branch:
```
 FAIL  client/modules/IDE/components/EditorAccessibility.unit.test.jsx
  ● <EditorAccessibility /> › renders empty message with no lines

    TestingLibraryElementError: Unable to find an accessible element with the role "listitem" and description "There are no lint messages"
```

Changes: The issue (not sure when it was first introduced) is that the "there are no lint messages" list item element does not have an aria-describedby attribute so the original method of fetching by role and description does not actually detect what it's looking for. This switches the test to look for the string instead.


raclim:
Hi! I'm going to close this for now just for organization, but feel free to revisit or reopen this if it feels like something we should come back to!

PR #3331, opened 2025-02-01
Fixed a comment typo in custom-hooks.js

kammeows:
It does not fix any issue, rather just a typo in a comment made in custom-hooks.js


raclim (approved):
Thanks for raising this! This is a great first catch. 

Ideally, we try to accept pull requests that have an associated issue, so for future pull requests please feel free to open an issue first and then attach them to it!

Issue #3339, opened 2025-02-10
Login Page Translation Error: "Invalid Username or Password" Needs Fixing


AR21SM:
### What is your operating system?

Windows

### Actual Behavior

The login page displays the message "Invalid username or password." in English, even though the rest of the interface is in other Language. This message is not translated, which creates an inconsistent user experience.

### Expected Behavior

The error message "Invalid username or password." should be translated to match the language of the rest of the interface for consistency.

### Steps to reproduce

1. Open the login page.
2. Set the interface language to a non-English language.
3. Enter an incorrect username or password.
4. Observe that the error message "Invalid username or password." appears in English, while the rest of the interface is in a different language.

![Image](https://github.com/user-attachments/assets/cb20520c-a455-4417-9b8f-3b019a0305a8)

Jatin24062005:
@AR21SM I have fixed these issues in my pull request. These will be updated soon. Thank you so much for highlighting the issues!
@raclim  Please Review my PR & merge the PR to close the Issue.

raclim:
Thanks for raising this issue! 

This is similar to another issue that's open, which addresses translations on the Sign Up page. Ideally, we want to ensure that we do not rely solely on machine translations, and have native speakers vet and verify them as well. Since there's an open PR with translations already, please feel free to share it around or provide any suggestions on how we could potentially get more eyes on them!

Jatin24062005:
@AR21SM 

<img width="1509" alt="Image" src="https://github.com/user-attachments/assets/52e931b3-ca9b-4155-83a6-e0cc6c108434" />

Here You Can see Issue is fixed now in these PR

developerzohaib786:
I will fix this issue.

PR #3340, opened 2025-02-10
Fix Login Page Translation Error: "Invalid Username or Password"

Jatin24062005:
i Just Add the Translation for the backend Error response in Each language to fix these issue and a translateSubmitError.

Progress towards #issue-#[3339](https://github.com/processing/p5.js-web-editor/issues/3339)

Changes:
i Just Add the Translation for the backend Error response in Each language to fix these issue and a translateSubmitError.


raclim:
Thanks for submitting this!

Since there's a lot of languages here on this pull request, I wanted to go over some options where this could go! 

- **Option 1**: I think I might've noted this already on the associated issue or other translation PRs, but since we try not to rely solely on machine translations and have incoming translations vetted by native speakers before merging them in, we can keep this PR open and add a "Help Wanted" tag to try to get more native speaker's eyes on this. If we go this route, we should update the description for this PR with a list of the languages that this includes, and potentially think of how to get more eyes on this. This process would definitely take a while!

- **Option 2**: Another option is to adjust this PR to only translate the languages that you may speak. We can still create a list of missing translations and add it to the original issue to inform incoming contributors. This approach also lets us merge in the new translations in chunks rather than all at once!

I personally feel that it might be better to go with the second option is possible, but let me know which sounds better to you or if you have any other suggestions, thanks!

Jatin24062005:
@raclim  I appreciate the review! I’ve carefully verified these translations using a dictionary and also took input from my connections on social media and friends who speak these languages. While I understand the need for review, I believe the translations are accurate and good to go.

If I were to submit a PR only for the languages I personally speak (English, Hindi, French, Bengali), it might lead to new issues about missing translations, which could cause more fragmentation and maintenance overhead for the organization.

That said, I’m happy to wait for additional reviews if needed. Let me know how you'd like to proceed!

Jatin24062005:
@raclim should i make changes ??

Harshit-7373:
@raclim I have checked this PR, and I can assure you that the translations for English, Hindi, French, Portuguese, and Korean are correct, as I know only these languages.

raclim:
After giving it some thought on how we want to handle translations, I think we might want to move away from accepting large submissions with a lot of different languages, and gear towards 1-2 languages per submission. Ideally, the 1-2 languages should be ones that the contributor should know and speak natively to ensure that we're getting proper translations. The reason for this shift is that I feel that we should encourage moving away from machine translations, and to have PRs be as ready for submission as possible. 

Although having multiple PRs for a translation might feel fragmented, I think this approach feels steadier and helps with reviewing and organization. Would you be able to update this PR to only include the languages that you speak?

Jatin24062005:
Got it! That makes sense. I'll update the PR to include only the languages I speak and remove the rest.

Jatin24062005:
@raclim  take a look again. Is there is anything else i should?

Jatin24062005:
ughh ... By mistake i just closed it

raclim:
After giving it a closer look, I'm seeing that the original error message for this is coming from the session controller file:  https://github.com/processing/p5.js-web-editor/blob/2bf10dbe2bd58a95cf77434c36528e0d9aae9bdf/server/controllers/session.controller.js#L12

This seems to indicate that the only error message I think we need to be looking out for is the one for "invalid credentials". I removed the translations for `networkErrors` and the extra handling for different error messages in `LoginForm.jsx`, since it seems like there's only one message here, though we can definitely change this if we do have more in the future. 

I think this feels good here, so I'm going to merge this in. Thanks so much for your work in providing the translations and making the adjustments!

raclim (approved):

PR #3358, opened 2025-03-03
Added support for ES Modules in p5.js-web-editor

### LINKED ISSUE #3304: Add import and export as ESM Modules
### Increasing Access

I have many problems just importing a class from other file in my sketch. The problem is that p5.js Web Editor doesnt work like all the JS frameworks or code. Also it doesnt show any messages to solve this, only doesnt work and leave the canvas in blank

### Feature request details

Use import/export like ESM estandard modules.
You can see an example of code that should work in p5.js Web Editor ( it works fine in Webstorm or VS Code), but doent work. You have to delete the export and import keywords to work properly.
Would be nice that p5.js Web Editor would at least show a messge that import/export doesnt work or suggest to change the code.

https://editor.p5js.org/egutie9/sketches/O_xfapq72

takshittt:
Fixes #3304 

Changes: Implemented support for import/export ESM standard modules.

Description: 
Screenshot 2025-03-04 at 4 31 37 AM
This pull request adds support for import/export ESM (ECMAScript Modules) standard modules.

* [x] is descriptively named and links to an issue number, i.e. `Fixes #3304 `


dipamsen:
This is an interesting idea.

I do think supporting ES Modules is a good idea since modules it is a perfectly valid way to write a p5 sketch, and sometimes has advantages over traditional scripts. However, I believe this is not how most users use/will use p5.js, due to many reasons: most p5.js tutorials use scripts instead of modules; p5.js itself is a UMD library - it doesn't expose a ESM distribution (by nature, it will be very hard if not impossible for it to do so).

I have the following questions:

1. Is it needed to have duplicated option for 'New Project' just for projects using ES Modules? I think someone who wants to use ESM to write their code has the know-how to change 3 lines of code from the default starter sketch, to get a 'ESM project'. I don't think this feature has enough benefit, but instead may be confusing to beginners.
2. Your `isModule` check should just be from whether it is linked as `type="module"` or not, checking whether `import` or `export` exists is a weak condition to know whether a file is a module. (eg. the word may match within comments and strings, or in a module there may be no imports or exports)
3. does this work on all types of imports and exports? (ref: [import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import), [export](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export))
4. Does it suffice to just replace the import path with the blob url? If so, a lot of code duplication can be removed.

PR #3375, opened 2025-03-07
Fix: Update results count dynamically in Find window.

### LINKED ISSUE #1859: Results count does not get updated in Find
#### Nature of issue?

- Found a bug

#### Details about the bug: 

- Web browser and version: 88.0.1 (64-bit)
- Operating System: Windows 10

#### Steps to reproduce:

- In the Find window, change the input field and the search options so that the number of occurrences in the editor varies.
**Actual result**: The label with the number of occurrences does not get updated.
**Expected result**: The label should be consistent with the number of highights in the editor.
- Still in the Find window, press Enter in the input field, or click the previous/next arrows
**Resuts**: Now, but only now, the result count gets updated.

Extra confusion comes from the fact that some clickable controls in Find make the count update (the previous/next arrows), but other clickable controls do no (search options like case sensitivity or "only whole words").

Attaching a screen grab to illustrate.

https://user-images.githubusercontent.com/22029901/117534132-741cfa00-aff0-11eb-8d59-8d1ff2d53230.mp4

Harshit-7373:
Fixes #1859 

Changes:

1. Fixed the issue where the results count in the Find window did not update immediately. Replaced redundant getSearchState(cm) calls with state in event listeners to ensure  dynamic results count updates.

### DIFF
--- a/client/utils/codemirror-search.js
+++ b/client/utils/codemirror-search.js
@@ -92,7 +92,7 @@ function persistentDialog(cm, text, deflt, onEnter, replaceOpened, onKeyDown) {
       state.replaceStarted = false;
       if (e.keyCode !== 13 && searchField.value.length > 1) {
         // not enter and more than 1 character to search
-        startSearch(cm, getSearchState(cm), searchField.value);
+        startSearch(cm, state, searchField.value); // Updated
       } else if (searchField.value.length < 1) {
         cm.display.wrapper.querySelector(
           '.CodeMirror-search-results'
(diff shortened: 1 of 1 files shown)


Jatin24062005:
Hey @Harshit-7373,
Since this issue no longer exists—possibly due to a fix in another PR or a recent update—there may be no need to solve it anymore.
if it may exists The issue is related to search, Me, @raclim and @kammeows are already working on it in PR [#3362](https://github.com/processing/p5.js-web-editor/pull/3362). It would be better to discuss and suggest changes there. However, the issue itself doesn’t seem to be present anymore.

Issue #1390, opened 2020-04-12
Enable auto-save as soon as project is started


kerjenwang:
#### Nature of issue?

- New feature request

#### New feature details:

Hi, is it possible to enable auto-save as soon as a user is logged in and starts a new project? Currently I believe the auto-save only enables AFTER the user saves the project for the first time...

Thanks!

catarak:
Thanks for the suggestion! The reasoning behind not starting it immediately is to prevent users from creating a bunch of empty sketches that they don't want. For example, if you opened the editor quickly to test something out but ultimately don't want to save your work. Open to other thoughts though!

It could make sense to turn on autosave if you're opening an existing project and then editing. It won't start autosave then until you've saved it once as well.

kerjenwang:
Hi Cassie,

Thanks for responding!

I see your point. How about enabling auto-save as soon as the project is renamed?

This actually happened to me and I lost a bunch of code. I started a project and renamed it, but I didn't actually explicitly save it because I thought by renaming the project it would automatically save the project under the new name. When the program froze and I closed the browser tab, the code was gone.

Alternatively, whenever a user starts a new project, is it possible to have a pop-up appear to remind the user to save the project after 5 or 10 minutes?

Thanks again!

catarak:
That's a good suggestion. It seems like if you're changing the project name then you probably care about the code you're writing. A reminder popup could also be good—this basically is the approach that CodePen takes, in which the save button will wiggle.

kerjenwang:
Thank you!

debajitdeb11:
@catarak Can I work on this?

catarak:
@debajitdeb11 sure! I don't think there's a 100% clear vision of what this should look like, so feel free to throw out some options before getting to coding. What I do know is, I don't like sketches saving automatically when you first open it, I find that it creates a bunch of junk sketches that I have to wade through the ones that I want. I think, instead, it would make sense to direct users to save if they've been working on a new sketch or editing an existing sketch. This is how CodePen directs users:
<img width="652" alt="Screen Shot 2020-09-03 at 1 19 56 PM" src="https://user-images.githubusercontent.com/6063380/92147578-56581f00-ede9-11ea-8b01-96f2c3db226a.png">

raclim:
To me, this feels related to #1635

PR #3387, opened 2025-03-12
Feature auto save before run

### LINKED ISSUE #1405: Feature request: auto save before run
#### Nature of issue?

- New feature request

#### Details about the bug: 

- Web browser and version: 
- Operating System: 

#### New feature details: Hi! One of the biggest issues I have with web editor is when I accidentally run a faulty code with for example infinite loop and it a) freezes the whole browser tab so the only thing I could find I can do is to reopen my sketch in a new tab and b) as I didn't save my sketch before faulty code launch, I lose a lot of progress and have to rewrite my code. I think it could be better if there was an option to auto save sketch before each launch in settings.

Jatin24062005:
Fixes #1405  #1390 

Changes:
Autosave now starts immediately when an authenticated user edits the default file for the first time.
Added preference check (preferences.autosave) to ensure autosave only starts if enabled.
Implemented first-change tracking using useRef to trigger immediate autosave only once.
Regular autosave (every 5 seconds) remains unchanged after the first edit.
Added a toast notification (Toast.AutosaveStarted) to inform users when autosave is enabled. 

### DIFF
--- a/client/modules/IDE/hooks/useSketchActions.js
+++ b/client/modules/IDE/hooks/useSketchActions.js
@@ -21,13 +21,22 @@ const useSketchActions = () => {
   const dispatch = useDispatch();
   const { t } = useTranslation();
   const params = useParams();
-
+  const preferences = useSelector((state) => state.preferences);
   function newSketch() {
-    if (!unsavedChanges) {
-      dispatch(showToast('Toast.OpenedNewSketch'));
+    dispatch(showToast('Toast.OpenedNewSketch'));
+
+    if (authenticated && preferences.autosave) {
       dispatch(newProject());
-    } else if (window.confirm(t('Nav.WarningUnsavedChanges'))) {
-      dispatch(showToast('Toast.OpenedNewSketch'));
+      dispatch(autosaveProject());
+
+      // Delay the second toast message to prevent overlap
+      setTimeout(() => {
+        dispatch(showToast('Toast.AutosaveEnabled'));
+      }, 1000);
+    } else if (
+      !unsavedChanges ||
+      window.confirm(t('Nav.WarningUnsavedChanges'))
+    ) {
       dispatch(newProject());
     }
   }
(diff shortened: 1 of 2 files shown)


raclim:

I do think this is a good feature to take on, but needs some diving into the prior conversations and maybe some research on how other platforms are integrating it. I think what's currently in this PR might run into the potential issue that was raised where we might be saving too many sketches that the user didn't intend to. 

I personally think the design raised in this comment (https://github.com/processing/p5.js-web-editor/issues/1390#issuecomment-686642767) might be a good approach, since it still signals to the user that they have unsaved changes without necessarily saving it first. 

Another issue we may potentially want to address here is how autosave is implemented and could be communicated better to users. I don't have any ideas here yet, but I think there are a few different opinions on it within the linked issues! I think we might want to dig through those first, and continue the discussion for how this can be implemented there!

Jatin24062005:
I’ve reviewed the concerns, and the current implementation aligns with the expected behavior:

1. Unauthenticated users can not  save their sketches, and this is working as intended.


2. Autosave only triggers if the user is authenticated and has explicitly enabled the autosave preference.


3. Users who don’t want autosave can simply turn it off, and the behavior remains the same as before.

I can Provide you the Video how it working if you want?

raclim:
The concern about saving sketches does apply only to authenticated users, as unauthenticated ones are not allowed to save any sketches without an account. The main concern revolves around saving sketches that a user did not intend to save. In this current setup, this might happen if a user starts editing a sketch. 

The autosave feature is currently also toggled "on" as a default within the user preferences in `server/models/user.js`, so this behavior will probably start as a default.

      autosave: { type: Boolean, default: true },

Although it's true that a user can turn it off, we probably want to avoid having this as a default and ensure that the user actively chooses to save the sketch first. I do think we should gear towards a design similar to the one raised in the linked comment (https://github.com/processing/p5.js-web-editor/issues/1390#issuecomment-686642767).

PR #3389, opened 2025-03-12
Refractor Redux in CollectionList.jsx

### LINKED ISSUE #2042: Refactoring Redux code: hooks, thunks, and redux-toolkit

Harshit-7373:
Fixes #2042

Changes:
1. This PR refactors the CollectionList component to modernize its Redux implementation with React-Redux hooks (useSelector and useDispatch).

### DIFF
--- a/client/modules/IDE/components/CollectionList/CollectionList.jsx
+++ b/client/modules/IDE/components/CollectionList/CollectionList.jsx
@@ -16,6 +16,8 @@ import { SketchSearchbar } from '../Searchbar';
 import CollectionListRow from './CollectionListRow';
 import ArrowUpIcon from '../../../../images/sort-arrow-up.svg';
 import ArrowDownIcon from '../../../../images/sort-arrow-down.svg';
+import * as ProjectsActions from '../../actions/projects';
+import * as ToastActions from '../../actions/toast';
 
 const CollectionList = ({ projectId, username: propsUsername, mobile }) => {
   const dispatch = useDispatch();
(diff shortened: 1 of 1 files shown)


raclim:
Thanks for giving this a shot! 

Based on the attached issue, I think this might be an incomplete solution, since the original issue aims to completely refactor the components to use Redux-Toolkit. I think I'm also seeing an error in the current implementation here with adding sketches to the collection. Because of these factors, I'm going to close this for now!

PR #3391, opened 2025-03-13
Toast notifications now announced properly by screen reader

### LINKED ISSUE #3382: Accessibility: Toast notification don't get announced on screen reader
### p5.js version

_No response_

Windows

### Web browser and version

Google Chrome	133.0.6943.143

### Actual Behavior

The toast notification for some actions (like "sketch saved") doesn't get announced by the screen reader

### Expected Behavior

Since a toast is a notification, it should be announced in a timely manner particularly since the toast will go away. 
Sometimes a toast like "autosave enabled" does get announced. 
I think there is an ARIA live region encapsulating the toast, but since the toast element would disappear from the DOM once visibility is switched off, the live region cannot be announced in due time.

Thoughts on approaching this and similar issues:
1).Set an announcement live region in the page, and update it whenever any status update sent in a toast happens.
2). Increase the duration of the toast might be helpful for giving enough time for the user to finish their action

### Steps:
1. Open a sketch on editor.p5js.org 
2. Login with your credentials
3. Open up a screen reader
4. Edit and save a sketch
5. The toast shows up "Sketch saved", however, it doesn't get announced by the screen reader.
### Snippet:
```js

// Paste your code here :)

```

Harshit-7373:
Fixes #3382

Changes:
1. I updated the time limit for both toast notifications ("Sketch Saved" and "AutoSave Enabled"). Earlier, the "Sketch Saved" notification used to disappear very quickly, within just 2-3 seconds. Now, both notifications remain visible for 10-12 seconds.

2. I added an ARIA live region with the status set to "assertive", ensuring that as soon as a toast notification appears, screen readers announce it immediately.

https://github.com/user-attachments/assets/664f4a3d-4d4f-4395-9764-69513cc9d7c2

This video shows the duration for which the "Sketch Saved" and "AutoSave Enabled" notifications remain visible on the screen.

Screenshot 2025-03-13 at 4 21 19 PM

This image shows the announcement made by the screen reader.

### DIFF
--- a/client/modules/IDE/components/Toast.jsx
+++ b/client/modules/IDE/components/Toast.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { useTranslation } from 'react-i18next';
 import { hideToast } from '../actions/toast';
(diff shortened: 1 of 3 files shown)


dsaw:
I think the issue warrants having a discussion on how the approaches impact usability before embarking on a fix - the approaches listed in the linked ticket are possibilities but all of them have implications. 
The fix has a region with 'assertive" status means that any toast will get announced and potentially interrupt what the user is doing, making it intrusive at times. Actions like 'Saving the sketch' may not be considered that urgent  enough to warrant an intrusive alert. 
Similarly, increasing the duration although a decent option may still be inadequate from a UX perspective as the notification may outlast the action for too long.  Thinking about the use cases and the implications is important from an accessibility standpoint.  @raclim

Harshit-7373:
I have successfully changed the ARIA live region to "polite," ensuring that the screen reader will finish the current speech before announcing the toast message. However, the original duration for "Sketch Saved" was only 3-4 seconds, making it difficult for the screen reader to announce it in time. By the time the screen reader finishes the current speech, the "Sketch Saved" message will have already disappeared. Therefore, I believe increasing the duration is a necessary update to effectively resolve this issue. @raclim @dsaw

client/modules/IDE/components/Toast.jsx
```diff
@@ -9,19 +9,40 @@ export default function Toast() {
...
-      >
-        <ExitIcon focusable="false" aria-hidden="true" />
-      </button>
-    </section>
+    <>
+      {/* Global ARIA Live Region */}
+      <div
+        id="toast-live-region"
+        aria-live="polite"
+        style={{ position: 'absolute', left: '-9999px' }}
```
raclim, line 32: We probably want to preserve the original setup here, such as referencing `toast__close` rather than adding the styling here.

raclim:
Thanks for taking a shot at this! 

I think I'm not entirely sure yet if this might be the approach we want to go with, and agree with @dsaw that we might want to discuss several options for implementing this first (for example, where are all the places in the editor that the toast is called, how long should the toast be up for, etc), and potentially look into how other platforms might be integrating something similar. 

Since this might take a bit more research first, I think it might be best to take a step back before implementing this. I'm going to close this for now—I think there were some good points raised within this PR though, so I think this discussion should be continued within the linked issue!

PR #3395, opened 2025-03-14
Fix navbar dropdown behavior

### LINKED ISSUE #3394: Navbar dropdowns open on hover instead of click
### p5.js version

1.11.1

Windows

### Web browser and version

132.0.6834.160

### Actual Behavior

When clicking on a dropdown menu (e.g., "File"), it opens as expected. However, moving the cursor to another menu item (e.g., "Edit") automatically closes the previously opened dropdown and opens the new one, even without clicking.

Image

In addition, when hovering over a menu item (e.g., "Edit") after another dropdown is open, clicking it for the first time does not open its dropdown. Instead, it just closes the previously opened dropdown. A second click is required to actually open the new dropdown.

Image

### Expected Behavior

A dropdown menu should only open when explicitly clicked. Moving the cursor to another menu item should not trigger its dropdown unless clicked.

Also, clicking on a menu item should immediately open its dropdown, regardless of whether another dropdown was previously open. A second click should not be necessary.

### Steps:
1. Click on the "File" menu to open its dropdown.
2. Move the cursor to the "Edit" menu and hover over it (do not click).
3. Click on "Edit" once.
4. Observe that the "File" dropdown closes, but the "Edit" dropdown does not open.
5. Click "Edit" again, and now the dropdown opens.

kammeows:
Fixes #3394 

Changes:

I have simply removed the onMouseOver function in Menubar to prevent auto-opening.


raclim (approved):
Thanks so much for your work on this and for explaining the issue so thoroughly! After giving this some thought, I think having the dropdown not be visible on hover unless a user intentionally wants to focus on it makes sense to me. 

I'm going to merge this in—@tespin feel free to let me know if there might be some context I might be missing here!

tespin:
hi all! historically, the expected hover behavior for menubars containing dropdown menus is that they do open on hover -- you'll see this if you open up something like google sheets and hover over that menubar. you can also turn to the menubar of your browser (where it says "File", "Edit", etc) for a more immediate example. the limitation is that the open-on-hover behavior should only fire if a dropdown menu is already focused and open. 

the wai-aria guidelines don't prescribe a behavior one way or the other, but they do lay out some [success criteria](https://www.w3.org/WAI/WCAG22/Understanding/content-on-hover-or-focus.html) for cases like ours where content does appear on hover. basically they just say that if hover causes any content changes, the interaction should be predictable. since open-on-hover is widely implemented, i would opt towards leaving it in. 

the issue of requiring a second click to open a new submenu, however, is problematic, and does get addressed in #3320 ! i've been caught up at work but i am hoping to push that out relatively soon.

raclim:
Ohhh I see, thanks so much for the additional context @tespin! In that case I might revert this PR then! 

 I'm sorry that we couldn't get this one in @kammeows but thanks so much again for looking into this and giving it a shot!

PR #3397, opened 2025-03-15
Enhance hover effect on console Clear button  #2592

### LINKED ISSUE #2592: Insufficient hover effect on console Clear button
### p5.js version

v1.8.0

Windows

### Web browser and version

119.0.6045.124

### Actual Behavior

while hovering the cursor over the Clear button the background color is not visible properly 
Screenshot 2023-11-11 143108

### Expected Behavior

we should increase its intensity so that we can see properly

### Steps:
1. Go to p5.js editor
2. Try to hover over the clear button

Harshit-7373:
Fixes #2592 

Changes:

1. Updated the colour of the hover effect for the Clear button in the console in _console.scss file.
2. The colour is mentioned in the theme file for all the 3 (Light , Dark and High contrast themes).

For Light Mode :
Light mode

For Dark Mode : 
Dark mode

For High-Contrast Mode : 
High-Contrast Mode

### DIFF
--- a/client/styles/components/_console.scss
+++ b/client/styles/components/_console.scss
@@ -86,19 +86,23 @@
 
 .preview-console__clear {
 	@include themify() {
-		@extend %link;
-		color: getThemifyVariable('secondary-text-color');
-		&:hover {
-			color: $p5-light-pink; // Use the pink color directly for hover
+	  @extend %link;
+	  color: getThemifyVariable('secondary-text-color');
+	  &:hover {
+		color: $p5-light-pink; 
+  
+		@if (getThemifyVariable('logo-color') == $yellow) {
+		  color: $yellow; 
 		}
+	  }
 	}
 	background: transparent;
 	border: none;
 	padding-right: #{math.div(10, $base-font-size)}rem;
 	.preview-console--collapsed & {
-		display: none;
+	  display: none;
 	}
-}
+  }
 
 .preview-console__body {
 	display: flex;
--- a/client/styles/components/_console.scss
+++ b/client/styles/components/_console.scss
@@ -86,17 +86,20 @@
 
 .preview-console__clear {
 	@include themify() {
-	@extend %link;
-	color: getThemifyVariable('secondary-text-color');
-	&:hover {
-	    color: getThemifyVariable('logo-color');
-	  }
+		@extend %link;
+		color: getThemifyVariable('secondary-text-color');
+
+		&:hover {
+			color: getThemifyVariable('logo-color');
+		}
 	}
+
 	background: transparent;
 	border: none;
 	padding-right: #{math.div(10, $base-font-size)}rem;
+
 	.preview-console--collapsed & {
-	display: none;
+		display: none;
 	}
 }
 
--- a/client/styles/components/_console.scss
+++ b/client/styles/components/_console.scss
@@ -88,16 +88,13 @@
 	@include themify() {
 		@extend %link;
 		color: getThemifyVariable('secondary-text-color');
-
 		&:hover {
 			color: getThemifyVariable('logo-color');
 		}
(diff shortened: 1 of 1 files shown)


Harshit-7373:
@raclim , it was discussed in the comment section of the issue that changing only the text color is the best option.

If you want, I can add a smooth transition similar to how it works for different features in the p5.js web editor. Please share your thoughts on this.

client/styles/components/_console.scss
```diff
@@ -89,7 +89,7 @@
 		@extend %link;
 		color: getThemifyVariable('secondary-text-color');
 		&:hover {
-			color: getThemifyVariable('heavy-text-color');
+			color: $p5-light-pink; // Use the pink color directly for hover
```
raclim: I think this looks good with the `light` and `dark` themes, but we probably want to use `$p5-contrast-yellow` instead for the `contrast` theme. Because we want to adjust the color based on theme, we probably also want to keep using the `getThemifyVariable()` function as well. Could you adjust this line to get the color based on theming? 

Besides this, I think this overall is a step in the right direction!

raclim (changes requested):

client/styles/components/_console.scss
raclim: I might also remove the extra comment as well, since it's mostly self-explanatory.

Harshit-7373:
@raclim I have updated the requested Changes. Please Check them once.

Screenshot 2025-03-18 at 1 38 56 AM
Here is an image , That shows that the required changes are shown correctly.

client/styles/components/_console.scss
Harshit-7373: @raclim I have deleted those Self Explanatory Comments in the updated commit. The updated commit solved the requested change.

Harshit-7373: I also found that there is no (p5-contrast-yellow) in the themes section mentioned in the SCSS variables file. There is only (p5-yellow) present, and I made the changes based on that.

client/styles/components/_console.scss
```diff
@@ -86,22 +86,26 @@
 
 .preview-console__clear {
 	@include themify() {
-		@extend %link;
-		color: getThemifyVariable('secondary-text-color');
-		&:hover {
-			color: getThemifyVariable('heavy-text-color');
+	  @extend %link;
+	  color: getThemifyVariable('secondary-text-color');
+	  &:hover {
```
raclim: Looking at this more closely, rather than adding the conditional `@if (getThemifyVariable('logo-color') == $yellow)`, you can use the `getThemifyVariable` utility function directly—it already takes care of determining the correct theme color for you. The correct code would be: 

```
&:hover {
		color: getThemifyVariable('logo-color');
	  }
```

client/styles/components/_console.scss
```diff
@@ -86,22 +86,26 @@
...
+		  color: $yellow; 
 		}
+	  }
 	}
 	background: transparent;
 	border: none;
 	padding-right: #{math.div(10, $base-font-size)}rem;
 	.preview-console--collapsed & {
-		display: none;
+	  display: none;
```
raclim: Please remove the extra spaces on lines `103` and `105`, and the extra change on line `111`, since they introduce new changes that aren't related to this PR!

raclim (changes requested):
Thanks for the update! After looking a bit more closely, I think it might be best to add a few changes, which I made note of. Let me know if you have any questions about them!

Harshit-7373:
@raclim I have Corrected and Updated the file taking reference of the requested changes.

raclim (approved):
Thanks for making the changes! I made a few extra adjustments to the formatting to ensure minimal changes, so I think this is good to go.

PR #2677, opened 2023-12-03
fix project owner username hover color

PiyushChandra17:
Fixes #2676 

Changes:
* Currently we do not have any styles associated with project owner `username` 
* Added `className` to that link
* Added hover effect 
username-hover

I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] has no test errors (`npm run test`)
* [x] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #3408, opened 2025-03-19
Fix background color issue on hover for sketch name by username #2676

### LINKED ISSUE #2676: Hovering on sketch name by `username` doesn't change the background color to `logo color` in "light" && "dark" mode
### Increasing Access

It will provide better user interface for the client side

Hovering on sketch name by `username` doesn't change the background color to `logo color` in "light" && "dark" mode as it's reproducible in "high contrast" mode

username

Harshit-7373:
Fixes #2676 

Changes:
1. I have updated the classname and the SCSS file to resolve this issue.

Here is a video with the required changes :

https://github.com/user-attachments/assets/1deffb41-c223-4d34-9bee-b03423d7abda

### DIFF
--- a/client/modules/IDE/components/Header/Toolbar.jsx
+++ b/client/modules/IDE/components/Header/Toolbar.jsx
@@ -61,7 +61,6 @@ const Toolbar = (props) => {
       </button>
       <button
         className={playButtonClass}
-        id="play-sketch"
         onClick={() => {
           props.syncFileContent();
           dispatch(startSketch());
@@ -104,7 +103,10 @@ const Toolbar = (props) => {
             return (
               <p className="toolbar__project-project.owner">
                 {t('Toolbar.By')}{' '}
-                <Link to={`/${project.owner.username}/sketches`}>
+                <Link
+                  className="toolbar__username"
+                  to={`/${project.owner.username}/sketches`}
+                >
                   {project.owner.username}
                 </Link>
               </p>
--- a/client/styles/components/_toolbar.scss
+++ b/client/styles/components/_toolbar.scss
@@ -126,6 +126,23 @@
 	}
 }
 
+.toolbar__username {
+	cursor: pointer;
+	@include themify() {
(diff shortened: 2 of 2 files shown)


client/modules/IDE/components/Header/Toolbar.jsx
```diff
@@ -61,7 +61,6 @@ const Toolbar = (props) => {
       </button>
       <button
         className={playButtonClass}
-        id="play-sketch"
```
Jatin24062005, line 64: Why you removed  this???

client/styles/components/_toolbar.scss
```diff
@@ -126,6 +126,23 @@
...
+	cursor: pointer;
+	@include themify() {
+	  color: getThemifyVariable('secondary-text-color');
+	  
+	  &:hover {
+		// Default hover color for light/dark theme
+		color: $p5-light-pink;
+  
+		// If the logo color is yellow, override the hover color
+		@if (getThemifyVariable('logo-color') == $yellow) {
```
raclim, line 139: I think one issue here is similar to the one I noted in [your other PR](https://github.com/processing/p5.js-web-editor/pull/3397#discussion_r1999514842), where we can use just the `getThemifyVariable` utility function directly rather than the conditional.

client/modules/IDE/components/Header/Toolbar.jsx
raclim: Yes, in the context of this PR, I think this should not be removed!

raclim:
Thanks so much for looking into this! 

We actually have a [PR that addresses this issue](https://github.com/processing/p5.js-web-editor/pull/2677), submitted by the same contributor who originally opened it. At the time, I closed it because I felt we needed a clearer, unified design direction for how links should appear before moving forward with any changes.

That said, the updates in that PR are quite similar to what you've proposed here, so I think it makes sense to move ahead with the original one to keep things consistent and acknowledge the initial contribution. With that in mind, I'm going to go ahead and close this for now.

PR #3409, opened 2025-03-20
Fix French and English typos

omahs:
Fix French and English typos


raclim (approved):
Thanks for catching these!

PR #3421, opened 2025-03-26
Fix: Use Title Case for Settings Modal Tab Titles

### LINKED ISSUE #3415: Settings modal should use title case for tab titles
### Increasing Access

Unsure / makes adding additional tabs more consistent with typical UI copy

Currently, the settings modal uses sentence case for tab titles: "General settings."
This should be title case: "General Settings."

priyanjal1234:
Fixes #3415

Updated tab titles in the settings modal to follow title case.

Changed from "General settings" → "General Settings".

### DIFF
--- a/translations/locales/en-US/translations.json
+++ b/translations/locales/en-US/translations.json
@@ -169,7 +169,7 @@
   },
   "Preferences": {
     "Settings": "Settings",
-    "GeneralSettings": "General settings",
+    "GeneralSettings": "General Settings",
     "Accessibility": "Accessibility",
     "Theme": "Theme",
     "LightTheme": "Light",


Jatin24062005:
Hey @priyanjal1234  Thanks for the PR! While the title case change is correct, this update is quite minimal and doesn’t significantly impact the editor’s usability or functionality. It might be better to bundle similar UI/UX consistency improvements together rather than submitting a single PR for such a small change. What do you think?

raclim:
Thanks for sharing your thoughts @Jatin24062005 and for submitting this fix @priyanjal1234! We're aiming to keep changes minimal and focused on the specific issue at hand to make tracking and merging them easier. Since the attached issue is specifically about title casing, the scope of this PR looks good.

That said, the issue has a "Status: On Hold" label, which means it's not ready to be worked on just yet. Since the work has already been complete, we can keep this PR—but for future reference, please hold off on working on issues marked as "On Hold".

raclim (approved):
Thanks so much for your work on this @priyanjal1234!

PR #3455, opened 2025-04-13
Fix: Prevent duplicate file renaming in sidebar

### LINKED ISSUE #3454: Enable file renaming via double-click in the sidebar
### Increasing Access

This feature adds support for renaming files in the p5.js Web Editor by double-clicking on a file name in the sidebar. This mirrors common UX behavior found in most modern code editors (e.g., VS Code, Atom), making the workflow smoother and more intuitive. 

This feature increases access by supporting a more familiar and discoverable UX pattern for users of all experience levels and minimizes reliance on right-click menus, which are less accessible on touch devices or for users who prefer keyboard/tap-based workflows.

It should listen for onDoubleClick on the file name button and trigger the existing rename logic.

kammeows:
Fixes #3454 and #3456 

Changes:
- Imported useSelector from react-redux to access the global files state from the Redux store to read the current list of files and folders in the editor
- Added a function named checkDuplicate that checks if there are any sibling files with same name in the parent folder and returns a boolean value.


kammeows:
@raclim Currently, if a user tries to rename a file to a name that already exists in the same directory, the name simply reverts to the original without any visual feedback. No error message or dialog is shown.
Please let me know if a warning message, inline error text, or a dialog would be preferred to inform the user when renaming isn’t allowed.

raclim (approved):
Thanks for finding these issues and suggesting these fixes for them! I really like the renaming on double-click implementation! 

In terms of giving some sort of visual feedback or warning—I think maybe we can try what you currently have here for now and see how folks respond. Thanks so much again for your work on this!

PR #3492, opened 2025-05-22
Detect browser language to set default language for new users

### LINKED ISSUE #3490: Language Selection via URL Query Parameter
### Increasing Access

Currently, users can set the language of the p5.js editor by using the language dropdown on the upper-left hand navigation menu (pictured below). While language preferences are saved to a user's account, they have to be set manually by the user. 

Image

Adding support for a URL query parameter to set the language would make it easier to share the p5.js editor in different languages. This was initially raised by @SableRaf! 

### Feature request details

**What Would This Look Like?** 
The query parameter to determine the language would be set at the end of a url, such as `https://editor.p5js.org/path?lang=fr`. This would load the editor in French. 

**What Are Some Considerations?** 
- Should visiting a URL with a language parameter also update the user's saved language preference, or only apply for that session?
- What happens if the `lang` value is invalid or unsupported? 

Please feel free to continue fleshing out what this might look like and further considerations that would need to be made here in the comments below!

deveshidwivedi:
Fixes #3490 

Changes: This update improves the user experience for first-time visitors by automatically setting the app language based on their browser's language settings.
Previously, we didn’t have a utility that handled language detection. With this change, the app now checks the user's preferred language and picks the closest supported language available. If the user has already selected a language (like in their saved preferences), that choice is still respected.

### DIFF
--- a/client/utils/language-utils.js
+++ b/client/utils/language-utils.js
@@ -2,21 +2,6 @@
  * Utility functions for language detection and handling
  */
 
-function detectLanguageFromUserAgent(userAgent) {
-  const langRegexes = [
-    /\b([a-z]{2}(-[A-Z]{2})?);/i, // matches patterns like "en;" or "en-US;"
-    /\[([a-z]{2}(-[A-Z]{2})?)\]/i // matches patterns like "[en]" or "[en-US]"
-  ];
-
-  const match = langRegexes.reduce((result, regex) => {
-    if (result) return result;
-    const matches = userAgent.match(regex);
-    return matches && matches[1] ? matches[1] : null;
-  }, null);
-
-  return match;
-}
-
 function getPreferredLanguage(supportedLanguages = [], defaultLanguage = 'en') {
   if (typeof navigator === 'undefined') {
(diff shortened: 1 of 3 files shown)


client/utils/language-utils.js
```diff
@@ -0,0 +1,107 @@
+/**
+ * Utility functions for language detection and handling
+ */
+
+function detectLanguageFromUserAgent(userAgent) {
```
SableRaf: Parsing the `userAgent` seems rather complex. Could we simplify this by using [`navigator.language`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language) and/or [`navigator.languages`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/languages)? cc @raclim

deveshidwivedi: Removed the `userAgent` parsing completely.

raclim (approved):
I tested it out and this seems to look good to me! Thanks so much @deveshidwivedi for your work on this and @SableRaf for your review! :)

PR #3497, opened 2025-05-26
Add tooltip for "Upload File" button when user is not authenticated

### LINKED ISSUE #3079: Notation for the need of login to upload (image) files
### Increasing Access

This feature will allow people to navigate the use of image() and upload file feature easily. More documentation increases access.

Recently, I was working on the p5 web editor without logging in. I was trying to play with image() and trying to upload an image from my computer. However, only after a while I realized I needed to login to have the option to upload.

Either within the web editor or the image() reference page, if there is an indicator that you need to login to use the upload file feature might be helpful.

takshittt:
Fixes #3079

Changes:

- Added a tooltip to the "Upload File" button to inform users that authentication is required before uploading.
- This enhances the user experience by providing guidance instead of leaving the button inactive without explanation.

Screenshot 2025-05-27 at 12 45 46 AM

* [x] is descriptively named and links to an issue number, i.e. `Fixes #3079`

### DIFF
--- a/client/modules/IDE/components/FileNode.jsx
+++ b/client/modules/IDE/components/FileNode.jsx
@@ -343,16 +343,20 @@ const FileNode = ({
                       {t('FileNode.AddFile')}
                     </button>
                   </li>
-                  {authenticated && (
-                    <li>
-                      <button
-                        aria-label={t('FileNode.UploadFileARIA')}
-                        onClick={handleClickUploadFile}
-                      >
-                        {t('FileNode.UploadFile')}
-                      </button>
-                    </li>
-                  )}
+                  <li>
+                    <button
+                      disabled={!authenticated}
+                      className={!authenticated ? 'tooltipped-login' : ''}
+                      aria-label={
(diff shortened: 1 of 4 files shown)


yugalkaushik:
@takshittt Since the "Upload File" button isn’t shown at all when the user is not logged in, adding a tooltip about authentication isn’t needed. Users won’t see the button unless they’re logged in, so the tooltip wouldn’t be visible or helpful.

takshittt:
Hi @yugalkaushik 
Thanks for pointing that out! I’ve actually updated the behavior so that the "Upload File" button is always visible, even when the user is not logged in — it's just disabled in that state. A tooltip or inline message explains that login is required to use it.

This approach improves accessibility and user experience by making the UI more predictable. Users can immediately see what actions are available, and once they log in, the button becomes active without needing to rediscover it.

Let me know if you have any feedback!

raclim:
I'm sorry that this PR ended up falling out of date! Closing since the original issue has been addressed.

PR #3500, opened 2025-05-30
Converts CM5 into CM6

khanniie:
Fixes #issue-number

Changes:


client/modules/IDE/components/Editor/codemirror.js
```diff
@@ -99,118 +65,96 @@ export default function useCodeMirror({
...
+    setCurrentLine(state.doc.lineAt(state.selection.main.head).number);
+
+    if (updateView.docChanged) {
+      debouncedOnChange();
+    }
+  }
+
   // When the container component enters the DOM, we want this function
   // to be called so we can setup the CodeMirror instance with the container.
   function setupCodeMirrorOnContainerMounted(container) {
```
raclim, line 91: I think in the previous version of the file there was a `teardownCodeMirror()` function around line 254 that handled cleanup/removing event listeners. Would it make sense to add a similar cleanup here (or in a separate useEffect hook) and calling something like `cmView.current?.destroy()`? Sorry if I missed something here that's already handling that!

raclim (commented):
Thanks so much for your work on this—I think this overall feels really solid! 

I did some manual testing, and everything seems to function as expected besides the elements that you've already noted as TODOs! I just had a small question in `codemirror.js`, but aside from that that this feels like it's in a great place to me!

client/modules/IDE/components/Editor/codemirror.js
khanniie: omg!!! you're SO right -- great catch! thank you so much!

raclim (approved):
Thanks so much! :)

PR #3540, opened 2025-07-15
pr05 Typescript #2: automatically resolve imports of ts files & migrate instance of client/utils file

clairep94:
# pr05 Typescript Migration: Auto-resolve import of ts files & migrate instance of client/utils file

**IMPORTANT: Should be reviewed after #3533 for clarity**
  - This branch is built off of the branch for #3533 out of necessity, so there are commits that are unrelated to this ticket
  - Relevant commits start at [8b6d4df](https://github.com/processing/p5.js-web-editor/pull/3540/commits/8b6d4df652fd78c05889a9f86311dd63dae0e76f)

## Context:
- Resolve issue with needing to update import extensions (eg. from js to ts), so that ts imports automatically resolve for **both the client folder & tests**
  - Not sure if this resolves imports for the server folder yet -- will resolve this on the PR to migrate an instance of a server file later if so
- Migrate instance of a client/utils file

## Changes:
- [x] Auto-resolve imports for local instance of project client folder
- [x] Auto-resolve imports for jest tests
- [x] Migrate `client/utils/generateRandomName` & associated mock
- [x] Delete `client/utils/isSecurePage` -- checked and this utility function is unused in the project

## I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] has no test errors (`npm run test`)
* [x] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #3553, opened 2025-07-26
pr05 Typescript #3: Migrate client/utils folder

clairep94:
# pr05 Typescript Migration 3: Migrate the client/utils folder

**IMPORTANT: Should be reviewed after #3540 for clarity**
  - This branch is built off of the branch for #3540
  - Relevant commits start at [5b1fd1e](https://github.com/processing/p5.js-web-editor/pull/3554/commits/5b1fd1e43af0080130d256a15b41999d0f3bc9c7)

## Context:
- Rebuild of https://github.com/clairep94/p5.js-web-editor/pull/5 --> forgot to do `git mv` for the first few files and directly renamed, so this looks like deleting and creating a new file for git, and erases the previous history.

- Migrate as much of the client/utils folder with the following steps:
1) `git mv someUtil.js someUtil.ts`. If possible, commit without the `no-verify` flag, otherwise with `no-verify`. 
I forgot to add `no-verify` to my commit message in some of these initial ones
2) Add unit test to secure current behaviour
3) Check for all instances of useage for the utility function to determine types & edgecases of types
4) Add types & handle edgecases & update test
5) Refactor if beneficial
6) Add JSDocs

- I skipped the intellisense autogen files & the codemirror utility files due to unfamiliarity, but migrated most of the rest of the folder

## Changes:
Please see annotations on [files](https://github.com/processing/p5.js-web-editor/pull/3553/files)
- [x] apiClient
- [x] device
- [x] metaKey
- [x] language-utils
- [x] formatDate
- [x] consoleUtils
- [x] dispatcher
- [x] evaluateExpression
- [x] reduxFormUtils
- [x] getConfig (+ many instances of useage throughout the client folder)
- [x] NEW: parseStringToType

## Notes:


client/common/useKeyDownHandlers.js
clairep94, line 1: update to use isMac() utility -- does same thing as the old line 33-34, so removes repeated code

client/modules/IDE/actions/project.js
```diff
@@ -15,7 +15,7 @@ import {
 } from './ide';
 import { clearState, saveState } from '../../../persistState';
 
-const ROOT_URL = getConfig('API_URL');
+const ROOT_URL = getConfig('API_URL', { nullishString: true });
```
clairep94: updated this to return a `''` if no API_URL is found --> used in ln261 as a template literal so it was previously doing
```
const win = window.open(`${undefined}/projects/${projectId}/zip`, '_blank');
```

which would give:
```
window.open(`undefined/projects/${projectId}/zip`, '_blank'))
```

client/modules/IDE/actions/project.js
```diff
@@ -307,6 +307,8 @@ export function cloneProject(project) {
       (file, callback) => {
         if (
           file.url &&
+          S3_BUCKET &&
+          S3_BUCKET_URL_BASE &&
```
clairep94, line 312: Added this check for safety because I think `str.includes(undefined)` runs the same as `str.includes("undefined")`
so 

`"foo undefined bar".includes(undefined); // true`

client/modules/IDE/actions/uploader.js
```diff
@@ -5,9 +5,11 @@ import { handleCreateFile } from './files';
...
 export const s3BucketHttps =
   getConfig('S3_BUCKET_URL_BASE') ||
-  `https://s3-${getConfig('AWS_REGION')}.amazonaws.com/${getConfig(
-    'S3_BUCKET'
-  )}/`;
+  `https://s3-${getConfig('AWS_REGION', {
+    nullishString: true
+  })}.amazonaws.com/${getConfig('S3_BUCKET', {
+    nullishString: true
+  })}/`;
```
clairep94: Same as https://github.com/processing/p5.js-web-editor/pull/3553/files#r2235978386

client/modules/IDE/components/AssetSize.jsx
```diff
@@ -1,10 +1,10 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
 import prettyBytes from 'pretty-bytes';
-
 import getConfig from '../../../utils/getConfig';
+import { parseNumber } from '../../../utils/parseStringToType';
 
-const limit = getConfig('UPLOAD_LIMIT') || 250000000;
+const limit = parseNumber(getConfig('UPLOAD_LIMIT')) || 250000000;
```
clairep94, line 7: Use the new `parseNumber` utility to transform the string from `getConfig` to a number

Currently this file is still jsx so it would be ok to use the string version of this number, but it's not as safe

client/modules/IDE/components/Header/Nav.jsx
clairep94, line 1: Refactor the portions of using `getConfig` for readability instead of inline

use new `parseBoolean` function to parse the string val of `getConfig` to a boolean instead
I wonder if this file was previously only checking if `process.env.LOGIN_ENABLED` had a value??

So if the env file looked like:
```
LOGIN_ENABLED=false
UI_COLLECTIONS_ENABLED=false
```

`getConfig('LOGIN_ENABLED') // "false"`
`!getConfig('LOGIN_ENABLED') // false`??

client/utils/consoleUtils.test.ts
clairep94, line 1: tried my best with this test, I'm not 100% sure how the line offset works, so I didn't test its calculation, just that it should be a number in the tupple that's returned

client/utils/device.ts
clairep94, line 1: PR > Files changed view on Github shows that this is a delete + new file, due to too many changes from the original due to refactoring, so let me know if it would be preferable to go back to the unrefactored!

I refactored for readability & handling  the edgecase of not having a navigator or navigator.userAgent, but that might be excessive

client/utils/dispatcher.ts
clairep94, line 1: Similar to above, GH shows this in PR mode as a deletion + new file, but it seems the history is still retained when you view git blame:
https://github.com/processing/p5.js-web-editor/blame/2362807016746d3b3798085437c2406db2d31757/client/utils/dispatcher.ts

Kept in the comments from a previous maintainer & added JSDocs. I think the number of new lines of jsdocs made it pass the diff tolerated percentage before github shows it as a file deletion instead of a file update

clairep94: No changes other than adding types & no refactors on this one

client/utils/evaluateExpression.test.ts
clairep94, line 1: Not 100% sure if these tests match the function's useage on previewEntry and if there are other cases that need to be added?

client/utils/formatDate.ts
clairep94, line 1: Small refactor for readability + added JSDocs. History is retained despite looking like a delete + new file here
https://github.com/processing/p5.js-web-editor/blame/2362807016746d3b3798085437c2406db2d31757/client/utils/formatDate.ts

Did small refactor to return `''` early if `!parsed`

Also corrected what I thought was a typo here:
https://github.com/processing/p5.js-web-editor/pull/3553/files#diff-e2130c2674d12d8f2724335d2d8d2b7ade8d366efa4bf42717c2af4eec4fa6aaL28

```
const diffInMs = differenceInMilliseconds(now, parsed);

      if (Math.abs(diffInMs < 10000)) {
        return i18next.t('formatDate.JustNow');
      } else if (diffInMs < 20000) {
        return i18next.t('formatDate.15Seconds');
      } else if (diffInMs < 30000) {
        return i18next.t('formatDate.25Seconds');
      } else if (diffInMs < 46000) {
        return i18next.t('formatDate.35Seconds');
      }
```

It seems like they're using `Math.abs` incorrectly --> it's using `Math.abs(true) // 1` and `Math.abs(false) // 0`

I updated to be :
```
const diffInMs = Math.abs(differenceInMilliseconds(now, parsed));

// rest of checks for what range the absolute diff between now and the time something happened falls under
```

client/utils/getConfig.ts
clairep94, line 1: History retained here:
https://github.com/processing/p5.js-web-editor/blame/2362807016746d3b3798085437c2406db2d31757/client/utils/getConfig.ts


Refactors:
- Remove the circular logic at
https://github.com/processing/p5.js-web-editor/pull/3553/files#diff-8fcb54386d32ff76c5399959faeb4e72bad616564277e8774dff9b20d9fa55b3L1-L4

- Made an internal `_getConfig` function that doesn't warn when the value of the key doesn't exist --> use this to check if the environment is test or not
- Update `getConfig` to also take `nullishString` as an option property --> if true, return `''` instead of `undefined` when no value exists.

For when we want a `number` or `boolean` I added a new utility file called `parseStringToType` below
- I was going to do the parsing within `getConfig`, as I did with my proof-of-concept PR from my grant application, but I thought actually keeping the parsing in a separate function keeps things more modular and testable & keeps the return types of `getConfig` consistent for predictability

client/utils/language-utils.ts
clairep94, line 1: History retained here:
https://github.com/processing/p5.js-web-editor/blame/2362807016746d3b3798085437c2406db2d31757/client/utils/language-utils.ts

Create a unit test to secure all behaviour for this function prior to refactor, then did a refactor to make the long for-loop & nested if-statements more readable

clairep94: Previous implementation for contxt:

https://github.com/processing/p5.js-web-editor/pull/3553/files#diff-3500dc8e015b1b12abf5952f3e882a00798c9ce666b6eba2554b96735d5398eeL14-L77

client/utils/metaKey.js
clairep94, line 1: this old implementation uses `navigator.platform` which is deprecated:
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform

I updated this file to use `utils/device` to determine if the client is using a mac

client/utils/parseStringToType.test.ts
clairep94, line 1: New utility functions to parse a string (from getConfig) to a specified type:

for `parseNumber` it returns a number if parseable or undefined/0 depending on the nullishNumber flag

for `parseBoolean` it returns a boolean if parseable or undefined/false depending on the nullishBoolean flag

client/utils/parseStringToType.ts
clairep94, line 1: New utility functions to parse a string (from getConfig) to a specified type:

for parseNumber it returns a number if parseable or undefined/0 depending on the nullishNumber flag

for parseBoolean it returns a boolean if parseable or undefined/false depending on the nullishBoolean flag

client/utils/reduxFormUtils.ts
clairep94, line 1: History retained here:
https://github.com/processing/p5.js-web-editor/blame/2362807016746d3b3798085437c2406db2d31757/client/utils/reduxFormUtils.ts

- Moved the username regex to top of file for readability
- Changed the username length check to `.length` instead of regex for readability: https://github.com/processing/p5.js-web-editor/pull/3553/files#diff-932633409951987d095611a326d0b6b624bd482b07f1e27a9ca7987f6b52a71bR32
- Moved the internal helper functions higher, and grouped the public functions together on the bottom of the file

- Added types -- let me know if this organisation makes sense as this was the most stylistic/open-ended thing I did on the PR
- Also not sure if we want to have a root/types folder, or if it makes sense to keep types collocated where they are most used, and potentially exported, as I did here?

- No other refactors in each function internally

clairep94:
@khanniie @raclim Ready for review! I tried to annotate the files with comments to provide context / explainations for any refactors, but let me know if we should do a sync-review of this one since it's so big

client/modules/IDE/actions/project.js
khanniie: what do we think the "correct" or expected behavior would be if there was no API_URL? like is "/projects/${projectId}/zip" a valid ROOT_URL or would that cause a different error down the line? or in what cases is it undefined?

client/utils/getConfig.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Internal function to retrieve env vars, with no error handling.
+ * @returns String value of env variable or undefined if not found.
+ */
+function _getConfig(key: string): string | undefined {
```
khanniie: FYI i know we don't enforce this very consistently but i dont think we are meant to use underscores in naming, as we "follow" the airbnb guide https://github.com/airbnb/javascript?tab=readme-ov-file#naming--leading-underscore

client/modules/IDE/actions/project.js
khanniie: good call!!

client/utils/parseStringToType.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Parses a string into a number.
+ * - Returns `0` for nullish input if `nullishNumber` is true.
+ * - Returns `undefined` otherwise for nullish or unrecognized input.
+ */
+export function parseNumber(
+  str?: string,
+  nullishNumber = false
+): number | undefined {
+  if (str == null) {
```
khanniie: if we're letting string be typed as str?, then is undefined also a valid nullish value? should it be the more general `if(!str)` ?

client/utils/parseStringToType.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Parses a case-insensitive string into a boolean.
+ * - Returns `false` for nullish input if `nullishBool` is true.
+ * - Returns `undefined` otherwise for nullish or unrecognized input.
+ */
+export function parseBoolean(
+  str?: string,
+  nullishBool = false
+): boolean | undefined {
+  if (str == null) {
```
khanniie: same here

khanniie (commented):
still working through the files but thought I'd sent a first wave!

client/utils/device.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Checks if the user's OS is macOS based on the `navigator.userAgent` string.
+ * This is the preferred method over `navigator.platform`, which is now deprecated:
+ *   - see https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform
+ */
+export function isMac(): boolean {
+  return typeof navigator?.userAgent === 'string'
```
khanniie: do we need `typeof` ? why not something like this:

```
navigator?.userAgent?.toLowerCase().includes('mac') ?? false
```

client/utils/dispatcher.ts
```diff
@@ -0,0 +1,100 @@
+// Inspired by
+// https://github.com/codesandbox/codesandbox-client/blob/master/packages/codesandbox-api/src/dispatcher/index.ts
+
+const frames: {
+  [key: number]: { frame: Window | null, origin: string }
+} = {};
+let frameIndex = 1;
+
+/** Codesandbox dispatcher message types */
```
khanniie: wonder if we can just use typescript enums here instead?

enum MessageTypes = {
START,
STOP, 
etc
}

client/utils/dispatcher.ts
```diff
@@ -0,0 +1,100 @@
...
+export type MessageType = typeof MessageTypes[keyof typeof MessageTypes];
+
+/**
+ * Codesandbox dispatcher message
+ *   - type: 'START', 'STOP' etc
+ *   - payload: additional data for that message type
+ */
+export type Message = {
+  type: MessageType,
+  payload?: any
```
khanniie: is there any way to use `unknown` instead of `any` ?

client/utils/dispatcher.ts
```diff
@@ -0,0 +1,100 @@
...
+  listener = callback;
+  return () => {
+    listener = null;
+  };
+}
+
+function eventListener(e: MessageEvent) {
+  const { data } = e;
+
+  // should also store origin of parent? idk
```
khanniie: i assume this is leftover from the js file but should we remove the dead code comments?

client/utils/evaluateExpression.ts
```diff
@@ -1,11 +1,18 @@
-function __makeEvaluateExpression(evalInClosure) {
-  return (expr) =>
+type EvalResult = {
+  result: any,
+  error: boolean
+};
+
+type EvalInClosureFn = (expr: string) => EvalResult;
+
+function __makeEvaluateExpression(evalInClosure: EvalInClosureFn) {
```
khanniie: avoid using underscores for naming functions

client/utils/evaluateExpression.ts
```diff
@@ -1,11 +1,18 @@
-function __makeEvaluateExpression(evalInClosure) {
-  return (expr) =>
+type EvalResult = {
+  result: any,
```
khanniie: i haven't looked closely enough at where this is used, but is there any way to avoid using `any` here as well? can we do `unknown` and then cast it later after doing some type guard checks? 

https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types

client/utils/evaluateExpression.ts
```diff
@@ -19,7 +26,11 @@ function evaluateExpression() {
       }
       result = (0, eval)(newExpr); // eslint-disable-line
```
khanniie: can we change this eslint disable to be more specific to what it's disabling?

client/utils/formatDate.ts
```diff
@@ -0,0 +1,65 @@
...
+  const date = maybeDate instanceof Date ? maybeDate : parseISO(maybeDate);
+
+  if (isValid(date)) {
+    return date;
+  }
+
+  return null;
+}
+
+export default {
```
khanniie: I'm not sure exactly how this is getting imported but i wonder if we can just export these individually instead of in a dict wrapped as one default? totally understand that it's a holdover from the js file though

khanniie (commented):
still working through the files but have a few more comments!

client/utils/getConfig.ts
```diff
@@ -0,0 +1,50 @@
...
+  const env: Record<string, string | undefined> =
+    (typeof global !== 'undefined' ? global : window)?.process?.env || {};
```
khanniie: why not do `global !== undefined` ?

khanniie: or even 
```
const configSource = global ?? window;
const env = configSource?.process?.env ?? {}
```

client/utils/getConfig.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  return env[key];
+}
+
+type GetConfigOptions = {
```
khanniie: should this be an interface? maybe using optional? 

https://google.github.io/styleguide/tsguide.html#prefer-optional-over-undefined

client/utils/getConfig.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (warn) {
+      console.warn(`getConfig("${key}") returned null or undefined`);
+    }
+    return nullishString ? '' : undefined;
+  }
+
+  return value;
+}
+
+export default getConfig;
```
khanniie: same here can we convert to named export? https://google.github.io/styleguide/tsguide.html#exports

client/utils/language-utils.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * Utility functions for language detection and handling
+ */
+
+function getPreferredLanguage(
+  supportedLanguages: string[] = [],
+  defaultLanguage: string = 'en'
+): string | undefined {
+  if (typeof navigator === 'undefined') {
```
khanniie: same here i think we should be able to do something like if (!navigator) return defaultLanguage

client/utils/language-utils.ts
```diff
@@ -0,0 +1,53 @@
...
+  // Fallback to navigator.language
+  if (navigator.language) {
+    const match = findMatch(navigator.language);
+    if (match) return match;
+  }
+
+  return defaultLanguage;
+}
+
+export default getPreferredLanguage;
```
khanniie: can we put the default export on the function itself here too?

khanniie: or even better... maybe we can make it a named export instead

https://google.github.io/styleguide/tsguide.html#exports

client/utils/reduxFormUtils.ts
```diff
@@ -0,0 +1,156 @@
+import i18n from 'i18next';
+
+/* eslint-disable */
```
khanniie: what are we disabling here? can we be more specific about the rule?

client/utils/reduxFormUtils.ts
```diff
@@ -0,0 +1,156 @@
...
+export type FormErrors = Partial<
+  Email & Username & Password & ConfirmPassword & CurrentPassword & NewPassword
+>;
+
+// === Internal helper functions: =====
+
+/** Processes form & mutates errors to add any `username` & `email` errors */
+function validateUsernameEmail(
+  formProps: Partial<UsernameAndEmail>,
+  errors: Partial<FormErrors>
```
khanniie: I'm not sure i understand what happens when we do Partial<Partial<....>> do we need the double partial?

client/utils/reduxFormUtils.ts
```diff
@@ -0,0 +1,156 @@
...
+
+// ====== PUBLIC: ========
+
+// Account Form:
+export type AccountForm = UsernameAndEmail & CurrentPassword & NewPassword;
+export type AccountFormErrors = Partial<AccountForm>;
+
+/** Validation for the Account Form */
+export function validateSettings(
+  formProps: Partial<AccountForm>
```
khanniie, line 74: if AccountFormErrors is already `Partial<AccountForm>` then can we also just use it for formProps?

client/utils/reduxFormUtils.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+  return errors;
+}
+
+// Login form:
+export type LoginForm = UsernameAndEmail & Password;
+export type LoginFormErrors = Partial<LoginForm>;
+
+/** Validation for the Login Form */
+export function validateLogin(formProps: Partial<LoginForm>): LoginFormErrors {
```
khanniie: same here, why not reuse LoginFormErrors? or rename it to be more general?

client/utils/evaluateExpression.ts
clairep94: fixed!

client/utils/getConfig.ts
clairep94: fixed! I turned the rule override on eslint from `off` to `warn` for this `no-underscore-dangle` rule as well

client/utils/parseStringToType.ts
clairep94: good call! I think I was overcomplicating thinking about if `str` is an empty string, but we would wanna handle that like it was undefined so this `if(!str)` is better

clairep94: fixed

client/utils/parseStringToType.ts
clairep94: fixed

client/utils/device.ts
clairep94: fixed!

client/utils/language-utils.ts
clairep94: Screenshot 2025-08-13 at 22 24 46

I tried updating to `!navigator` but it seemed to cause this error above whereas the `typeof` check seems to circumvent this

khanniie: i actually wonder if they're right that we should be using jsdom as the test environment? it makes sense that `navigator` is a window element in the browser only and thus would be undefined in the default jest testing env : O if that's too high lift though i don't mind switching back. i guess tldr i think the more "correct" solution is to switch the test env to jsdom but typeof seems like a fair hack around it

client/modules/IDE/actions/project.js
```diff
@@ -15,9 +15,11 @@ import {
...
 import { clearState, saveState } from '../../../persistState';
 
-const ROOT_URL = getConfig('API_URL');
-const S3_BUCKET_URL_BASE = getConfig('S3_BUCKET_URL_BASE');
-const S3_BUCKET = getConfig('S3_BUCKET');
+const ROOT_URL = getConfig('API_URL', { throwErrorIfNotFound: true });
+const S3_BUCKET_URL_BASE = getConfig('S3_BUCKET_URL_BASE', {
+  throwErrorIfNotFound: true
+});
+const S3_BUCKET = getConfig('S3_BUCKET', { throwErrorIfNotFound: true });
```
clairep94: @khanniie what do you think of this update?

`getConfig` `options` param now has `throwErrorIfNotFound`, which throws and error if the value is not found (or is an empty string)
This allows the compiler to throw errors immediately when attempting a local build if these things are missing

I've done a search for all instances where `getConfig` is used and made a guess of what I think would be useful to fail at compile-time if the config variable doesn't exist, but this is totally a guess

client/modules/Preview/EmbedFrame.jsx
```diff
@@ -232,9 +232,9 @@ p5.prototype.registerMethod('afterSetup', p5.prototype.ensureAccessibleCanvas);`
   }
 
   const previewScripts = sketchDoc.createElement('script');
-  previewScripts.src = `${window.location.origin}${getConfig(
-    'PREVIEW_SCRIPTS_URL'
-  )}`;
+  previewScripts.src = `${
+    window.location.origin
+  }${getConfig('PREVIEW_SCRIPTS_URL', { nullishString: true })}`;
```
clairep94, line 237: @khanniie I don't have this in my local env, which is auto gen from the env.example, so I assumed this was not fail-on-not-found worthy, what do you think?

client/modules/User/components/CookieConsent.jsx
```diff
@@ -77,6 +77,8 @@ const CookieConsentButtons = styled.div`
   }
 `;
 
+const GOOGLE_ANALYTICS_ID = getConfig('GA_MEASUREMENT_ID');
```
clairep94, line 80: Also don't have this in my local env, so I didn't set it to fail on not-found

client/utils/getConfig.ts
```diff
@@ -0,0 +1,75 @@
...
+  throwErrorInTestEnv?: boolean; // this is only to test getConfig and should never be set to override defaults
+}
+
+export const isTestEnvironment = getEnvVar('NODE_ENV') === 'test';
+
+const defaultGetConfigOptions: GetConfigOptions = {
+  warn: !isTestEnvironment,
+  nullishString: false,
+  throwErrorIfNotFound: false,
+  throwErrorInTestEnv: false
```
clairep94: @khanniie 
Without this, I wasn't able to test if getConfig throws errors when we set `{throwErrorIfNotFound: true}` because the env is 'test' when we run the test

What do you think, is there another way to set the env of the test run for a single test?

Another alternative would be to do `{throwErrorIfNotFound: !isTestEnv}` for whereever we're using `getConfig`, but I was thinking maybe that's not the cleanest to keep doing that when we use the function

client/utils/reduxFormUtils.ts
clairep94: Updated! I tried to just have the form's props & errors to the Partial of each form

- I thought using `AccountFormErrors` as the prop for formProps would look a bit confusing for someone reading the name of the type in the future without checking the type definition

client/utils/language-utils.ts
clairep94: Done!

client/utils/reduxFormUtils.ts
clairep94: Done!

client/utils/getConfig.ts
clairep94: Done!

client/utils/getConfig.ts
clairep94: Done! I did everything in the other PRs as `type` instead of `interface` so I'll go take another look at those

client/utils/getConfig.ts
clairep94: Done!

client/utils/language-utils.ts
clairep94: fixed! the client tests are already using jsdom so I just had to explicitly add that we're accessing navigator from window 

```
const {navigator} = window
```

client/components/SkipLink.tsx
```diff
@@ -3,8 +3,8 @@ import classNames from 'classnames';
 import { useTranslation } from 'react-i18next';
 
 type SkipLinkProps = {
```
khanniie: this should be interface i think?

client/modules/IDE/actions/project.js
khanniie: tysm!! appreciate you making the change! doesn't really affect the implementation but are we sure it's the compiler throwing the error and not the console at runtime? either way i think it's good we're throwing an error instead of letting it fail silently

client/modules/Preview/EmbedFrame.jsx
khanniie: hmmmm tbh I don't really know if it's necessary or not, maybe @raclim would know?

client/modules/User/components/CookieConsent.jsx
khanniie: SGTM!

client/utils/dispatcher.ts
```diff
@@ -0,0 +1,91 @@
...
+  EXECUTE = 'EXECUTE'
+}
+/* eslint-enable no-shadow */
+
+/**
+ * Codesandbox dispatcher message
+ *   - type: 'START', 'STOP' etc
+ *   - payload: additional data for that message type
+ */
+export type Message = {
```
khanniie: interface?

client/utils/evaluateExpression.ts
```diff
@@ -0,0 +1,44 @@
+type EvalResult = {
```
khanniie: interface?

client/utils/getConfig.ts
```diff
@@ -0,0 +1,69 @@
...
+  return env[key];
+}
+
+interface GetConfigOptions {
+  warn?: boolean;
+  nullishString?: boolean;
+  throwErrorIfNotFound?: boolean;
+}
+
+const defaultGetConfigOptions: GetConfigOptions = {
```
khanniie: can we make this constant case? https://google.github.io/styleguide/tsguide.html#identifiers-constants

client/utils/getConfig.ts
khanniie: not sure if i understood the question exactly right, but i think there are ways to set the env for tests! i think you can override the values in the module? 

https://stackoverflow.com/questions/48042200/invalidate-node-cache-when-using-jest

^ not sure if this above works for your case but we can definitely look into mocks / stubs for env if that's what you were wondering

client/modules/IDE/components/Header/Nav.jsx
khanniie: good catch!!

khanniie (dismissed):
it's looking really good!!! I had a few more tiny comments but overall thank you so much for your hard work!

client/modules/IDE/actions/project.js
clairep94: Hmm I just tried this manually by removing API_URL from my local env and it didn't seem to throw in my local build, so I'm going to remove this! 

Will look into doing the thing you mentioned where we have another module that checks for 'required' env variables at compile or build time

clairep94:
@khanniie @raclim Ready for re-review!

khanniie (approved):
thank you so much for all the work!!

client/modules/Preview/EmbedFrame.jsx
raclim: I think it's created in [`server/views/previewIndex.js`](https://github.com/processing/p5.js-web-editor/blob/34123d5d727c6c20ae9a5c18714ecf9107c4ba62/server/views/previewIndex.js#L17) for rendering the preview! I think this is currently okay as is for now

raclim (approved):
Thanks so much for your work on this @clairep94 and for your review @khanniie! I think this overall feels, good and will merge this in!

PR #3565, opened 2025-08-03
pr05 Typescript Migration #5: Migrate client/common folder

clairep94:
# pr05 Typescript Migration 5: Migrate the client/commonfolder

## Context:
- Rebuild of https://github.com/processing/p5.js-web-editor/pull/3564/files for cleaning commits

- Migrate as much of the client/common folder with the following steps:
1) `git mv someComponent.jsx someComponent.tsx`. If possible, commit without the `no-verify` flag, otherwise with `no-verify`. 
2) Add unit test to secure current behaviour
3) Check for all instances of useage for the common component to determine props types
4) Add props types & unit test update
5) Refactor if beneficial -- did minimal refactoring on this one compared to previous PRs and focused on typescript
6) Add JSDocs

## Changes:
- [x] .prettierrc -- remove hardcoded `parser: "babel"` setting
  - This was causing prettier to **always** use babel even though eslint config has overrides for ts and tsx files 
- [x] RouterTab
- [x] Button
- [x] ButtonOrLink
- [x] IconButton
- [ ] icon
  - **was not able to figure out how to migrate this as just updating the file to ts broke the snapshot tests for the generated classes**
- [x] usePrevious
- [x] useKeyDownHandler
  - Had to cherrypick the `isMac()` utility from #3553 
  - Deleted `DocumentKeyDown` -- not used anywhere anymore, likely carry-over/legacy for class components
- [x] onModalClose
- [x] useSyncFormTranslation

Didn't migrate the storebook files as I wasn't able to run storybook locally to verify against silent regression

## Notes:


clairep94:
@raclim @khanniie Ready for review! Can be reviewed separately from the utils folder migration PR

client/common/Button.tsx
```diff
@@ -1,24 +1,94 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import styled from 'styled-components';
-import { Link } from 'react-router-dom';
-
+import { Link, LinkProps } from 'react-router-dom';
 import { remSize, prop } from '../theme';
 
 const kinds = {
```
khanniie: think this could also be a typescript enum

client/common/Button.tsx
```diff
@@ -1,24 +1,94 @@
...
   primary: 'primary',
   secondary: 'secondary'
-};
+} as const;
 
 const displays = {
   block: 'block',
   inline: 'inline'
+} as const;
```
khanniie: same here!

client/common/Button.tsx
```diff
@@ -1,24 +1,94 @@
...
+
+const buttonTypes = {
+  button: 'button',
+  submit: 'submit'
+} as const;
```
khanniie: and here

client/common/Button.tsx
```diff
@@ -1,24 +1,94 @@
...
+type Kind = keyof typeof kinds;
+type Display = keyof typeof displays;
+type ButtonType = keyof typeof buttonTypes;
+
+type StyledButtonProps = {
+  kind: Kind;
+  display: Display;
+};
+
+type SharedButtonProps = {
```
khanniie: can this be an interface instead? same for all other similar cases with the Props

client/common/Button.tsx
```diff
@@ -1,24 +1,94 @@
...
+   */
+  iconOnly?: boolean;
+  /**
+   * The kind of button - determines how it appears visually
+   */
+  kind?: Kind;
+  /**
+   * Specifying an href will use an <a> to link to the URL
+   */
+  href?: string | null;
```
khanniie: feels like if href can be undefined then we should avoid also letting it be null?

client/common/Button.tsx
```diff
@@ -1,24 +1,94 @@
...
+  /**
+   * An ARIA Label used for accessibility
+   */
+  'aria-label'?: string | null;
```
khanniie: same here, prefer to have it either be undefined or null but not both

client/common/Button.tsx
```diff
@@ -1,24 +1,94 @@
...
+  /**
+   * Specifying a to URL will use a react-router Link
+   */
+  to?: string | null;
```
khanniie: and here if possible

client/common/Button.tsx
```diff
@@ -1,24 +1,94 @@
...
+  /**
+   * If using a button, then type is defines the type of button
+   */
+  type?: ButtonType;
+  /**
+   * Allows for IconButton to pass `focusable="false"` as a prop for SVGs.
+   * See @types/react > interface SVGAttributes<T> extends AriaAttributes, DOMAttributes<T>
+   */
+  focusable?: boolean | 'true' | 'false';
```
khanniie: can we have it just be the boolean type? if the parent that is using this component has the information stored as a string we should have it be the parents' responsibility to cast it to a boolean type before it's passed in

client/common/Button.tsx
```diff
@@ -181,69 +249,7 @@ const Button = ({
   );
 };
 
-Button.defaultProps = {
```
khanniie, line 184: it's so great that we can get rid of all of this extra code at the bottom of the React files now haha, thanks for doing this conversion!!

client/common/ButtonOrLink.tsx
```diff
@@ -1,14 +1,47 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
-import PropTypes from 'prop-types';
 
 /**
- * Helper for switching between <button>, <a>, and <Link>
+ * Accepts all the props of an HTML <a> or <button> tag.
  */
+export type ButtonOrLinkProps = {
```
khanniie: can this be an interface?

client/common/Button.tsx
khanniie: https://google.github.io/styleguide/tsguide.html#prefer-interfaces

client/common/IconButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+    width: 100%;
+    height: 100%;
+  }
+`;
+
+export type IconButtonProps = Omit<
+  ButtonProps,
+  'iconBefore' | 'display' | 'focusable'
+> & {
+  icon?: ComponentType<{ 'aria-label'?: string }> | null;
```
khanniie: same here maybe prefer to remove the | null and let the nullish cases all be undefined

client/common/usePrevious.ts
```diff
@@ -0,0 +1,16 @@
+import { useEffect, useRef } from 'react';
+
+/**
+ * Used in Menubar to store the previous value of a number.
+ * @param value - The current value to track.
+ * @returns The previous value before the current render, or undefined if none.
+ */
+export default function usePrevious(value: number): number | undefined {
```
khanniie: is there any way we can avoid having it be typed to numbers specifically? Maybe this is a rare case where it makes sense to use `any` or `unknown`? that way the hook can be used for more general cases in the future

clairep94:
Ready for re-review @khanniie @raclim

client/common/Button.tsx
```diff
@@ -1,33 +1,24 @@
...
-import { Link, LinkProps } from 'react-router-dom';
+import { Link } from 'react-router-dom';
 import { remSize, prop } from '../theme';
 
-const kinds = {
-  primary: 'primary',
-  secondary: 'secondary'
-} as const;
+enum Kinds {
+  primary = 'primary',
```
khanniie: i think enums should use constant case for the fields! https://google.github.io/styleguide/tsguide.html#naming-rules-by-identifier-type

also, i don't think you really need the strings fwiw, there may be a few rare cases where the value itself gets used for some javascript/api logic but if it's used primarily as a categorization method in typescript-only code, i think we can do something like this? 

```
enum Kinds {
   PRIMARY,
   SECONDARY,
}
```

same thing for the others!

client/common/Button.tsx
khanniie: bump!

client/common/Button.tsx
```diff
@@ -249,7 +250,5 @@ const Button = ({
   );
 };
 
-Button.kinds = kinds;
-Button.displays = displays;
-
-export default Button;
+Button.kinds = Kinds;
```
khanniie: feels like instead of doing this, we could just export Kinds and Displays? like

`export enum Kinds`

and then other files can do this:

`import {Kinds as ButtonKinds} from Button `

khanniie (dismissed):
looks great!! thanks so much for all the changes!

client/common/Button.tsx
```diff
@@ -1,28 +1,112 @@
...
-};
+/**
+ * Enum for the visual style of a Button.
+ *
+ * These values transpile to lowercase strings (`'primary' | 'secondary'`)
+ * that map directly to keys in the `Button` object in `theme.js` for styling.
+ */
+export enum ButtonKinds {
+  PRIMARY = 'primary',
+  SECONDARY = 'secondary'
```
clairep94, line 14: @khanniie I had a look at updating this to just

```
export enum ButtonKinds {
  PRIMARY,
  SECONDARY
}
```

But I think theme.js is all stored in lowercase, so I think this might be a good middle ground to allow us to keep theme.js unchanged in lowercase, while using Google TS guidelines for constant casing in uppercase

Screenshot 2025-08-28 at 20 26 28

Screenshot 2025-08-28 at 20 20 27

clairep94: Otherwise we have to cast to lowercase in the part where we make the styled button


Also it allows for a bit of flexibility in case contributors use `<Button type="submit"/>` if they're not fully aware of the TS migration?

clairep94:
@khanniie @raclim 
Just finished addressing the new feedback and ready for re-review!

I think it would be best to finish this one before I open the PR to migrate /client/components, because /client/components ended up being a bit dependent on changes from this one

client/common/Button.tsx
khanniie: ah, no worries at all!! that sounds good to me, thanks for investigating and adding a comment! that's perfect!

khanniie (approved):
Thank you so much for your hard work! Looks great!

raclim (approved):
Thanks so much again for your work on this! Will merge this one and #7 right afterwards!

PR #3594, opened 2025-08-18
Final GSoC Work: Context-Aware Autocomplete, Renaming, and Refactoring Enhancements

kammeows:
This PR completes the GSoC project by introducing context-awareness to the p5.js web editor's autocomplete and refactoring features. It includes:

- Context-Aware Autocomplete
The autocomplete hinter now filters suggestions based on the current code context. It uses a JSON of blacklisted functions and parses the code’s AST to show only relevant suggestions:
image

- User-Defined Code Suggestions
Autocomplete now suggests user-defined variables, functions, and classes, as well as methods for both user-defined and p5 classes.
image
image

- Context-Aware Renaming
Select a word, press F2, and enter the new name in the dialog. All occurrences are renamed correctly according to the context.
image

- Jump to Definition
Hold Ctrl and click a word to jump to where the variable or function is defined.

- Accessibility
All new features are fully accessible to users relying on digital accessibility tools.


.env.example
```diff
@@ -19,7 +19,7 @@ MAILGUN_KEY=<your-mailgun-api-key>
 ML5_LIBRARY_USERNAME=ml5
 ML5_LIBRARY_EMAIL=examples@ml5js.org
 ML5_LIBRARY_PASS=helloml5
-MONGO_URL=mongodb://localhost:27017/p5js-web-editor
+MONGO_URL=mongodb://127.0.0.1:27017/p5js-web-editor
```
diyaayay: I suppose this was committed by mistake.

README_hinter_and_refactoring_changes.md
```diff
@@ -0,0 +1,83 @@
...
+### p5-scope-function-access-map.json
+Purpose: Defines which p5.js functions are allowed or disallowed inside functions like setup, draw, preload, etc.
+
+### p5-reference-functions.json
+Purpose: A flat list of all available p5.js functions.
+
+Used to:
+
+- Differentiate between built-in and user-defined functions
+- Filter out redefinitions or incorrect hints
```
diyaayay, line 83: Let's keep the README as a separate GitHub Gist, and not include it in the final PR to be merged.

diyaayay:
I tested a few things, such as the autocomplete for p5-specific functions and the warnings triggered when functions like `setup` or `draw` are used in undesired places. They work well.
The `Jump to Definition` feature also works fine. I tested it across multiple files, and it correctly jumps to the function declaration in another file when the function is called from there. When multiple functions with the same name exist in different files, it follows the runtime behavior. However, instead of just moving the cursor to the function declaration, it would look visually better if the function name were briefly highlighted, for example:
image

But the functionality works well.

diyaayay:
The context-aware renaming feature works with <kbd>Fn</kbd> + <kbd>F2</kbd> on some systems. I tested it with variable scopes:
```js
function setup() {
  createCanvas(400, 400);
}

var x = 10; 

function draw() {
  var x = 10; // local variable (shadows global x)
  background(220);
  x = 20;     // modifies only the local x
}
```
I tested context-aware renaming and it works for functions within the same file.
However, when I rename a function in `sketch.js` that is defined in another file, the function definition in that other file is not renamed.
Example:
sketch.js
```js
function setup() {
  createCanvas(400, 400);
}

function draw() {
  background(220);
  hello(); // call
}
```
hi.js
```js
function hello() { // definition
  console.log("Hello");
}
```
Renaming `hello()` in sketch.js does not update the function definition in hello.js.
Is this the intended behavior?

kammeows:
Regarding context-aware renaming, yes, that’s the expected behavior. From what I’ve seen in most code editors, the renaming only works within a single file. It won’t rename the same function or variable if it’s defined in another file. It would also make the browser editor's functionality more complicated so I thought it would be best to not include such a feature.

kammeows:
Thank you for your feedback and all your suggestions! I'll also try to add the jump to definition highlight feature.

raclim:
Thanks so much for your work on this @kammeows! 

I think this looks really great so far, and I really love how much smoother this overall looks! I don't think I have too much to add on to @diyaayay's comments, and I feel that they covered most of my initial thoughts while testing out the feature!

I do have a quick question—while loading this branch, I ran into a small version mismatch issue between `eslint` and `eslint-plugin-storybook`, which I resolved by changing these versions. I was wondering, was this something that you faced as well?

kammeows:
Hey @raclim! Thank you for taking a look on this. No I did not encounter any such issue. If it comes up again, we can look into whether we need to adjust the package versions in the repo to prevent others from hitting the same issue.

client/index.jsx
```diff
@@ -53,3 +53,5 @@ render(
   </Provider>,
   document.getElementById('root')
 );
+
+export default store;
```
raclim, line 111: Since it seems like `client/storeInstance.js` is exporting the Redux store, can we remove this line here? I don't think I see any other references to here, but we may also want to update any other references to `store` to point to `storeInstance` instead!

README_hinter_and_refactoring_changes.md
raclim: Agree with @diyaayay here! 

It'd be great if you could move this into the [`contributor_docs` folder](https://github.com/processing/p5.js-web-editor/tree/develop/contributor_docs), and I can update how it's organized and referenced afterwards!

client/modules/IDE/utils/ScreenReaderHelper.jsx
```diff
@@ -0,0 +1,26 @@
+export default function announceToScreenReader(message, assertive = false) {
+  const liveRegion = document.getElementById('rename-aria-live');
+  if (!liveRegion) return;
+
+  // liveRegion.setAttribute('aria-live', assertive ? 'assertive' : 'polite');
```
raclim: It'd be best to remove any extraneous comments, or note why it should be kept commented out!

client/modules/IDE/utils/ScreenReaderHelper.jsx
```diff
@@ -0,0 +1,26 @@
+export default function announceToScreenReader(message, assertive = false) {
```
raclim, line 1: This is a great add and will be so helpful in other parts of the app as well!

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -154,6 +161,16 @@ class Editor extends React.Component {
 
     delete this._cm.options.lint.options.errors;
 
+    this._cm.getWrapperElement().addEventListener('click', (e) => {
+      const isMac = /Mac/.test(navigator.platform);
```
raclim: Is it possible to use this pre-existing [device util](https://github.com/processing/p5.js-web-editor/blob/develop/client/utils/device.ts) here?

raclim (commented):
Thanks so much again for your work on this, this looks really awesome! 🙌 

I was able to give it a closer review and added in a few more comments. I think on top of them, it'd be great if you could organize some of the files within `client/modules/IDE`! 

Some ways you could approach organizing them are grouping them into subfolders within `client/modules/IDE`, storing in other `client` folders (like `client/utils`), or creating new folders (like `client/helpers`). Feel free to go with what you feel might be best, or let me know if you have any questions about it!

client/modules/IDE/components/Editor/index.jsx
kammeows: Is it alright if I changed the function in device util from export function isMac() to export default function isMac() or if you want named export, then should I disable the eslint rule for this file instead?

client/index.jsx
```diff
@@ -43,13 +43,18 @@ const App = () => (
...
-  <Provider store={store}>
-    <ThemeProvider>
-      <Suspense fallback={<Loader />}>
-        <App />
-      </Suspense>
-    </ThemeProvider>
-  </Provider>,
-  document.getElementById('root')
-);
+const rootEl = document.getElementById('root');
```
raclim, line 97: It'd be great to add in an extra comment above just to give some more context as to why `rootEl` was added!

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -172,6 +189,7 @@ class Editor extends React.Component {
       [`Shift-${metaKey}-E`]: (cm) => {
         cm.getInputField().blur();
       },
+      F2: (cm) => this.renameVariable(cm),
```
raclim: I'm sorry I think I'm having some difficulty with triggering this key! Is it possible that it be a Mac/Windows key difference (i.e maybe Mac might need something like `Ctrl` + `F2`)?

raclim: Also would be good to add to the keyboard shortcuts! 
Screenshot 2025-09-10 at 3 10 45 PM

client/modules/IDE/components/Editor/index.jsx
```diff
@@ -209,7 +227,13 @@ class Editor extends React.Component {
     }
 
     this._cm.on('keydown', (_cm, e) => {
-      // Show hint
+      if (
```
raclim, line 233: Maybe this could also have a short commented note above (like Skip handling if it's a paste (Ctrl/V) or any Ctrl/Alt shortcut)? Also okay to keep it as is if you feel that it's more straightforward and might not be needed!

client/utils/ScreenReaderHelper.js
```diff
@@ -0,0 +1,25 @@
+export default function announceToScreenReader(message, assertive = false) {
```
raclim, line 1: I think since this file doesn't contain any `jsx`, it might be better to rename this as a `.js` file!

client/utils/contextAwareHinter.js
```diff
@@ -0,0 +1,190 @@
+/* eslint-disable */
```
raclim: I saw a few files have `eslint` disabled — was there a specific reason for that? If not, I think we could remove the disable and let `eslint` run normally!

client/utils/p5-reference-functions.json
```diff
@@ -0,0 +1,379 @@
+{
+  "functions": {
```
raclim, line 2: Would this list likely be manually updated?

client/utils/renameVariableHelper.jsx
```diff
@@ -0,0 +1,209 @@
+/* eslint-disable */
```
raclim: Similar question about `.jsx` -> `.js` here and `eslint` for this and the next few files!

raclim (commented):
Thanks so much for being open to more feedback, I really appreciate it! Just added in a few questions and thoughts!

client/modules/IDE/components/Editor/index.jsx
raclim: Yes, I think changing the function in the device util to export a default function is alright! I might double check if other files that are calling this function might need to be updated with this change.

client/utils/p5-reference-functions.json
kammeows: So to make this json, I first downloaded this file from https://github.com/processing/p5.js-website/tree/main/src/content/reference/en/p5 and wrote a script to just fetch all the function names from it. I thought it best not to include it in the codebase because it could cause some confusion. Do you think it's best to add it somewhere, maybe in the utils?

raclim (approved):
Going to merge this in! Thanks so much again for all of your hard work on this and excited to see this out soon!

raclim:
Sorry I just realized we spoke about checking in again before merging this in 😅 

Please feel free to check this out in `develop`, and follow up with any new changes/fixes you'd like to make in subsequent PRs before the end of next week! I'll also be doing that as well as a part of our routine release process. Thanks so much again!

PR #3619, opened 2025-09-01
pr05 Typescript Migration #7: migrate client components

clairep94:
# pr05 Typescript Migration 7: Migrate the client/components folder

## Context:
- Rebuild of https://github.com/processing/p5.js-web-editor/pull/3616 after getting stuck on MenubarSubmenu
- Built on top of https://github.com/processing/p5.js-web-editor/pull/3565, should be reviewed after this PR is merged.

- Migrate as much of the client/components folder with the following steps:
1) `git mv someComponent.jsx someComponent.tsx`. If possible, commit without the `no-verify` flag, otherwise with `no-verify`. 
2) Check for all instances of useage for the common component to determine props types
3) Add props types & unit test update
4) Add JSDocs

- did not refactor or add many unit tests.

## Changes:
Client/components:
- [x] Dropdown
  - Saw that only the styled components wrapper was being used and only used once, so migrated the wrapper to Dropdown/DropdownMenu & deleted the file
- [x] Dropdown/DropdownMenu
- [x] Dropdown/MenuItem
- [x] Dropdown/TableDropdown
- [x] Menubar/Menubar
- [x] Menubar/MenubarItem
- [ ] Menubar/MenubarSubmenu -- was not able to figure out the logic of this file. Got stuck on previous PR attempt and abandonned. Will take a look again on a separate PR.
- [x] Menubar/contexts
- [x] PreviewNav
- [x] RootPage
- [x] useAsModal -- deleted as this is unused

Client/common:
- [x] useModalClose
  - update to support generic typing 
- [x]  usePrevious
  - update to support generic typing 

Other:
- any updates to use named exports

## Notes:

I have verified that this pull request:

* [x] has no linting errors (`npm run lint`)
* [x] has no test errors (`npm run test`)
* [x] is from a uniquely-named feature branch and is up to date with the  `develop` branch.
* [x] is descriptively named and links to an issue number, i.e. `Fixes #123`

PR #3623, opened 2025-09-03
pr05 Typescript #8: migrate client/components/Menubar/MenubarSubmenu

clairep94:
# pr05 Typescript Migration 8: Migrate the client/components/Menubar/MenubarSubmenu

## Context:
- Had difficulty migrating/understanding MenubarSubmenu when building PR 7, so chose to open a PR without this file and to resolve it separately.
- Should be reviewed after https://github.com/processing/p5.js-web-editor/pull/3619 is merged for clarity.

## Changes:
- [x] Menubar/MenubarSubmenu
- [x] Menubar/contexts

Other:
- any updates to use named exports

## Notes:


client/components/Menubar/MenubarSubmenu.tsx
```diff
@@ -37,15 +46,28 @@ export function useMenuProps(id) {
  * MenubarTrigger
  * -----------------------------------------------------------------------------------------------*/
 
+/** Custom subset of valid values for aria-hasPopup for the MenubarTrigger */
+enum MenubarTriggerAriaHasPopup {
+  MENU = MenubarListItemRole.MENU,
+  LISTBOX = MenubarListItemRole.LISTBOX,
+  TRUE = 'true'
+}
```
clairep94, line 59: @khanniie I'm not 100% if this was the correct approach

I was a bit confused because the original file had the below:
```
MenubarTrigger.propTypes = {
  role: PropTypes.string,
  hasPopup: PropTypes.oneOf(['menu', 'listbox', 'true'])
};

MenubarList.propTypes = {
  children: PropTypes.node,
  role: PropTypes.oneOf(['menu', 'listbox'])
};
```

I made the assumption the menu bar list item roles and menubar trigger aria-hasPopup are associated in the relationship defined in the enums?? 
Should we just remove TRUE? I didn't see TRUE being used elsewhere

client/components/Menubar/MenubarSubmenu.tsx
```diff
@@ -62,112 +84,113 @@ export function useMenuProps(id) {
...
+      setActiveIndex,
+      menuItems,
+      registerTopLevelItem,
+      hasFocus
+    } = useContext(MenubarContext);
+    const { id, title, first, last } = useContext(SubmenuContext);
+    const { isOpen, handlers } = useMenuProps(id);
+
+    // `ref` is always a button from MenubarSubmenu, so safe to cast.
+    const buttonRef = ref as React.RefObject<HTMLButtonElement>;
```
clairep94, line 111: wasn't sure how to resolve this other than casting, but I thought was safe bc this function is only used within this file once and it's always a button ref

client/components/Menubar/MenubarSubmenu.tsx
```diff
@@ -219,26 +246,26 @@ MenubarList.defaultProps = {
...
  */
-
-function MenubarSubmenu({
+export function MenubarSubmenu({
   children,
   id,
   title,
-  triggerRole: customTriggerRole,
-  listRole: customListRole,
+  triggerRole = 'menuitem',
```
clairep94: not sure if string is too loose here, or if there is a specific enum we should be using?

clairep94:
@raclim @khanniie ready for review! this is build off of PR 7, which was big so might be best to wait till PR 7 is merged to look at this one

client/components/Menubar/MenubarSubmenu.tsx
khanniie: based on the web docs (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-haspopup#description), it looks like they're supposed to be the same!

> "Make sure the role of the element that serves as the container for the   popup content is a [menu](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/menu_role), [listbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role), [tree](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tree_role), [grid](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/grid_role) or [dialog](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role) and that the value of aria-haspopup matches the role of the popup container."

in other words, my understanding is that `aria-haspopup` is telling the assistive device what kind of container the "trigger button" is about to open, so it should match the type of our submenu containers.

so it should be either menu or listbox, like you said -- i guess true is a valid value (docs say "The value true is the same as menu") but if we're never using it, we should remove

client/components/Menubar/MenubarSubmenu.tsx
khanniie: sgtm!! it's kind of surprising to me that typescript wouldn't be able to infer that from the React.forwardRef function params since you gave it the HTMLButtonElement type there but i trust that it doesn't work without casting haha

client/components/Menubar/MenubarSubmenu.tsx
```diff
@@ -18,7 +17,17 @@ import {
 } from './contexts';
 import TriangleIcon from '../../images/down-filled-triangle.svg';
 
-export function useMenuProps(id) {
+/** Custom subset of valid list item roles for the Menubar list items */
+export enum MenubarListItemRole {
```
khanniie: small nit but it looks like these are values for the container, not the items within the container? should we update the comment and interface name?

client/components/Menubar/MenubarSubmenu.tsx
khanniie: defer to you but we could add another enum for list item roles! since you already have one for the list itself. i think the type would be either menuitem (for the 'menu' containers) or option (for the 'listbox' containers)

khanniie (dismissed):
thanks so much for putting this together! it was definitely a very complicated file haha but great work

client/components/Menubar/MenubarSubmenu.tsx
khanniie: maybe like "MenuContainerRole"

raclim: Also agree, I think it might make sense to update the name here!

raclim:
Thanks so much for your work on this! I also think this looks great, and I really like how the refactor for the menubarsubmenu component feels more straightforward and easy to read! Can hold on merging if you'd like to make any further tweaks!

clairep94:
@raclim @khanniie ready for re-review!

client/components/Menubar/MenubarSubmenu.tsx
```diff
@@ -37,15 +52,27 @@ export function useMenuProps(id) {
  * MenubarTrigger
  * -----------------------------------------------------------------------------------------------*/
 
+/** Custom subset of valid values for aria-hasPopup for the MenubarTrigger */
+enum MenubarTriggerAriaHasPopup {
```
khanniie, line 56: this is a bit of a nit but if it looks exactly the same as menucontainerrole why not just use that instead of duplicating?

khanniie (approved):
one small comment but overall LGTM! thank you so much!

client/components/Menubar/MenubarSubmenu.tsx
raclim: I think this might be ok, since semantically they're a bit different even though the values are the same, and might give room to add more diff values later? But also feel its ok if it might be better to reuse!

khanniie: sgtm! i don't have a strong opinion, happy to just get it in!

PR #3641, opened 2025-09-11
feat: fixed word wrap preference issue

### LINKED ISSUE #3527: word wrap off still wraps
### p5.js version

n/a

Windows

### Web browser and version

firefox 139 / chrome 138

### Actual Behavior

the word wrap setting always wraps the code text

### Expected Behavior

when word wrap is off - don't wrap

### Steps to reproduce

1. settings => word wrap = off
2. type long text
3. observes it still wraps

Paulie-Aditya:
Fixes #3527 

Changes:
State variable was linewrap but was getting referenced as wordwrap, so even though the dispatch function was working, the state wasn't getting updated

* [ ✅] has no linting errors (`npm run lint`)
* [✅ ] has no test errors (`npm run test`)
* [✅ ] is descriptively named and links to an issue number, i.e. `Fixes #123`


Paulie-Aditya:
Working as expected now:


https://github.com/user-attachments/assets/f5db0eda-87e1-4b47-b4d1-76c651e1d26d

raclim (changes requested):
Thanks so much for your work on this, I think this overall feels good!

It seems like the only issue here might be that the `Preferences.unit.test.jsx` might need to be updated to align with the changes here in the `Preferences` component. 

I do also think it's a bit confusing to have both `linewrap` and `wordwrap` within the repository—we may want to stick to one term in the future! Maybe after this is merged in, there could be another issue for this!

Paulie-Aditya:
> Thanks so much for your work on this, I think this overall feels good!
> 
> It seems like the only issue here might be that the `Preferences.unit.test.jsx` might need to be updated to align with the changes here in the `Preferences` component.
> 
> I do also think it's a bit confusing to have both `linewrap` and `wordwrap` within the repository—we may want to stick to one term in the future! Maybe after this is merged in, there could be another issue for this!

Hey @raclim ! in the `Preferences.unit.test.jsx` , I already see it referencing setLineWrap, which in fact is the correct convention according to the codebase, just in case there's an error from my side, do let me know and I will do the required changes, but it looks good from my end.

Edit: the function is correct but it is referenced as wordwrap, I have changed it to linewrap everywhere in the same file for the time being, in case of any changes, do let me know

raclim (dismissed):
That sounds good to me! Thanks so much for following up with these changes so swiftly! 

Will merge this in before the end of the week!

Paulie-Aditya:
Great! Thanks!

raclim (approved):
Sorry for the extra change here! Since the tests didn't pass, I updated the naming in places that refer to the UI components. “Wordwrap” is still used in our translations, and updating the translation itself will likely need further discussion. I also added a TODO about consolidating terminology later. Since it’s a small change, I went ahead and made the updates and will merge this in!

PR #3672, opened 2025-09-27
pr05 Typescript Migration #13: Migrate the User Model (& API Key schema) 

clairep94:
Migration of the user model to ts
- made some assumptions/work arounds to solve type errors
- This is a best effor, I will attempt the user controller next -- I expect that that attempt may result in tweaks to this file afterwards
  - I actually originally attempted to migrate the entire user system (controller, model, routes) but this was too difficult to do without previous PRs being finalised and merged. Please see [here](https://github.com/clairep94/p5.js-web-editor/pull/13/files) for that attempt
- Added a test for the user model before migrating, so I could ensure nothing breaks
- All tests pass & all typechecks pass

## Changes:
- [x] Move apikey schema to its own file
- [x] Add test for user model
- [x] Extract types from user and api key
  - Note: my original approach was to use mongoose `inferSchemaType` and other infers but this was honestly really hard to debug and it was easier to explicity define what we expect from the types and feed it to the mongoose scheme definition, and attach the extra Document methods
- [x] Migrate apiKey and user to ts & to named export


server/models/user.ts
```diff
@@ -79,13 +43,13 @@ const userSchema = new Schema(
...
-      enum: ['none', 'essential', 'all'],
-      default: 'none'
+      enum: Object.values(CookieConsentOptions),
+      default: CookieConsentOptions.NONE
     },
     banned: { type: Boolean, default: false },
     lastLoginTimestamp: { type: Date }
   },
-  { timestamps: true, usePushEach: true }
+  { timestamps: true }
```
khanniie, line 52: we don't need usePushEach anymore? is it set to that by default? just wanted to make sure it didn't slip through the cracks!

khanniie (dismissed):

server/models/user.ts
raclim: Thanks for catching this! It was used for older mongo versions, and is no longer needed past v5, so i think we might be good here!

clairep94: Hi guys sorry for the delayed response! Yess once I migrated this file to ts, `usePushEach` was highlighting in red as it was deprecated for the current version of mongo, so I think we're meant to delete it

raclim (approved):
Merging this in!

Issue #3694, opened 2025-10-20
Builds Failing: Missing TypeScript Transpiling in Examples Config


raclim:
## Description
The Docker/CI build started failing during the build:examples step with the following error:

```
Module parse failed: Identifier 'UserModel' has already been declared
You may need an appropriate loader to handle this file type
```

This is probably caused by `webpack/config.examples.js`, which currently only transpiles `.js` files. Building the examples imports `server/models/user.ts`, so when `webpack` attempted to parse TypeScript, it caused a syntax error.

## Steps to Reproduce
Run `npm run build` in the command line, which will show an error.

mangit955:
I also investigated this issue and arrived at the similar fix by updating the webpack/config.examples.js to include Typescript transpilling support. Here is my pr for reference: https://github.com/processing/p5.js-web-editor/pull/3698 
Happy to close it if existing PR #3696 already resolves this

raclim:
Thanks so much for your interest in this issue @mangit955 ! This has been resolved by the earlier PR—I'm sorry that we couldn't get yours in, but please feel free to check out our other issues!

Issue #16, opened 2016-07-21
⌘T for reformat and ⌘R map to run


bensk:
Thanks!
👍

catarak:
The comments are now working. For reference, the CodePen shortcuts.
<img width="601" alt="screen shot 2016-07-21 at 3 42 42 pm" src="https://cloud.githubusercontent.com/assets/6063380/17036321/d1d6e0f4-4f59-11e6-8372-11ebb1118c91.png">
Since ⌘ + T and ⌘ + R are basic browser shortcuts, maybe we want to use other shortcuts? Open to thoughts.

cc @shiffman @lmccart

catarak:
Also considering: Command + Enter to Run, Command + Shift + Enter to Stop, as this is a standard keyboard shortcut.

lmccart:
I agree, lets avoid ⌘ + T and ⌘ + R. What about ctrl + R, etc? Or is that refresh on windows?

shiffman:
I like command enter!   So maybe. . . .

⌘ENTER --> RUN
⌘⇧ENTER --> STOP
⇧TAB --> AUTO-FORMAT

(@lmccart, yes, ctrl+r is refresh on windows.)

catarak:
This is _almost_ done, just need to add the keyboard shortcuts to the keyboard shortcut modal

kaganjd:
Gah, just seeing this now! Sorry bout that, I'll do it today.

catarak:
no worries, I also forgot to tell you

PR #3696, opened 2025-10-22
pr05 Typescript Migration #16: Fix build error webpack example config

clairep94:
Fixes #issue-number #3694

Changes:
- update regex in `webpack/config.examples.js` to match ts & tsx files (previously only js and jsx)

Before:

Screenshot 2025-10-22 at 14 38 02

After:

Screenshot 2025-10-22 at 14 38 23


webpack/config.examples.js
```diff
@@ -22,7 +22,7 @@ module.exports = [
     module: {
       rules: [
         {
-          test: /\.js$/,
+          test: /\.[jt]sx?$/,
```
clairep94, line 25: main change to solve the issue, screenshot in description was taken after just this change

include tsx just in case?

I updated the rest of the `babel-loader` regex patterns to be this pattern to include tsx jsx just in case so it's a bit more flexible even if the fetching script is from the server. happy to revert if stricter is preferred @raclim

raclim: Thanks so much! This sounds good to me! If needed, we can follow up with any changes!

raclim (approved):
Thanks so much again for following up on this, will get this in and take a look at the other PR!

Issue #20, opened 2016-07-21
Use p5.js or p5.min.js in index.html


shiffman:
I noticed that the minified versions of the library are referenced in `index.html`.   I think we should probably use the full versions for things like the friendly error messages, etc.  (This can eventually be preference with an editable template, etc.)

catarak:
In retrospect this is obvious... but I did not realize the friendly errors aren't in the minified versions. Finally switching this out.

catarak:
also I don't know how to test whether or not this is working... help? @shiffman

lmccart:
I removed the friendly errors from the most recent version of p5.js because they need another pass to make them consistent and work well. So I think it makes sense for us to go with p5.min.js for the time being.

PR #3704, opened 2025-10-27
Pr05 Typescript Migration #20: Refine server/types & client/reducer types

clairep94:
## Context:
While doing the rough-work PR below, I realised that can be some refinements on the types previously defined. I also figured out how to manually test my local server via Postman, so I logged some things I wasn't sure about the shape of 

(eg. What user-related type is attached to the `Express.user`? What is the shape of the `user` state in the `redux store`? Is it the same as the server?)

This also helped identify some areas where there are multiple sources of truth of `user` related properties in the client, which I've noted in comments, but have not addressed in this PR to not include logic changes.

## Changes:
- [x] Add `GetRootState` type --> used to define the `getState` param of `redux` `actions
- [x] Migrate `client/modules/User/actions`
- [x] Migrate `client/modules/User/reducers`
- [x] Update `PublicUser.apiKeys` to `SanitisedApiKeys[]` -- previously `DocumentArray<ApiKeyDocument>`, but they should not include `hashedKey` and should be POJO 
  - [x] `UserDocuement.apiKeys` still has `DocumentArray<ApiKeyDocument>`
  - [x] Update testUtils > `createMockUser` & its callers to reflect this

`user.apiKeys` is displayed in the client below & shows it's the `SanitisedApiKey` type without `hashedKey`:

Screenshot 2025-10-27 at 12 47 55

Please see #3705 for actual migration of the client/modules/User > API key stuff

## Notes:
Cherry-picked relevant commits from [the giant rough-work PR here](https://github.com/clairep94/p5.js-web-editor/pull/23) for easier review

## Checks:


client/modules/User/actions.ts
```diff
@@ -1,79 +1,124 @@
...
+/**
+ * - Method: `POST`
+ * - Endpoint: `/signup`
+ * - Authenticated: `false`
+ * - Id: `UserController.createUser`
+ *
+ * Description:
+ *   - Create a new user
+ */
+export function signUpUser(formValues: CreateUserRequestBody) {
```
clairep94, line 44: copying these JSDocs from the relevant server user controller method, so we get the benefit of the hints on hover in the client as well

Only did this for the `user.controller` methods, so there are some methods below which are in the `server/passport` controller or otherwise that haven't been migrated yet

client/modules/User/reducers.ts
```diff
@@ -1,6 +1,27 @@
...
+
+export const user = (
+  state: Partial<PublicUser> & {
+    authenticated: boolean;
+    // TODO: use state of user from server as single source of truth:
+    // Currently using redux state below, but server also has similar info.
+    resetPasswordInitiate?: boolean;
+    resetPasswordInvalid?: boolean;
+    emailVerificationInitiate?: boolean;
+    emailVerificationTokenState?: 'checking' | 'verified' | 'invalid';
```
clairep94, line 19: Not sure if we might want to open an issue where we migrate to use just the server `PublicUser.verified` as the server has that, but the `redux` `store.user` has these additional properties?

Not sure if the logic above makes sense, as I haven't dug deeper into it with a proposal yet but just wanted to highlight

The server `PublicUser.verified` enum is the below, so there's a mismatch:
```
/** Possible email confirmation states */
export const EmailConfirmationStates = {
  Verified: 'verified',
  Sent: 'sent',
  Resent: 'resent'
} as const;
```

server/controllers/user.controller/__testUtils__.ts
```diff
@@ -29,7 +29,7 @@ export const mockBaseUserSanitised: PublicUser = {
   email: 'test@example.com',
   username: 'tester',
   preferences: mockUserPreferences,
-  apiKeys: ([] as unknown) as Types.DocumentArray<ApiKeyDocument>,
+  apiKeys: [],
```
clairep94, line 32: for a `PublicUser`, apiKeys is `SanitisedApiKey`[]

server/controllers/user.controller/__testUtils__.ts
```diff
@@ -42,6 +42,7 @@ export const mockBaseUserSanitised: PublicUser = {
 export const mockBaseUserFull: Omit<User, 'createdAt'> = {
   ...mockBaseUserSanitised,
   name: 'test user',
+  apiKeys: ([] as unknown) as Types.DocumentArray<ApiKeyDocument>,
```
clairep94, line 45: for a `UserDocument`, apiKeys is `DocumentArray<ApiKeyDocument>` -- It needs access to methods like `apiKeys.pull`

server/controllers/user.controller/__testUtils__.ts
```diff
@@ -58,9 +59,15 @@ export const mockBaseUserFull: Omit<User, 'createdAt'> = {
...
+      ...mockBaseUserFull,
+      ...overrides
+    } as UserDocument;
+  }
   return {
-    ...(unSanitised ? mockBaseUserFull : mockBaseUserSanitised),
+    ...mockBaseUserSanitised,
     ...overrides
-  };
+  } as PublicUser;
```
clairep94, line 72: I tried to make the return type of each case super explicit, but we seem to still need to cast the return value of calling this function anyway to resolve type-errors for when the usecase requires `UserDocument` methods

server/controllers/user.controller/helpers.ts
```diff
@@ -1,19 +1,33 @@
...
  * @returns Sanitised user
  */
-export function userResponse(user: PublicUser | UserDocument): PublicUser {
+export function userResponse(user: UserDocument): PublicUser {
   return {
     email: user.email,
     username: user.username,
     preferences: user.preferences,
-    apiKeys: user.apiKeys,
+    apiKeys: user.apiKeys.map((el) => sanitiseApiKey(el)),
```
clairep94, line 30: This is one of the bigger changes on this PR:
After manual testing/logging what userResponse gives us on Postman, I saw that apiKeys is the `SanitisedApiKey`, with the last element containing `token`

server/types/apiKey.ts
```diff
@@ -8,22 +8,26 @@ export interface IApiKey extends VirtualId, MongooseTimestamps {
...
 }
 
 /** Mongoose document object for API Key */
 export interface ApiKeyDocument
   extends IApiKey,
     Omit<Document<Types.ObjectId>, 'id'> {
-  toJSON(options?: any): SanitisedApiKey;
-  toObject(options?: any): SanitisedApiKey;
+  toJSON(options?: any): IApiKey;
+  toObject(options?: any): IApiKey;
```
clairep94, line 19: Previously typed this as `SanitisedApiKey`, but it's the full ApiKey till it gets sanitised

clairep94:
@raclim @khanniie

client/modules/User/reducers.ts
```diff
@@ -1,6 +1,27 @@
+import type { CookieConsentOptions, PublicUser } from '../../../common/types';
 import * as ActionTypes from '../../constants';
 
-const user = (state = { authenticated: false }, action) => {
+// User Action:
+export type UserAction = {
```
khanniie, line 5: should this be interface? : O

client/modules/User/reducers.ts
khanniie: ahhh i think i understand what you mean... definitely worth digging into, if you want to make an issue like you said!

khanniie (approved):
thanks so much!!

client/modules/User/reducers.ts
raclim: Thanks for catching this! I agree that it makes sense to do some cleanup here using `PublicUser.verified`as the source of truth and keeping UI-specific flags on the client side. I'm happy to open an issue for this over the following week, or feel free to do so if you already have context!

raclim (approved):
Thanks so much @clairep94 for taking the time to comb through your past PRs, refine the types, and catch the mismatch on the server and client side's verification state and @khanniie for your review!

PR #3731, opened 2025-11-16
fix: update deprecated codemirror import to restore Storybook build

rohith-1205:
### Summary
Storybook’s CSS build was failing because Sass no longer supports the ~ tilde import syntax. 

### Problem
The file `client/styles/main.scss` still uses the old pattern:
@import '~codemirror/lib/codemirror';

This causes Sass to throw:
"Can't find stylesheet to import"

### Fix
Updated the SCSS import to the modern path:
@import 'codemirror/lib/codemirror.css';

This resolves the Storybook build issue and improves contributor experience.

No breaking changes.

### DIFF
--- a/client/styles/main.scss
+++ b/client/styles/main.scss
@@ -6,7 +6,7 @@
 @import 'base/reset';
 @import 'base/base';
 
-@import '~codemirror/lib/codemirror';
+@import 'codemirror/lib/codemirror.css';
 @import '~codemirror/addon/lint/lint';
 @import '~codemirror-colorpicker/addon/codemirror-colorpicker';
 @import '~dropzone/dist/dropzone';


rohith-1205:
Here is Storybook running after applying the SCSS import fix. 
(Note: The ThemeProvider issue is unrelated and exists on main branch too.)

WhatsApp Image 2025-11-16 at 4 24 51 PM

raclim:
Thanks for your work on this! 

At the moment, we've been upgrading CodeMirror altogether from v5 to v6 in the [`develop-codemirror-v6`](https://github.com/processing/p5.js-web-editor/tree/develop-codemirror-v6) branch, which I believe is set to address this issue. We generally encourage contributors to ensure that each PR is associated with an issue to avoid extra work or miscommunication, which I think is what might've happened here. I'm sorry that we couldn't get this in, but please feel free to check out some of our other open issues!

PR #3770, opened 2025-12-30
fix: resolve 502 error on project download (Fixes #3746)

### LINKED ISSUE #3746: Unable to download game
### p5.js version

1.11.1

Windows

### Web browser and version

142.0.7444.147 (Official Build) (64-bit) 

### Actual Behavior

when I click the  download button it will load for about a minute before loading this page (server location blocked for privacy) this also happens on my Chromebook running chrome. I have been able to almost instantaneously download in the past. 

Image

### Expected Behavior

download the  zipfile for my game to my harddrive

### Steps:
1. attempt to download file 2.it fails 
### Snippet:
```js

// Paste your code here :)
https://editor.p5js.org/SLDTyp0/sketches/cC4r8IR1Z```

### PR DESCRIPTION
**Pull request by Nixxx19**
## Problem
The `downloadProjectAsZip` function was calling `buildZip` without awaiting it, causing a 502 Bad Gateway error when downloading projects. This led to:
- Unhandled promise rejections if `buildZip` threw errors
- No timeout on S3 file fetches, causing requests to hang indefinitely
- Race conditions when clients disconnected before completion

## Changes Made
- **Added `await` to `buildZip` call** in `downloadProjectAsZip` to ensure errors are properly caught and the function waits for completion before responding
- **Added 30-second timeout** to axios requests when fetching files from S3 to prevent hanging requests
- **Improved error handling** by checking `res.headersSent` before sending error responses to avoid crashes on client disconnects
- **Enhanced error logging** with more descriptive error messages for better debugging

## Testing

### DIFF
--- a/server/controllers/project.controller.js
+++ b/server/controllers/project.controller.js
@@ -234,10 +234,12 @@ async function addFileToZip(file, files, zip) {
   } else if (file.url) {
     try {
       const res = await axios.get(file.url, {
-        responseType: 'arraybuffer'
+        responseType: 'arraybuffer',
+        timeout: 30000 // 30 second timeout to prevent hanging requests
       });
       zip.file(file.name, res.data);
     } catch (e) {
+      console.warn(`Failed to fetch file from ${file.url}:`, e.message);
       zip.file(file.name, new ArrayBuffer(0));
     }
   } else {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Nixxx19** [comment]
https://github.com/user-attachments/assets/af255472-2a0d-4da8-938e-fd3b270d0041

## Video Demonstration

I've included a video showing the download functionality working correctly after the fix:

✅ **Before the fix:** Downloads would hang for ~1 minute and then show a 502 Bad Gateway error
✅ **After the fix:** Downloads complete successfully and the zip file is generated properly

The video shows:
- A p5.js sketch being downloaded
- The download completing successfully
- The zip file (`clock_2025_12_30_16_13_37.zip`, 5.2 MB) being saved without errors

This confirms that the changes resolve issue #3746.

File: server/controllers/project.controller.js [hunk 1]
```diff
@@ -234,10 +234,12 @@ async function addFileToZip(file, files, zip) {
   } else if (file.url) {
     try {
       const res = await axios.get(file.url, {
-        responseType: 'arraybuffer'
+        responseType: 'arraybuffer',
```
Jatin24062005: This won't help it will force  make wait to backend not the cloudflare service.

Jatin24062005:
@Nixxx19 According to your video did u able to set the custom path of zip file in your machine????

Jatin24062005:
@Nixxx19 Thanks for working on this and adding timeouts to prevent hanging requests — that’s a good safety improvement for the backend.

However, this change does not fully address the 502 Bad Gateway issue observed when downloading projects.

The root cause appears to be that ZIP generation (including remote asset fetching and JSZip.generateAsync) is performed synchronously before any response headers are sent. When this process takes longer than Cloudflare’s gateway timeout, Cloudflare terminates the request and returns a 502 — even though the backend may still be working.

Adding axios timeouts helps avoid infinite waits inside the Node process, but it does not prevent Cloudflare from timing out while waiting for the origin to start responding.

Suggested improvement:

Send response headers immediately and stream the ZIP output (generateNodeStream) instead of buffering the entire archive in memory.

Alternatively, move ZIP generation to an async/background job and return a download link once ready.

Starting the response early ensures Cloudflare keeps the connection open and avoids gateway errors, while also reducing memory usage on the server. 
Try to keep engaging Cloudflare rather than timeout What you thinks @raclim ?

Nixxx19:
@Jatin24062005, Thanks for the detailed feedback.

The primary goal of this PR is to fix the unhandled promise and request hanging by properly awaiting buildZip and adding timeouts, which resolves the 502 issue in the scenarios I tested (as shown in the demo).

I agree that streaming the zip or offloading generation to a background job could be a more robust long-term solution, but that would be a larger architectural change. This PR focuses on a minimal, safe fix for the current bug.

raclim:
Thanks for looking into this @Nixxx19 and for your review @Jatin24062005! 

I think this helps improve error handling and timeouts, but also agree that it might not fully resolve the issue yet and we might want to move towards a longer term solution, which could look like streaming the zip, changing the file limit sizes, etc. 

I'll update this as progress towards the issue since I think a lot of the changes here are beneficial, and we can keep the original issue open.

raclim (approved):

Issue #3779, opened 2026-01-08
Update URL version parsing to handle non-string entries in p5Versions


lirenjie95:
### p5.js version

2.1.2 and 1.11.11

### What is your operating system?

Windows

### Web browser and version

Firefox 146.0.1(64-bit)

### Actual Behavior

common/p5Versions.js now mixes objects into p5Versions (lines 7-16), but some consumers still assume p5Versions is a string[]; e.g., client/utils/parseURLParams.js uses p5Versions.includes(ver) and then calls v.split('.') inside filter, which will both fail to recognize object-wrapped versions like 2.1.1/1.11.11 and can throw TypeError: v.split is not a function when ?version=1.11 or ?version=1 is provided (since the filter iterates over the objects).

### Expected Behavior

If p5Versions is intended to stay heterogeneous, consider exporting/using a normalized string list for validation and matching in the URL parser (and any similar callers). Suggestion: we can just have a very simple check like this:
```JavaScript
const versions = p5Versions.map((v) => (typeof v === 'string' ? v : v.version));
```

### Steps to reproduce

It was introduced in #3525

Shra-05:
hi,
I’d like to work on this issue if it’s still available.
My understanding is that p5Versions now contains both strings and objects, but some consumers assume it’s a string array and call methods like includes and split, which causes runtime errors.
I plan to normalize p5Versions into a string-only list (using v.version when needed) before validation in the URL parsing logic.
Please let me know if that approach sounds good before I start.

Thanks!

lirenjie95:
> I plan to normalize p5Versions into a string-only list (using v.version when needed) before validation in the URL parsing logic. 

Sounds correct to me. Can you only use string-only list for related functions but keep the original p5Versions array? Thanks!

Shra-05:
Perfect, thanks! 🚀  
I’ll start working on this using a string-only derived list for the relevant URL parsing functions while keeping p5Versions unchanged.

I’ll share a PR shortly. Happy to take the assignment if that helps 🙂

lirenjie95:
@raclim can you please provide any suggestion and assign this task to @Shra-05 ? Thanks!

soham555-maker:
Hi @lirenjie95 and @raclim 👋

I’ve reproduced the issue locally (/?version=1.11, /?version=1) and confirmed the v.split is not a function error occurs due to object entries in p5Versions.

[...]

PR #3801, opened 2026-01-17
Fix and Migrate Version URL Parser

### LINKED ISSUE #3761: Migrate Version URL Parser to TypeScript
### Increasing Access

This migration supports the ongoing TypeScript conversion effort, making the codebase more approachable for new contributors through type safety, autocomplete, and self-documenting function signatures.

#### Files to Migrate

- `client/utils/parseURLParams.js`
- `client/utils/parseURLParams.test.js`

**Context:** The files handle URL parameter parsing for p5.js version selection and library toggles.

#### Known Type Improvements Needed

1. `validateVersion(version)` should accept `string | null`
2. `validateBool(value, defaultValue)` should accept `value: string | null, defaultValue: boolean`
3. `parseUrlParams(url)` should accept `string` and return type should be an exported interface
4. `getNewestVersion(versions)` should enforce `string[]` parameter and return `string`

#### Success Criteria

- Files compile without errors
- All existing tests pass
- No implicit `any` types

### LINKED ISSUE #3762: URL version parsing fails for major/major.minor patterns
### p5.js version

_No response_

Windows

### Web browser and version

_No response_

### Actual Behavior

Fails to load when trying to pass in major/major.minor patterns. (`?version=1` and `?version=2.1`)

Loads default version when given `?version=2.1.2`.

This occurs because the `p5Versions` array in `common/p5Versions.js` now contains objects rather than simply strings. The parser was not made to handle this and needs to be updated.

### Expected Behavior

The most recent version that matches the major or major.minor should successfully be set up when valid prefixes are inputted. Should default to default version when invalid version is passed in.

As of December 18, 2025:

- `?version=2.1` -> 2.1.2
- `?version=1` -> 1.11.11
- `?version=9` -> 1.11.11

### Steps to reproduce

Try to configure version using either the major or major.minor pattern in the URL. I also put some examples below.

Example URLs:

- https://editor.p5js.org?version=2
- https://editor.p5js.org?version=2.1
- https://editor.p5js.org?version=2.1.2 (This one will open up with default version instead)
- https://editor.p5js.org?version=1
- https://editor.p5js.org?version=9
- https://editor.p5js.org?version=0.5

Hugo-RM:
Fixes #3761
Fixes #3762

### Changes

- Fixes bug where p5Versions containing objects caused URL version parsing to fail
- Migrates parseUrlParams and test file to TypeScript

### Testing

- Added tests for version patterns
- Verified invalid inputs properly fall back to defaults
- Tests use regex assertions and validity checks to ensure only valid versions are returned

### Testing Strategy Note

The version parsing tests use regex patterns (e.g., `/^1\.\d+\.\d+$/`) rather than hardcoded expected versions (e.g., `'1.11.11'`). This approach:

- Validates that the correct version prefix is returned
- Verifies the result is a valid version from `p5VersionStrings`
- Avoids test failures when new p5.js versions are released

**Alternative approach:** Tests could assert exact versions (e.g., `expect(result.version).toBe('1.11.11')`), which would catch sorting bugs but require updates whenever `p5Versions` changes.

Sorting bugs are very unlikely to occur unless version numbers change drastically. Current tests simply attempt to prevent future silent errors.

I'm happy to switch to hardcoded assertions or explore other testing strategies if you have suggestions!


client/utils/parseURLParams.ts
```diff
@@ -1,18 +1,29 @@
...
- * @returns {string} The highest version from the array
- */
-function getNewestVersion(versions) {
+function getVersionString(
+  item: string | { version: string; label: string }
+): string {
+  return typeof item === 'string' ? item : item.version;
+}
+
+export const p5VersionStrings = p5Versions.map(getVersionString);
```
clairep94: Is this being called anywhere else? Otherwise we can probably keep it local and not export it

client/utils/parseURLParams.test.ts
clairep94, line 1: Nice! thanks for doing this

client/utils/parseURLParams.ts
```diff
@@ -72,7 +83,7 @@ function validateBool(value, defaultValue) {
   return defaultValue;
 }
 
-export function parseUrlParams(url) {
+export function parseUrlParams(url: string): ParsedUrlParams {
```
clairep94, line 92: Could we add JSDocs for what this function is for?

clairep94:
Looks great, thanks for migrating this! Just some small comments 🙏

client/utils/parseURLParams.test.ts
```diff
@@ -0,0 +1,118 @@
...
+  describe('version parsing', () => {
+    // Uses regex since p5Versions may be updated over time.
+    // Checks to ensure version is valid too.
+
+    test('parses a valid p5 version and falls back for invalid versions', () => {
+      const good = parseUrlParams('https://example.com?version=1.4.0');
+      expect(good.version).toBe('1.4.0');
+
+      const bad = parseUrlParams('https://example.com?version=9.9.9');
+      expect(bad.version).toBe(currentP5Version);
```
lirenjie95: The invalid-version test relies on '9.9.9' never becoming a real release. Once p5 publishes that tag, the test will report a failure even though the fallback logic works. Using an obviously invalid token (e.g., 'invalid-version' or a value derived from p5Versions that is guaranteed to be absent) would make the check resilient to future releases.

client/utils/parseURLParams.ts
```diff
@@ -1,18 +1,29 @@
...
 
-/**
- * Sorts version strings in descending order and returns the highest version
- * @param {string[]} versions - Array of version strings (e.g., ['1.11.2', '1.11.1'])
- * @returns {string} The highest version from the array
- */
-function getNewestVersion(versions) {
+function getVersionString(
+  item: string | { version: string; label: string }
+): string {
```
lirenjie95, line 20: Looks like it can also resolve #3779. Can you help double check it?

client/utils/parseURLParams.ts
Hugo-RM: It's only used in the `parseURLParams.test.ts` file. I can move it to be local if preferred. It's a small bit of code so I wasn't sure if I should avoid code duplication or not.

client/utils/parseURLParams.ts
Hugo-RM: Will do so.

client/utils/parseURLParams.test.ts
Hugo-RM: Good point. I didn't think about future releases breaking this. I'll update it to use 'invalid-version'.

client/utils/parseURLParams.ts
Hugo-RM: Yes the issue is resolved by this.

lirenjie95 (approved):
LGTM

ghost (approved):
NICE

clairep94 (approved):

PR #3805, opened 2026-01-18
Unhandled exceptions disclosures potential sensitive information

kitlord:
Changes:
Adds a global error handler to prevent potential sensitive information leak upon unhandled exceptions.

@raclim Please review

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -195,6 +195,20 @@ app.get('*', async (req, res) => {
   res.type('txt').send('Not found.');
 });
 
+// Global error handler for unhandled errors
+app.use((error, req, res, next) => {
+  console.error('Unhandled error:', error);
+
+  if (res.headersSent) {
+    return next(error);
+  }
+
+  const statusCode = error.status || 500;
+  return res.status(statusCode).json({
+    error: 'Internal server error'
+  });
+});
+
 // start app
 app.listen(process.env.PORT, (error) => {
   if (!error) {


yugalkaushik:
@Iron-56 Can you please provide a scenario or instance where this global handler is required or an example of a sensitive information leak that currently occurs without this middleware.

kitlord:
> @Iron-56 Can you please provide a scenario or instance where this global handler is required or an example of a sensitive information leak that currently occurs without this middleware.

I'm afraid I can’t share reproduction steps or concrete exploit scenarios publicly, since they would expose sensitive information.
I’m happy to provide full details privately to the maintainers when they are available.

server/server.js
```diff
@@ -195,6 +195,20 @@ app.get('*', async (req, res) => {
...
 
+// Global error handler for unhandled errors
+app.use((error, req, res, next) => {
+  console.error('Unhandled error:', error);
+
+  if (res.headersSent) {
+    return next(error);
+  }
+
+  const statusCode = error.code || 500;
```
raclim: Thanks for looking into this! Just wanted to confirm, will `error.code` be a string in this case (I'm assuming it will be similar to the error codes listed in the [Node docs](https://nodejs.org/docs/latest-v18.x/api/all.html#all_errors_nodejs-error-codes))? If so, then I think it'll throw an invalid status code error with `res.status(statusCode).json`.

kitlord: I fixed it, it was supposed to be error.status, to send the status code. Thank you for catching it!

raclim (approved):
Thanks for the update, this looks good to me!

PR #3858, opened 2026-02-10
Fix MongoDB config and remove unnecessary DB dependency from preview server

siddocode:
This PR fixes MongoDB configuration inconsistencies and removes an unnecessary
database dependency from the preview server.

### Changes
- Standardize MongoDB configuration using `MONGO_URI`
- Maintain backward compatibility with `MONGO_URL`
- Update session store to use a single Mongo connection source
- Remove MongoDB connection from preview server (not required at runtime)
- Align `.env.example` with actual runtime expectations

### Why
On a fresh local setup, the preview server attempted to connect to MongoDB using
credentials that may not exist, causing the application to crash. The preview
server does not require database access, so removing this dependency simplifies
local development and prevents startup crashes.

### Testing
- Tested locally with Docker MongoDB
- Verified editor and preview servers start successfully


kitlord:
Hey I tested and it works out, but doesn't changing the cookie lifetime affect the editor itself?

siddocode:
> Hey I tested and it works out, but doesn't changing the cookie lifetime affect the editor itself?

The cookie maxAge remains unchanged (4 weeks), so session duration and editor behavior are unaffected.

I removed the explicit ttl option from MongoStore.create() because connect-mongo automatically derives TTL from the session cookie's maxAge.

This avoids redundant configuration while keeping the same session lifetime behavior.

kitlord:
@raclim could you merge the PR? Sorry for the late review, got busy with some work.

PR #3860, opened 2026-02-11
Fix inconsistent click area for “Back to Editor” button on login and signup pages

siddocode:
## Problem

On the Login and Signup pages, the “Back to Editor” button was only clickable when the cursor was directly over the text or icon.

However, on the main editor view, the entire visual button area was clickable.

This resulted in inconsistent and unintuitive interaction behavior across different views.

## Root Cause

The `.nav__back-link` element did not occupy the full interactive area on auth pages, causing the click target to be limited to inline content instead of the full button region.

## Fix

- Updated `.nav__back-link` styles to ensure the entire visual area is clickable.
- Ensured consistent behavior across:
  - Editor view
  - Login view
  - Signup view
- Updated related Nav snapshots.
- Improved `EditorAccessibility` test query to use `getByText` for robustness.

## Testing

- All client tests passing (225/225)
-  Snapshots updated intentionally
-  Manually verified behavior across affected pages

Before: 

https://github.com/user-attachments/assets/90002cd3-270f-4464-82b1-2eebe804bc41

After:

https://github.com/user-attachments/assets/2ebc4be7-235c-4b73-83de-275ad8f52716

### DIFF
--- a/common/p5Versions.js
+++ b/common/p5Versions.js
@@ -5,7 +5,8 @@ export const currentP5Version = '1.11.11'; // Don't update to 2.x until 2026
 // JSON.stringify([...document.querySelectorAll('._132722c7')].map(n => n.innerText), null, 2)
 // TODO: use their API for this to grab these at build time?
 export const p5Versions = [
-  { version: '2.2.0', label: '(Beta)' },
+  { version: '2.2.1', label: '(Beta)' },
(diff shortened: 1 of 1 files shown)


yugalkaushik:
@sid2487 Could you please review the PR again, recheck the files and push only the files that are required for this PR?

PR #3897, opened 2026-02-16
Fix: use async bcrypt in findMatchingKey #3878

### LINKED ISSUE #3878: Synchronous `bcrypt.compareSync()` used with `await` in async `findMatchingKey` method
### p5.js version

latest

None

### Web browser and version

all

### Actual Behavior

In `server/models/user.ts` at line 176, the `findMatchingKey` method uses `bcrypt.compareSync()` (synchronous) with `await` inside an `async` function:

```javascript
userSchema.methods.findMatchingKey = async function findMatchingKey(
  candidateKey: string
): Promise<{ isMatch: boolean; keyDocument: ApiKeyDocument | null }> {
  for (const k of this.apiKeys) {
    try {
      const foundOne = await bcrypt.compareSync(candidateKey, k.hashedKey);
      // ❌ compareSync is synchronous — await does nothing
```

Two problems:
1. `bcrypt.compareSync()` blocks the Node.js event loop while computing the hash comparison. For each API key checked, the entire server is unresponsive during that time.
2. `await` on a non-Promise (boolean) return value has no effect — it resolves immediately but gives a false impression that the operation is asynchronous.

### Expected Behavior

Use the asynchronous `bcrypt.compare()` which returns a Promise and does not block the event loop:

```javascript
const foundOne = await bcrypt.compare(candidateKey, k.hashedKey);
// ✅ async version — does not block the event loop
```

### Steps to reproduce

1. Open `server/models/user.ts`
2. Go to line 176 inside the `findMatchingKey` method
3. Observe `bcrypt.compareSync()` is used with `await`
4. Under concurrent load with multiple API key authentication requests, the synchronous hash comparison blocks the event loop, degrading server responsiveness for all users

Nixxx19:
Fixes #3878

## Changes

Replaced `bcrypt.compareSync()` with `bcrypt.compare()` in the `findMatchingKey` method to prevent blocking the Node.js event loop during API key authentication.

**Before:**
- Used synchronous `bcrypt.compareSync()` which blocks the event loop
- `await` on a boolean return value had no effect

**After:**
- Uses asynchronous `bcrypt.compare()` which returns a Promise
- Properly awaits the async operation, keeping the server responsive under concurrent load

**Why this is better:**

- **Non-blocking**: `bcrypt.compare()` runs asynchronously, so API key verification no longer blocks the Node.js event loop during hash comparisons
- **Correct async semantics**: `await` now waits on a real Promise instead of a boolean, matching the intent of the `async` method signature
- **Better performance under load**: Under concurrent API key authentication requests, the server remains responsive since comparisons don't serialize the event loop

* [x] is descriptively named and links to an issue number, i.e. `Fixes #3878`


raclim (approved):
Thanks so much!

Nixxx19:
> Thanks so much!

no worries!

PR #3939, opened 2026-02-25
Set session cookie secure flag based on NODE_ENV

### LINKED ISSUE #3909: Session cookie not enforced secure in production
### p5.js version

_No response_

None

### Web browser and version

_No response_

### Actual Behavior

Session cookie is set with `secure: false` hardcoded in `server/server.js`. Session cookies can be sent over HTTP, increasing session hijacking risk if the app is (or is ever) served over HTTP in production.

**Location:** `server/server.js` line 94

### Expected Behavior

In production (`NODE_ENV === 'production'`), the session cookie should have `secure: true` so it is only sent over HTTPS.

### Steps to reproduce

1. Open `server/server.js` and find the session `cookie` config.
2. Confirm `secure: false` is set with no environment-based override.
3. In production over HTTPS, cookie could still be sent over HTTP if there is any downgrade or misconfiguration.

### Snippet:

```js
// server.js
cookie: {
  httpOnly: true,
  secure: false,   // should be true in production
  maxAge: 1000 * 60 * 60 * 24 * 28
}

Nishthajain7:
### Issue:
Fixes #3909 

### Problem
The session cookie was currently configured with `secure: false`
### Changes:
Update the session configuration to conditionally enable the secure flag when running in production.


yugalkaushik (approved):
@ksen0 the changes look good. This is a great and clean fix that keeps development environments unaffected while properly improving session cookie security. Even though production realistically should never run over plain HTTP, it is still great to enforce secure: true in production as an additional safety measure. Nice work.

ksen0 (approved):

raclim:
Thanks so much for your work on this @Nishthajain7 and for your review @yugalkaushik!

After doing some testing in our staging environment in preparation for our next release, I noticed that there was an issue where login succeeds but its session would not persist, where all subsequent authenticated requests (i.e saving sketches, updating preferences, account details) would return a `403` response because the session user is missing. 

Looking into it, I saw that the response headers for `/login` weren't setting the `sessionId` cookie in staging, but would appear in my local environment, which I believe is causing the requests after login to fail authentication. I think this might've been introduced by changing `secure: isProduction`, since staging's `NODE_ENV` is also `production`, but am not entirely sure if this is the case yet. 

I also noticed that this issue doesn't happen in our ReleaseHub environments, which also run in production mode. I think this might indicate that the issue might be related to the deployment infrastructure or request handling—for example, maybe our current CloudFlare SSL configuration could be impacting it? 

I'm going to revert this change for now to see if it restores the previous behavior and if so, unblock the release. If this does resolve the issue in staging, I'll follow up and confirm!

raclim:
Confirming that the issue where the login sessions would not persist no longer appears on staging. I think we would ideally want to integrate the changes here—I reopened the original issue, so would be down to revisit this once CodeMirror 6 is released, which will be sometime next week!

PR #3952, opened 2026-02-28
fix(auth): prevent crash when Google OAuth email is missing 

### LINKED ISSUE #3907: Google OAuth: missing email array validation
### p5.js version

_No response_

None

### Web browser and version

_No response_

### Actual Behavior

Google strategy accesses `profile._json.emails[0]` without checking that `emails` exists and has length. If Google does not return emails, this throws `TypeError: Cannot read property '0' of undefined` and crashes the OAuth flow.

**Location:** `server/config/passport.js` Google Strategy, lines 244, 261, 265

### Expected Behavior

Before using `profile._json.emails[0]`, code should verify `profile._json?.emails` exists and has at least one element. If not, call `done(null, false, { msg: '...' })` instead of throwing.

### Steps to reproduce

1. Use (or mock) a Google OAuth profile that has no `emails` or empty `emails` array.
2. Complete Google sign-in so the strategy callback runs.
3. Observe unhandled TypeError and OAuth failure instead of a user-friendly message.

### Snippet:

```js
// passport.js - Google strategy
const existingUser = await User.findOne({
  google: profile._json.emails[0].value  // crashes if emails undefined
}).exec();
// ...
const primaryEmail = profile._json.emails[0].value;
// ...
req.user.google = profile._json.emails[0].value;
```

KanhaiyaPandey:
## Summary

Fixes an issue in the Google OAuth strategy where the code directly accessed
`profile._json.emails[0].value` without validating that `emails` exists and
contains at least one entry.

If Google does not return an email array, this results in:
TypeError: Cannot read property '0' of undefined

This PR adds proper validation and ensures the OAuth flow fails gracefully
instead of crashing.

## Changes

- Added safe check for `profile.emails` (or `_json.emails`) before accessing index 0
- Extracted email into a `primaryEmail` variable
- If no email is found, return:
  `done(null, false, { msg: 'Google account does not provide an email address.' })`
- Replaced all direct usages of `profile._json.emails[0].value`

## Why This Is Needed

Google does not guarantee that the `emails` array will be present
(depending on scopes, account type, or privacy settings).
Without validation, the OAuth flow crashes instead of failing gracefully.

## Testing

- Mocked a Google profile without `emails`
- Confirmed OAuth flow no longer throws a TypeError
- Verified user-friendly failure message is returned instead

Closes #3907

### DIFF
--- a/server/config/passport.js
+++ b/server/config/passport.js
@@ -240,8 +240,21 @@ passport.use(
     },
     async (req, accessToken, refreshToken, profile, done) => {
       try {
+        const emailEntries = profile?.emails || profile?._json?.emails || [];
+        const primaryEmail = (
+          emailEntries.find(
+            (email) => typeof email?.value === 'string' && email.value.trim()
+          ) || {}
+        ).value;
+
+        if (!primaryEmail) {
+          return done(null, false, {
+            msg: 'Google account does not provide an email address.'
+          });
+        }
+
         const existingUser = await User.findOne({
-          google: profile._json.emails[0].value
+          google: primaryEmail
         }).exec();
 
         if (existingUser) {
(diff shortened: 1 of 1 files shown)


yugalkaushik (approved):
LGTM This is a strong fix.  Just few suggestions, instead of using (find(...) || {}).value, we could extract the found entry into a variable and use optional chaining and currently the code checks email.value.trim() but the trimmed value is not stored so a cleaner approach might be `const primaryEmail = primaryEmailEntry?.value.trim();`

KanhaiyaPandey:
Thanks for the suggestion! I’ve updated the implementation to extract the primary email entry into a variable and used optional chaining with a properly trimmed value for better readability and safety.

yugalkaushik (approved):
Perfect

raclim:
Closing this PR as well since at the time that this was opened, the original issue had an assignee who submitted a resolution for it.

PR #3974, opened 2026-03-06
Add tests for DropdownMenu component

siddocode:
This PR adds unit tests for the DropdownMenu component.

The tests verify that the dropdown button renders correctly, the menu opens when clicked, menu items are displayed, the menu closes after selecting an item, and alignment works as expected.

This improves test coverage for the Dropdown components.

### DIFF
--- a/client/components/Dropdown/DropdownMenu.test.tsx
+++ b/client/components/Dropdown/DropdownMenu.test.tsx
@@ -16,7 +16,7 @@ describe('DropdownMenu', () => {
     );
   };
 
-  it('should render the dropdown button', () => {   
+  it('should render the dropdown button', () => {
     renderDropdown();
 
     const button = screen.getByRole('button', { name: 'More options' });
@@ -52,6 +52,4 @@ describe('DropdownMenu', () => {
       expect(screen.queryByRole('menu')).not.toBeInTheDocument();
     });
   });
-
-  
 });
--- a/client/components/Dropdown/DropdownMenu.test.tsx
+++ b/client/components/Dropdown/DropdownMenu.test.tsx
@@ -16,29 +16,23 @@ describe('DropdownMenu', () => {
     );
   };
 
-  it('should render the dropdown button', () => {
+  it('should render the dropdown button', () => {   
     renderDropdown();
 
     const button = screen.getByRole('button', { name: 'More options' });
     expect(button).toBeInTheDocument();
+
+    expect(screen.queryByRole('menu')).not.toBeInTheDocument();
   });
 
-  it('should open the dropdown menu when button is clicked', () => {
+  it('should open dropdown and render menu items when button is clicked', () => {
     renderDropdown();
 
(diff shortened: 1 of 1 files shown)


client/components/Dropdown/DropdownMenu.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  it('should render menu items when opened', () => {
+    renderDropdown();
+
+    const button = screen.getByRole('button', { name: 'More options' });
+    fireEvent.click(button);
+
+    expect(screen.getByText('Item One')).toBeInTheDocument();
+    expect(screen.getByText('Item Two')).toBeInTheDocument();
+    expect(screen.getByText('Item Three')).toBeInTheDocument();
```
clairep94: Can you put these together? Not sure you need a separate test in ln 26

client/components/Dropdown/DropdownMenu.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+      </DropdownMenu>
+    );
+  };
+
+  it('should render the dropdown button', () => {
+    renderDropdown();
+
+    const button = screen.getByRole('button', { name: 'More options' });
+    expect(button).toBeInTheDocument();
+  });
```
clairep94, line 26: Can you add a check that 'menu' is not in the DOM -- it renders in an initial closed state

client/components/Dropdown/DropdownMenu.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+      </DropdownMenu>
+    );
+
+    const button = screen.getByRole('button', { name: 'Aligned dropdown' });
+
+    fireEvent.click(button);
+
+    const menu = screen.getByRole('menu');
+    expect(menu).toBeInTheDocument();
+  });
```
clairep94: this test doesn't check that it supports left alignment, we should be looking for a computed style instead

clairep94: Feel free to also just omit this test

clairep94 (commented):
Great start!

clairep94 (changes requested):

client/components/Dropdown/DropdownMenu.test.tsx
siddocode: I've combined the tests as suggested.

client/components/Dropdown/DropdownMenu.test.tsx
siddocode: I've added the check to ensure the menu is not present initially.

client/components/Dropdown/DropdownMenu.test.tsx
siddocode: Removed the alignment test as suggested.

clairep94 (approved):

raclim (approved):
Thanks @sid2487 and @clairep94!

PR #3981, opened 2026-03-08
Add setup instructions for minio

### LINKED ISSUE #3975: Add minio setup instructions
### Increasing Access

New contributors will be able to test saving sketches, collections and assets without having the need to setup an AWS S3 which is harder for beginners.

Minio is a local equivalent for S3 to simulate/test AWS S3 behavior this would enable new comers to contribute to the editor quicker without the need to setup AWS S3

kitlord:
### Issue:
Fixes #3975 
Adds documentation for minio setup

### Changes:
Adds a doc for minio setup and adds minio folder in gitignore

### DIFF
--- a/contributor_docs/s3_configuration.md
+++ b/contributor_docs/s3_configuration.md
@@ -1,4 +1,103 @@
 # S3 Bucket Configuration
+
+This guide covers two options for S3 storage configuration:
+- **[Option 1: MinIO (Local Development)](#minio-local-development)** - Recommended for local development
+- **[Option 2: AWS S3 (Production)](#aws-s3-production)** - For production environments
+
+## MinIO (Local Development)
+
+MinIO is an S3-compatible object storage server that can run locally, making it ideal for development without needing AWS credentials or incurring cloud storage costs.
+
+### Installation
+
+Install `minio-client` from a package manager then, choose one of the following installation methods:
+
+#### Option A: AUR (Arch Linux)
+```bash
+yay -S minio
+```
+
+#### Option B: Download Binary (Linux)
+```bash
+wget https://dl.min.io/server/minio/release/linux-amd64/minio
+chmod +x minio
+sudo mv minio /usr/local/bin/
+```
+
+Visit [https://www.min.io/download](https://www.min.io/download) for other installation options.
+
+### Setup
+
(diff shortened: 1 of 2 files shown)


contributor_docs/s3_configuration.md
```diff
@@ -1,4 +1,103 @@
 # S3 Bucket Configuration
+
+This guide covers two options for S3 storage configuration:
+- **[Option 1: MinIO (Local Development)](#minio-local-development)** - Recommended for local development
+- **[Option 2: AWS S3 (Production)](#aws-s3-production)** - For production environments
+
+## MinIO (Local Development)
```
raclim, line 77: Thanks for raising this! 

I'm wondering could we maybe switch the order of the options so that the AWS S3 one is first and the MinIO is second? I think it would be great to encourage users to try to get as close to the setup we have as much as possible, while the MinIO option could be present for users who would like to explore a different option.

contributor_docs/s3_configuration.md
```diff
@@ -1,4 +1,103 @@
...
+
+MinIO is an S3-compatible object storage server that can run locally, making it ideal for development without needing AWS credentials or incurring cloud storage costs.
```
raclim: For the section regarding cloud storage costs, I believe that creating an AWS account and S3 bucket is free and does not require any additional costs, but I might need to double check!

contributor_docs/s3_configuration.md
raclim: I'm also a little hesitant to recommend MinIO as the primary method for local development, OS type, since at the moment we only have docs for Linux. Maybe we could rephrase this?

raclim:
Thanks for working on this! I added in a few notes on option order and recommendation, but I think overall this looks great!

contributor_docs/s3_configuration.md
kitlord: Yeah sure we can switch the order of aws s3 and minio since it would be more relevant, also minio actually does have docs and download for windows so we could either give the docs link to setup minio in windows mention or explicitly say the setup for windows

kitlord:
@raclim I have changed the order for aws and minio

raclim (approved):
Thanks this looks great!

PR #4054, opened 2026-03-29
fix: Multiple Sketch/Collection Selection and Visibility Control 

### LINKED ISSUE #3618: Multiple Sketch/Collection Selection and Visibility Control for Collections
### Increasing Access

Unsure

I really like the visibility control feature for sketches. However, it would be much more useful if multiple sketches in the “Sketches” view could be selected and have their visibility changed at once. In addition, this multi-selection UI could also serve to add sketches to a collection, download or delete them more efficiently.

If visibility control were extended to Collections, it would also be valuable to adjust the visibility of entire collections, or of all sketches within a collection, at once. Together, these improvements would significantly enhance efficiency in sketch management.

localhost-legend:
Summary

      - **client/components/SketchesView.js**: Added multi-selection UI for sketches and collections, and extended visibility control to collections

      What changed
      Implement multi-selection UI in the Sketches view and extend visibility control to Collections, allowing users to select multiple sketches and adjust their visibility or add them to a collection.

       Testing
      I tested this locally and the changes work as expected. Happy to make any adjustments if needed!

      Closes #3618

### DIFF
--- a/client/components/SketchesView.js
+++ b/client/components/SketchesView.js
@@ -1,0 +1,64 @@
+import React, { useState } from 'react';
+import Sketch from './Sketch';
+import Collection from './Collection';
+
+const SketchesView = () => {
+  const [selectedSketches, setSelectedSketches] = useState([]);
+  const [collections, setCollections] = useState([]);
+
+  const handleSketchSelect = (sketch) => {
+    const isSelected = selectedSketches.includes(sketch);
+    if (isSelected) {
+      setSelectedSketches(selectedSketches.filter((s) => s !== sketch));
+    } else {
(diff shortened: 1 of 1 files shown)


raclim:
Thanks for your interest in this issue! 

Unfortunately, this PR might be having a similar issue to some of the other ones that you submitted, where it's not making changes within the correct `jsx` file to update the component, and is instead creating a separate JavaScript one. For a feature like this, we probably also want to be updating the state/actions as well to ensure user interactivity is processed. 

I don't think the original issue was ready to be for development work yet, and is still in the design ideation process. Due to these reasons, I'm going to close this PR for now.

PR #4082, opened 2026-04-15
Create SECURITY.md for vulnerability reporting

ksen0:
Replicates security policy from https://github.com/processing/p5.js/edit/main/SECURITY.md


raclim (approved):
Thanks for adding this in! I don't think we have anything within the docs that explicitly state this information yet.

PR #4083, opened 2026-04-15
fix loop protection breaking shader strings and p5.strands functions

### LINKED ISSUE #4080: Loop protection breaks shaders
### p5.js version

All

Mac OS

### Web browser and version

All

### Actual Behavior

Shaders that have a for loop in its source code in a string, or p5.strands shader functions, will break when you have not added `// noprotect` as a comment.

### Expected Behavior

Ideally, neither of these should cause the sketch to fail, so loop protection would skip making changes in strings or in p5.strands functions.

### Steps to reproduce

If you have a shader string that includes a for loop, such as in the following sketch, loop protection still gets added to the shader code, which produces invalid GLSL:

```js
let blur
function setup() {
  createCanvas(400, 400, WEBGL)
  blur = buildFilterShader({
    'vec4 getColor': `(FilterInputs inputs, in sampler2D canvasContent) {
      vec4 total = vec4(0., 0., 0., 0.);
      for (int i = 0; i < 20; i++) {
        total += getTexture(
          canvasContent,
          inputs.texCoord + 15. * vec2(
            cos(float(i)/20. * ${TWO_PI}),
            sin(float(i)/20. * ${TWO_PI})
          ) / inputs.canvasSize
        );
      }
      vec4 avg = total / 20.;
      return avg;
    }`
  })
}

function draw() {
  background(220)
  noStroke()
  circle(0, 0, 100)
  filter(blur)
}
```
Live (remove the `nopretect` to test): https://editor.p5js.org/davepagurek/sketches/ozs78nOXV

More complicated still is a p5.strands shader, which is still javascript, but gets compiled into GLSL, so it needs to not have loop protection added in it:
```js
let blur
function setup() {
  createCanvas(400, 400, WEBGL)
  blur = buildFilterShader(doBlur)
}

function doBlur() {
  filterColor.begin()
  let total = [0, 0, 0, 0]
  for (let i = 0; i < 20; i++) {
    total += getTexture(
      filterColor.canvasContent,
      filterColor.texCoord + 15 * [
        cos(i/20 * TWO_PI),
        sin(i/20 * TWO_PI)
      ] / filterColor.canvasSize
    )
  }
  let avg = total / 20
  filterColor.set(avg)
  filterColor.end()
}

function draw() {
  background(220)
  noStroke()
  circle(0, 0, 100)
  filter(blur)
}
```
Live (remove `noprotect` to test): https://editor.p5js.org/davepagurek/sketches/ByAq44OVU

Ideally, loop protection should not break code. To do this, we'd likely have to shift away [from the current regex-based approach](https://github.com/catarak/loop-protect/blob/2691548c37c7883e6d72f01ce1724de37b6f9692/lib/loop-protect.js#L29-L32) used by [catarak/loop-protect](https://github.com/catarak/loop-protect).

The original [jsbin/loop-protect](https://github.com/jsbin/loop-protect) has since been updated to use Babel's AST, so it would not transform strings (it's interesting to see the change in readme between the fork, which branched off of the jsbin one a while ago, and the more recently updated jsbin one, on its stance on ASTs.)

However, just updating to jsbin/loop-protect would not fix p5.strands functions: for that, we'd need to additionally (1) find p5.strands function calls (in the form of `build*Shader(...)` functions and `base*Shader().modify(...)` functions), and (2) skip any transpilation of functions passed into them, either inline functions or references to functions declared elsewhere.

Nixxx19:
### issue:
fixes #4080

### demo:
Screenshot 2026-04-15 at 4 46 25 PM

### changes:
- replaced regex-based loop protection with acorn ast-based approach in a new `jspreprocess.js` module
- strings are never visited by the ast parser, so glsl for loops inside template literals and quoted strings are no longer modified
- added two-pass detection for p5.strands shader functions: first pass collects function names passed to `build*shader()` and `base*shader().modify()`, second pass skips loop protection inside those functions
- removed `loop-protect` and `decomment` dependencies from `embedframe.jsx`
- added 13 unit tests covering all cases

### i have verified that this pull request:

* [x] is descriptively named and links to an issue number, i.e. `fixes #4080`

### DIFF
--- a/client/modules/Preview/jsPreprocess.js
+++ b/client/modules/Preview/jsPreprocess.js
@@ -1,5 +1,6 @@
 import * as acorn from 'acorn';
 import * as walk from 'acorn-walk';
+import escodegen from 'escodegen';
 
 const LOOP_TIMEOUT_MS = 100;
 
@@ -7,13 +8,9 @@ function isShaderCall(node) {
   const { callee } = node;
   const isBuildShader =
     callee.type === 'Identifier' && /^build\w*Shader$/.test(callee.name);
-  const isBaseShaderModify =
-    callee.type === 'MemberExpression' &&
-    callee.property.name === 'modify' &&
-    callee.object.type === 'CallExpression' &&
-    callee.object.callee.type === 'Identifier' &&
-    /^base\w*Shader$/.test(callee.object.callee.name);
-  return isBuildShader || isBaseShaderModify;
+  const isModifyCall =
+    callee.type === 'MemberExpression' && callee.property.name === 'modify';
+  return isBuildShader || isModifyCall;
 }
 
 function collectShaderFunctionNames(ast) {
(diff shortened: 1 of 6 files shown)


client/modules/Preview/jsPreprocess.js
```diff
@@ -0,0 +1,132 @@
...
+function isShaderCall(node) {
+  const { callee } = node;
+  const isBuildShader =
+    callee.type === 'Identifier' && /^build\w*Shader$/.test(callee.name);
+  const isBaseShaderModify =
+    callee.type === 'MemberExpression' &&
+    callee.property.name === 'modify' &&
+    callee.object.type === 'CallExpression' &&
+    callee.object.callee.type === 'Identifier' &&
+    /^base\w*Shader$/.test(callee.object.callee.name);
```
davepagurek: Might be worth noting that this might miss some things, e.g. if a plugin author exposes a shader of their own that has its own name or is just a variable and not a function. If we just check that it's a `MemberExpression` calling a property called `modify` that would capture all that, but may have some false positives if users have a method called `modify` in non-p5.strands code. I think either could be a reasonable compromise, @raclim let me know if you have any thoughts on which side to err on!

client/modules/Preview/jsPreprocess.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+  });
+  return names;
+}
+
+function collectLoopsToProtect(ast, shaderNames) {
+  const loops = [];
+
+  function visitNode(node, ancestors) {
+    const isInsideShader = ancestors.some((ancestor, idx) => {
```
davepagurek, line 114: nice, looks good!

client/modules/Preview/jsPreprocess.js
```diff
@@ -0,0 +1,132 @@
...
+export function jsPreprocess(jsText) {
+  if (/\/\/\s*noprotect/.test(jsText)) {
+    return jsText;
+  }
+
+  let ast;
+  try {
+    ast = acorn.parse(jsText, {
+      ecmaVersion: 'latest',
+      sourceType: 'script',
```
davepagurek: Maybe worth considering that users *could* use `<script type="module">` scripts, e.g. https://editor.p5js.org/davepagurek/sketches/-foIv7eQa -- worth testing to see if parsing fails in those if you use `sourceType: 'script'` and you have an import statement, or if we'd need to pass in the source type based on how the script tag imports it.

client/modules/Preview/jsPreprocess.js
```diff
@@ -0,0 +1,132 @@
...
+
+  const shaderNames = collectShaderFunctionNames(ast);
+  const loops = collectLoopsToProtect(ast, shaderNames);
+
+  if (loops.length === 0) return jsText;
+
+  const insertions = [];
+
+  loops.forEach((loop) => {
+    const { line } = loop.loc.start;
```
davepagurek: Possibly worth considering whether it's preferable to still do line-based insertions like this, which may have issues with multiple loops on the same line like what's implied by the comment in the current implementation:
https://github.com/processing/p5.js-web-editor/blob/37bc396e56ab0ed6fda9934ec75f89e79bcec918/client/modules/Preview/EmbedFrame.jsx#L66-L69

...or if instead it would be better to make changes at the AST level using Acorn by modifying the AST nodes, and then use a tool like `escodegen` to re-emit the full source code from that. (This is what we do in p5.strands after transpiling shader functions: https://github.com/processing/p5.js/blob/9498619e7b31f1b3a4e0fda5fcdc19d38e7eff20/src/strands/strands_transpiler.js#L1772) The downside of this approach is that `escodegen` doesn't preserve everything about your source code like whitespace and semicolon usage, so it might look a little different if you put a `debugger` statement and look at the code actually being run by the browser. Also it would mean adding another library, not sure how much that adds to the bundle size. But it would mean more targeted insertions and less likelihood for it happening incorrectly.

davepagurek (commented):
Great work on this! Consider this review nonblocking, just sharing some thoughts for you and @raclim to consider.

client/modules/Preview/jsPreprocess.js
Nixxx19: thank you!

client/modules/Preview/jsPreprocess.js
Nixxx19: good point, thank you! updated it to match any `.modify()` call instead of just `base*Shader().modify()`, which should cover plugin-defined shaders too. there could be false positives if someone has an unrelated `.modify()` call, but that feels like the safer tradeoff. happy to hear your thoughts on this.

client/modules/Preview/jsPreprocess.js
Nixxx19: thank you for catching this! added a fallback that tries `sourceType: 'script'` first, and if that throws (e.g. for a module script with import statements), it retries with `sourceType: 'module'`. added a test for this case too.

client/modules/Preview/jsPreprocess.js
Nixxx19: thank you for the suggestion! switched to the escodegen approach. we now modify the ast nodes directly and regenerate the source using `escodegen.generate()`. added escodegen as a dependency. the tradeoff on whitespace/formatting is real but feels worth it for the correctness guarantees. please let me know if the bundle size is a concern and we can fall back to character-offset string insertions instead, which also handles multiple loops on the same line correctly since we apply them back-to-front.

client/modules/Preview/jsPreprocess.js
raclim: I think @davepagurek's suggestion sounds good to me! I think we can keep the check broad for now, and refine it later down the line if needed.

client/modules/Preview/jsPreprocess.js
```diff
@@ -0,0 +1,238 @@
...
+}
+
+function injectVarDeclsIntoAst(ast, loops) {
+  const loopToVarDecl = new Map();
+  loops.forEach((loop, idx) => {
+    loopToVarDecl.set(loop, injectProtection(loop, idx));
+  });
+
+  walk.simple(ast, {
+    BlockStatement(node) {
```
davepagurek: Do you think we can simplify this step and modify the nodes in place in the earlier pass where we collect the loops to modify? Then we wouldn't need to find them again with another pass here? Or correct me if there's something that that would miss!

client/modules/Preview/jsPreprocess.js
Nixxx19: thank you both for the input! will keep it broad for now and revisit if false positives come up.

client/modules/Preview/jsPreprocess.js
```diff
@@ -0,0 +1,238 @@
...
+      return false;
+    });
+
+    if (!isInsideShader) loops.push(node);
+  }
+
+  walk.ancestor(ast, {
+    ForStatement: visitNode,
+    WhileStatement: visitNode,
+    DoWhileStatement: visitNode
```
raclim: Is there anything else(i.e `for...of` loops) we might want to cover here? Also okay with leaving it as is!

client/modules/Preview/jsPreprocess.js
Nixxx19: i chose the two-pass approach initially to keep the collection and injection steps separate, but you are right that since we already have the ancestors available in the walk, we can do it all in one pass. simplified it to a single `walk.ancestor` call that both checks for shader functions and injects the protection in place. updated in the latest commit. thank you for the suggestion!

client/modules/Preview/jsPreprocess.js
Nixxx19: `for...of` and `for...in` loops are parsed as separate ast node types (`ForOfStatement` and `ForInStatement`), so they are not currently covered. happy to add support for both if you think that would be a good addition!

Nixxx19:
added `--forceExit` to the jest config to fix the ci hang — `escodegen` keeps an open handle after tests finish which prevented jest from exiting cleanly

davepagurek:
Seems suspicious that `escodegen`, which is I think a synchronous function call, would hang the tests. Is there the potential that the AST has a loop in the data after it gets processed? that could potentially cause it to infinite loop while generating code.

Nixxx19:
> Seems suspicious that `escodegen`, which is I think a synchronous function call, would hang the tests. Is there the potential that the AST has a loop in the data after it gets processed? that could potentially cause it to infinite loop while generating code.

yeah i was thinking along the same lines, looking into the codebase right now to trace through what might actually be happening. will report back once i have a clearer picture of the issue

Nixxx19:
@davepagurek, the infinite loop got introduced when we moved to the single-pass approach where modifications happened inline during the walk. walk.ancestor iterates each block's body with a plain for loop, so inserting a var declaration before the current loop made the walker's index bump forward and land on the same loop again, revisiting it endlessly.                                                                                                                                            

the walk now collects each loop along with its parent block from the ancestors, and all modifications are applied after the walk completes so we never mutate arrays the walker is still iterating. verified locally on single, nested, sibling, and in-function loop cases. pushing the fix now.

Nixxx19:
works now, finally phew!

package.json
```diff
@@ -12,7 +12,7 @@
     "build:client": "cross-env NODE_ENV=production webpack --config webpack/config.prod.js",
     "build:server": "cross-env NODE_ENV=production webpack --config webpack/config.server.js",
     "build:examples": "cross-env NODE_ENV=production webpack --config webpack/config.examples.js",
-    "test": "NODE_ENV=test jest",
+    "test": "NODE_ENV=test jest --forceExit",
```
raclim, line 15: Since it seems like the issue with running tests got resolved, could we remove the `--forceExit`?

Nixxx19: yes, removed it!

Nixxx19:
good to go from my side now! all feedback addressed and ci is green. thanks for the thorough review @davepagurek @raclim!

raclim (approved):
Thanks for the quick turnaround on this! I think this looks good to me!

Nixxx19:
> Thanks for the quick turnaround on this! I think this looks good to me!

thanks for the review, happy to get this merged!

Issue #4085, opened 2026-04-15
Incorrect Asset Size Limit Error


Qianqianye:
### p5.js version

2.0.3

### What is your operating system?

Mac OS

### Web browser and version

chrome://146.0.7680.180 

### Actual Behavior

I tried to upload a small font file(50KB), but received an error indicating that I had exceeded the total asset size limit of 250MB, which blocks the upload. My current total asset size is approximately 68 MB, which is below the stated 250 MB limit. This suggests the limit may be miscalculated or not reflecting actual usage.

Screenshot of the error:
<img width="399" height="202" alt="Image" src="https://github.com/user-attachments/assets/756b9b6b-a91e-41ca-b91d-b83caf7ff1db" />

Screenshot of the total asset size:
<img width="294" height="82" alt="Image" src="https://github.com/user-attachments/assets/8b2e563b-5c5f-471e-bf61-fc5315378bba" />

### Expected Behavior

The upload should succeed, as my total asset usage is below the 250 MB limit.

### Steps to reproduce

1. Navigate to asset upload.
2. Attempt to upload a small file (e.g., font file).
3. Observe error message preventing upload.

raclim:
Thanks for flagging this! 

@yugalkaushik would you be interested in investigating into this if you have the chance, since you have familiarity with the parts related to the asset limit? I can also take a look into this!

yugalkaushik:
@raclim I’ll take a look at it

yugalkaushik:
I tried reproducing the issue, but everything seems to be working as expected on my end. Could you share any additional details

<img width="1919" height="870" alt="Image" src="https://github.com/user-attachments/assets/9c2bd0e3-30c5-425b-b64c-f7f122d6e281" />

https://github.com/user-attachments/assets/cc9da564-fc95-48da-8f22-69e1224b4216

raclim:
Also following up with other suggestions that could be impacting this: 
- the asset size was being miscalculated beforehand
- or the previously raised issue where some assets are not being fully deleted from user accounts despite deletion from the 'My Assets' tab. (Currently, there is progress towards asset cleanup but the bug still exists in the codebase)

Nixxx19:
i think i figured out why this happens. the size check on upload is counting every file in s3 for the user, but "my assets" only shows the ones that are actually referenced in a project. so if there are leftover files in s3 from older projects or deletions, they still count against the limit even though you cannot see them.

[...]

PR #4107, opened 2026-05-05
[Draft] Orphaned Assets Fix

raclim:
### Background

This PR addresses issues with orphaned assets accumulating in the AWS s3 Bucket. Currently, one confirmed cause can be reproduced by:  
1. Creating a New Project. 
2. Uploading an Asset. 
3. Closing the Project Window/Tab without Saving the Project. 
4. Comparing the listed assets within the database vs. the s3 bucket. The asset uploaded to the unsaved project will be visible in AWS, but the project will not exist in the database. 

Another potential cause is when project cloning is disrupted, since it appears that assets are uploaded to s3 before the project is successfully cloned. 

### Changes:

This PR proposes creating a `pending` folder within the S3 bucket that will hold uploaded user assets until a user saves their project. On project save, the asset should be removed from the `pending` folder and be moved to the general bucket. This PR is incomplete, and currently conflicts with other scenarios, such as when a user uploads assets to a saved project. 

This solution altogether is also just a suggestion, so I'm definitely open to other ideas on this as well!

- `aws.controller.js` : Reference `req.user.id` instead of `req.body.userId`. Adds in the `pending` folder when a user uploads an asset.
-  `server/utils/pendingAssets.js`: Adds a util that handles committing any pending assets to the general bucket on project save, which should copy the asset and delete it from the `pending` folder. 
- `server/controllers/project.controller.js`: Adds the `commitPendingAssets` util to the project controller and removes the portion checking for stale files. I felt that this wasn't successfully capturing 
- `client/modules/IDE/actions/project.js`: When a user attempts to clone a project, it will now delete any added s3 files if the project cloning fails. 

### DIFF
--- a/client/modules/IDE/actions/project.js
+++ b/client/modules/IDE/actions/project.js
@@ -306,6 +306,7 @@ export function cloneProject(project) {
     generateNewIdsForChildren(rootFile, newFiles);
 
     // duplicate all files hosted on S3
+    const copiedAssetKeys = [];
     each(
       newFiles,
       (file, callback) => {
@@ -319,10 +320,16 @@ export function cloneProject(project) {
           const formParams = {
             url: file.url
           };
-          apiClient.post('/S3/copy', formParams).then((response) => {
-            file.url = response.data.url;
-            callback(null);
-          });
+          apiClient
+            .post('/S3/copy', formParams)
+            .then((response) => {
+              file.url = response.data.url;
+
+              const objectKeyFromUrl = file.url.split('/').pop();
+              copiedAssetKeys.push(objectKeyFromUrl);
+              callback(null);
+            })
+            .catch(callback);
         } else {
           callback(null);
         }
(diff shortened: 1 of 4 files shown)


skyyash:
hi @raclim!

i was reviewing the PRs around this and realized: why are we allowing asset uploads before a project is saved in the first place?

it seems like both this and [#4085](https://github.com/processing/p5.js-web-editor/issues/4085) originate from that behavior. if a project doesn't exist yet, could we simply prompt the user to save/create the project before allowing uploads?

image

i might be missing an important use case here, but at first glance that feels simpler than introducing a pending-assets workflow in S3 and handling all the edge cases around moving assets between pending and committed states.

is there a reason we need to support uploads to unsaved projects?

ksen0:
Great question, thanks for flagging @skyash-dev . We ended up having a longer discussion about the intent for the current flow (@catarak @raclim and me). The original intent was in  line with the core goal of the editor: to make the it possible to make quick sketches _without saving_; including sketches with assets. (We noted that UX around saving has also changed since original implementation.) So I think keeping the approach of temporary storage enables this (linking open PRs:  https://github.com/processing/p5.js-web-editor/pull/4162, https://github.com/processing/p5.js-web-editor/pull/4161, https://github.com/processing/p5.js-web-editor/pull/4119).

PR #4108, opened 2026-05-06
add UI to toggle loop protection via <!-- noprotect -->

### LINKED ISSUE #4081: Add UI to toggle on/off loop protection comment
### Increasing Access

Loop protection currently breaks shaders (https://github.com/processing/p5.js-web-editor/issues/4080) but there is not a clear way to fix the issue when it happens. Adding some UI to toggle on/off loop protection makes it more discoverable.

Ideally the actual state being toggled just lives in the code. Similar to how the library version picker reads the HTML source code and changing the version changes the source code, a toggle for loop protection could add or remove the `// noprotect` comment.

As extra inspiration, here's what OP does:

Image

skyash-dev:
### Issue:
Fixes #4081

Adds a UI control for enabling/disabling loop protection from Preferences.
This implementation treats `<!-- noprotect -->` in `index.html` as the source of truth:
- turning loop protection off inserts `<!-- noprotect -->` in `index.html`
- which turns off loop protection globally (i.e. does not insert loop protection code in all the sketches).
- turning it on removes the comment
- the toggle state is derived from the `index.html`

### Demo:
image

### Changes:
- added loop protection toggle UI in Preferences
- derive loop protection state from `index.html`
- update `index.html` noprotect comment when toggled
- keep toggle and `<!-- noprotect -->` comment in sync
- added unit test


client/modules/IDE/components/Preferences/index.jsx
```diff
@@ -53,6 +54,16 @@ export default function Preferences() {
   const { versionInfo, indexID } = useP5Version();
   const cmRef = useContext(CmControllerContext);
   const [showStars, setShowStars] = useState(null);
+  const files = useSelector((s) => s.files);
+  const sketchFile = files.find(
```
davepagurek: A question I have (maybe @Nixxx19 knows?) is whether there's anything we need to look out for with multi-file sketches. Does each file have to have this comment?

client/modules/IDE/utils/loopProtection.js
```diff
@@ -0,0 +1,18 @@
+const NO_PROTECT_REGEX = /^\s*\/\/\s*noprotect\b.*\n?/m;
```
davepagurek: Nice 👍

client/styles/components/_overlay.scss
```diff
@@ -8,7 +8,6 @@
   bottom: 0;
   z-index: 9999;
   background-color: rgba(0, 0, 0, 0.5);
-  overflow-y: hidden;
```
davepagurek, line 11: What was this affecting?

client/modules/IDE/components/Preferences/index.jsx
Nixxx19: yeah, the // noprotect check runs per file in jsPreprocess, so each file needs its own comment to skip protection. since this pr only edits sketch.js, loops in other files would still get protected when the toggle is off. probably worth iterating over all .js files when toggling

davepagurek: How do you think we should translate the multi-file state into a single toggle? Maybe we count it as off until every js file has it?

Nixxx19: yeah id go with that, off only when every file has the comment, mixed reads on, toggling bulk-applies across every js file. also considered a three-state mixed/on/off ui to preserve per-file intent but mixed checkboxes are rare in p5 editor and probably confusing for beginners

client/styles/components/_overlay.scss
skyyash: `overflow-y: hidden` on `.overlay` was interfering with flexbox centering. during re-renders (toggle loop protection), if the modal content temporarily exceeded its height limits, the browser could miscalculate the centered position and shift the modal upward.
since .overlay already fills the viewport (position: fixed), the overflow rule was unnecessary. removing it fixed the centering issue without any side effects.

client/modules/IDE/components/Preferences/index.jsx
skyyash: @Nixxx19 mixed state might end up making the toggle confusing and of less use. 

now that loop protection has a dedicated UI, maybe we can instead treat the main `index` file as the source of truth for `noprotect` and apply the toggle behavior based on that?

client/styles/components/_overlay.scss
davepagurek: makes sense, thanks for explaining!

client/modules/IDE/components/Preferences/index.jsx
Nixxx19: yeah main file as source of truth works, simple + clean. only flag: helper.js could still have // noprotect while ui reads off the main file, so toggle wont reflect it

skyyash:
toggle padding/margins feel off, making the outlines appear uncentered. can i create a commit in this pr itself?

actual:
image

improved:
image

davepagurek:
I think we're good to go ahead with the comment living in index.html -- possibly a bit unexpected for users of other tools, but then it's in one spot, and people mostly will use the UI anyway.

For padding/margins, maybe make a separate issue/PR? It's always a little weird doing design updates since often these are fairly subjective and could use a bit of feedback, so we don't need to block this PR on those updates.

raclim (approved):
This overall seems okay to me, will merge this in! 

Thanks so much @skyash-dev and @davepagurek for your efforts on this!

PR #4120, opened 2026-05-19
fix firefox scrollbar from fractional split pane size

### LINKED ISSUE #4046: Full-frame sketches have scroll bars by default
### p5.js version

1.11.1

Mac OS

### Web browser and version

Firefox 148

### Actual Behavior

`createCanvas(windowWidth, windowHeight)` has some subtle scrolling:

Image

### Expected Behavior

Ideally there should be no scroll bars. I think this may be a recent regression, since it used to not be this way?

### Steps:
Make sure you can actually see scroll bars. On macs they only show up while actively scrolling while on a device with a trackpad. For me anyway, they show up all the time when I have a mouse. But you can turn them on all the time for testing: https://unsung.aresluna.org/testing-tip-always-show-scrollbars/

### Snippet:
```js
function setup() {
  createCanvas(windowWidth, windowHeight);
}

function draw() {
  background(220);
}
```

Live: https://editor.p5js.org/davepagurek/sketches/Rfq0Q21AD

Nixxx19:
fixes #4046

the iframe was inheriting fractional pixel dimensions from its flex parent (eg 787.5px), so the sketch inside (which sees an integer 788) overflowed by 0.5px and triggered a scrollbar in firefox/brave.

floored the iframe width and height via a ResizeObserver on its parent and added `display: block`. tested locally in brave, scrollbar is gone, chrome unchanged.

### DIFF
--- a/client/jest.setup.js
+++ b/client/jest.setup.js
@@ -5,3 +5,11 @@ import 'regenerator-runtime/runtime';
 // See: https://github.com/testing-library/jest-dom
 // eslint-disable-next-line import/no-extraneous-dependencies
 import '@testing-library/jest-dom';
+
+global.ResizeObserver = class {
+  observe() {}
+
+  unobserve() {}
+
+  disconnect() {}
+};
--- a/client/modules/IDE/components/PreviewFrame.jsx
+++ b/client/modules/IDE/components/PreviewFrame.jsx
@@ -1,18 +1,18 @@
-import React, { useRef, useEffect } from 'react';
+import React, { useRef, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { getConfig } from '../../../utils/getConfig';
 import { registerFrame } from '../../../utils/dispatcher';
 
 const Frame = styled.iframe`
-  min-height: 100%;
-  min-width: 100%;
+  display: block;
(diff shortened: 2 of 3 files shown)


yugalkaushik (changes requested):
@Nixxx19 I tested your PR locally and the scrollbars still exist. I checked on Brave and the issue is still present. On Chrome, there were no scrollbars before and there are still none.
Also, if you make any further changes, please share a preview as well.

Brave Preview:
image

Nixxx19:
hey @yugalkaushik thanks for testing and sorry for not attaching a screenshot earlier. just pushed a fix, tested locally in brave and the scrollbar is gone, screenshot attached. mind giving it another try?


Screenshot 2026-05-20 at 4 15 24 PM

yugalkaushik:
@Nixxx19 I tested it locally and everything looks clean now. When you get some time, could you please test it on Firefox as well so we can confirm the issue is fixed once and for all? I’ll approve the PR after you verify it there.

yugalkaushik:
I tested it on firefox and it works. @Nixxx19 Thank you for the amazing work.
@ksen0 This PR can be merged.
image

ksen0:
Thanks both!

ksen0 (approved):

PR #4156, opened 2026-06-06
[CodeMirror v6] fix stale react state in cm6 update listener

### LINKED ISSUE #4155: [CodeMirror v6] Update listener captures stale React state
### Actual Behavior

values captured by the CM6 update listener can become stale after the editor state has been created.

examples observed:
- autorefresh remains false after being enabled
- project id remains stale
- files array can become stale

this causes features that rely on these values to behave incorrectly.
for example, auto-refresh may not run when enabled. (can be also tested in cm6 release env: https://app-tedc8aa.release.editor.p5js.org/)

### Expected Behavior

the CM6 update listener should always use the latest react state values.
changes to `autorefresh`, project, and files should be reflected in
`onViewUpdate/onChange` callbacks.

### Steps to reproduce

Steps:
1. open the cm6 editor with autorefresh disabled (https://app-tedc8aa.release.editor.p5js.org/).
2. enable autorefresh.
3. edit the sketch.

Expected:
the sketch automatically reruns.

Actual:
the listener still sees autorefresh as false and the sketch does not rerun.

edit: fixed this locally!

skyash-dev:
### Issue:
Fixes #4155
the listener is installed when file states are created and can outlive subsequent React renders. as a result, callbacks could read stale values for autorefresh, project, and files.

### Changes

- added a ref for dynamic values used by onChange
- read autorefresh, project, and files from the ref instead of the captured closure

### Testing

- Verified autorefresh updates correctly after being enabled
- Verified files reflects newly created files
- Verified project id stays current

### DIFF
--- a/client/modules/IDE/components/Editor/codemirror.js
+++ b/client/modules/IDE/components/Editor/codemirror.js
@@ -49,18 +49,30 @@ export default function useCodeMirror({
   const fileId = useRef();
   fileId.current = file.id;
 
+  // CM6 update listeners are attached when file states are created and
+  // can outlive React renders. Keep dynamic values in a ref so onChange
+  // always sees the latest state.
+  const latestRef = useRef({});
+  latestRef.current = {
+    autorefresh,
+    project,
+    files
+  };
+
   // When the file changes, update the file content and save status.
   function onChange() {
+    const latest = latestRef.current;
+
     setUnsavedChanges(true);
     updateFileContent(fileId.current, cmView.current.state.doc.toString());
 
     // Save a local backup to localStorage for crash recovery (#3891).
     // This ensures work is recoverable even if the tab crashes
     // (e.g. from an infinite loop) before the server autosave fires.
-    const projectId = project?.id || 'unsaved';
-    saveLocalBackup(projectId, files);
+    const projectId = latest.project?.id || 'unsaved';
+    saveLocalBackup(projectId, latest.files);
 
-    if (autorefresh) {
+    if (latest.autorefresh) {
       clearConsole();
       startSketch();
     }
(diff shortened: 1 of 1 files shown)


client/modules/IDE/components/Editor/codemirror.js
```diff
@@ -49,18 +49,30 @@ export default function useCodeMirror({
   const fileId = useRef();
   fileId.current = file.id;
 
+  // CM6 update listeners are attached when file states are created and
+  // can outlive React renders. Keep dynamic values in a ref so onChange
+  // always sees the latest state.
+  const latestRef = useRef({});
```
khanniie: let's call this onChangeStateRef to be descriptive

Also, could you move the fileId ref into this as well? It's used for the same purpose and it's probably better to have them all in one place.

client/modules/IDE/components/Editor/codemirror.js
```diff
@@ -49,18 +49,30 @@ export default function useCodeMirror({
   const fileId = useRef();
   fileId.current = file.id;
 
+  // CM6 update listeners are attached when file states are created and
```
khanniie: I looked into the issue and your solution is perfect, but I think the real issue is that the debouncer for onChange is holding onto a stale closure (as you said) of the function when it first fires and starts waiting for the 1s. I do think we need to keep the debouncer to stop onChange from firing too often, so this is the best way to fix the issue, but maybe we can update the comment?

khanniie (approved):
A few small comments but thank you so much for identifying the issue and submitting a fix so fast! Really glad you caught this : D

client/modules/IDE/components/Editor/codemirror.js
skyyash: sounds good to me.

client/modules/IDE/components/Editor/codemirror.js
skyyash: hmm. updated!

PR #4161, opened 2026-06-06
fix: remove sketch file reference when deleting assets

### LINKED ISSUE #4160: Deleting an asset leaves a broken file reference in sketch files
### p5.js version

_No response_

Windows

### Web browser and version

_No response_

### Actual Behavior

When I delete an image or other asset from the Assets tab, it disappears from my assets list, but the filename still shows up in the sketch files sidebar. The file itself is gone, so the preview is broken.
Deleting the same file from the sketch files sidebar works correctly. The problem only happens when deleting from Assets.

### Expected Behavior

Deleting an asset from the Assets tab should remove it from storage and from every sketch that references it, same as deleting it from sketch files.

### Steps to reproduce

1. Open a sketch and upload an image.
2. Confirm the file appears in both the sketch files sidebar and the Assets tab.
3. Go to the Assets tab and delete that asset.
4. Open the sketch again and check the sketch files sidebar.

yugalkaushik:
### Issue:
Fixes #4160

Asset deletion now uses the same file delete endpoint as the sketch files sidebar, so both the storage object and the project file reference are removed together.

### Before
https://github.com/user-attachments/assets/07b403cf-552d-4114-b3d4-400ae9110953
### After
https://github.com/user-attachments/assets/ba956732-35f5-44ee-89ee-78de68ad73c0
### Changes:
1. Include fileId and parentId in the asset list API response when matching assets to sketch files
2. Update asset deletion to call /projects/:sketchId/files/:fileId instead of /S3/delete
3. Sync Redux state when the currently open sketch matches the deleted asset

### DIFF
--- a/client/modules/IDE/actions/assets.js
+++ b/client/modules/IDE/actions/assets.js
@@ -2,6 +2,7 @@ import { apiClient } from '../../../utils/apiClient';
 import * as ActionTypes from '../../../constants';
 import { startLoader, stopLoader } from '../reducers/loading';
 import { assetsActions } from '../reducers/assets';
+import { setProjectSavedTime } from './project';
 
 const { setAssets, deleteAsset } = assetsActions;
 
@@ -27,14 +28,24 @@ export function getAssets() {
   };
 }
 
-export function deleteAssetRequest(assetKey) {
-  return async (dispatch) => {
+export function deleteAssetRequest(asset) {
+  return async (dispatch, getState) => {
     try {
-      const path = assetKey.split('/').pop();
-      await apiClient.delete(
-        `/S3/delete?objectKey=${encodeURIComponent(path)}`
+      const response = await apiClient.delete(
+        `/projects/${asset.sketchId}/files/${asset.fileId}`,
+        { params: { parentId: asset.parentId } }
       );
-      dispatch(deleteAsset(assetKey));
+      dispatch(deleteAsset(asset.key));
+
+      const { project } = getState();
+      if (project.id === asset.sketchId) {
+        dispatch(setProjectSavedTime(response.data.project.updatedAt));
+        dispatch({
+          type: ActionTypes.DELETE_FILE,
+          id: asset.fileId,
+          parentId: asset.parentId
+        });
+      }
     } catch (error) {
       dispatch({
         type: ActionTypes.ERROR
--- a/client/modules/IDE/actions/assets.js
+++ b/client/modules/IDE/actions/assets.js
@@ -2,7 +2,6 @@ import { apiClient } from '../../../utils/apiClient';
 import * as ActionTypes from '../../../constants';
 import { startLoader, stopLoader } from '../reducers/loading';
 import { assetsActions } from '../reducers/assets';
-import { setProjectSavedTime } from './project';
 
 const { setAssets, deleteAsset } = assetsActions;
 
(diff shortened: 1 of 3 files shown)


client/modules/IDE/actions/assets.js
```diff
@@ -27,14 +28,24 @@ export function getAssets() {
...
-  return async (dispatch) => {
+export function deleteAssetRequest(asset) {
+  return async (dispatch, getState) => {
     try {
-      const path = assetKey.split('/').pop();
-      await apiClient.delete(
-        `/S3/delete?objectKey=${encodeURIComponent(path)}`
+      const response = await apiClient.delete(
+        `/projects/${asset.sketchId}/files/${asset.fileId}`,
+        { params: { parentId: asset.parentId } }
```
skyyash: nice! we don't need to store the response.

client/modules/IDE/actions/assets.js
```diff
@@ -27,14 +28,24 @@ export function getAssets() {
...
+
+      const { project } = getState();
+      if (project.id === asset.sketchId) {
+        dispatch(setProjectSavedTime(response.data.project.updatedAt));
+        dispatch({
+          type: ActionTypes.DELETE_FILE,
+          id: asset.fileId,
+          parentId: asset.parentId
+        });
+      }
```
skyyash, line 51: we can remove this. unlike `deleteFile`, there's no need to reactively update local state here since the user is on a different screen, and returning to the sketch will load the latest data anyway.

edit: also `getState()` returns the currently open project state, which may not even be the project associated with the asset being deleted.

client/modules/IDE/actions/assets.js
```diff
@@ -27,14 +28,24 @@ export function getAssets() {
...
 }
 
-export function deleteAssetRequest(assetKey) {
-  return async (dispatch) => {
+export function deleteAssetRequest(asset) {
+  return async (dispatch, getState) => {
     try {
-      const path = assetKey.split('/').pop();
-      await apiClient.delete(
-        `/S3/delete?objectKey=${encodeURIComponent(path)}`
```
skyyash, line 35: since #4162 displays all uploaded assets in the assets tab, including assets that are not referenced by any sketch, we need to check whether an asset is referenced before deleting it.

if the asset is not referenced by any sketch, we should use this api route directly. for referenced assets, we should continue using the existing deletion flow that also removes the asset reference from the sketch.

client/modules/IDE/components/AssetListRow.jsx
```diff
@@ -33,7 +33,10 @@ AssetMenu.propTypes = {
   item: PropTypes.shape({
     key: PropTypes.string.isRequired,
     url: PropTypes.string.isRequired,
-    name: PropTypes.string.isRequired
+    name: PropTypes.string.isRequired,
+    sketchId: PropTypes.string.isRequired,
+    fileId: PropTypes.string.isRequired,
+    parentId: PropTypes.string.isRequired
```
skyyash: also need to make the schema consistent for unreferenced (pending) assets so the frontend can handle both referenced and unreferenced assets

yugalkaushik: fixed

client/modules/IDE/actions/assets.js
yugalkaushik: fixed

client/modules/IDE/actions/assets.js
yugalkaushik: Done.

client/modules/IDE/actions/assets.js
yugalkaushik: Regarding getState, it's still present solely for the DELETE_FILE dispatch, which is guarded by if (project.id === asset.sketchId). This ensures that when a user deletes a referenced asset from the Assets tab while on the same sketch, the file tree sidebar updates immediately without requiring a page refresh. When the user is viewing a different sketch, the guard prevents any action. So getState is safe and necessary for this reactive UI update.

PR #4172, opened 2026-06-12
GSOC-E2E-1: Add Playwright CI workflow and sketch execution test

Geethegreat:
### Issue:
Fixes # 

### Demo:

https://github.com/user-attachments/assets/36dc541b-8993-4478-a725-0514bc004fff

### Changes:
This PR introduces an initial Playwright E2E testing setup for the p5.js Web Editor.

- Added Playwright configuration (playwright.config.ts)
- Added GitHub Actions workflow for Playwright E2E execution
- Added initial Playwright E2E tests covering:
- Sketch execution and console output validation
    - Populate CodeMirror via keyboard input
    - Run sketch through the Play button
    - Verify sketch startup and console output

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,4 +1,5 @@
 name: E2E Tests
+
 on:
   pull_request:
     branches:
@@ -8,20 +9,104 @@ jobs:
   e2e:
     name: End-to-end tests
     runs-on: ubuntu-latest
+
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
+
       - name: Use Node.js
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v4
         with:
           node-version: '18.20.x'
-      - run: npm install
+          cache: 'npm'
+
+      - name: Start MongoDB
+        uses: supercharge/mongodb-github-action@1.10.0
+        with:
+          mongodb-version: '6.0'
+
+      - name: Create .env file
+        run: |
+          cat > .env << 'EOF'
+          API_URL=/editor
+          CORS_ALLOW_LOCALHOST=true
+          TRANSLATIONS_ENABLED=true
+          UI_ACCESS_TOKEN_ENABLED=false
+          UPLOAD_LIMIT=250000000
+
+          PORT=8000
+          PREVIEW_PORT=8002
+          EDITOR_URL=http://localhost:8000
+          PREVIEW_URL=http://localhost:8002
+
+          MONGO_URL=mongodb://localhost:27017/p5js-web-editor
+
+          SESSION_SECRET=ci-session-secret
+          EMAIL_VERIFY_SECRET_TOKEN=ci-verify-token
+
+          # These accounts are used by the app to serve examples.
+          # Real values not needed for E2E tests — dummy strings prevent crashes.
+          EMAIL_SENDER=ci@example.com
+          EXAMPLE_USER_EMAIL=examples@p5js.org
+          EXAMPLE_USER_PASSWORD=hellop5js
+          GG_EXAMPLES_USERNAME=generativedesign
+          GG_EXAMPLES_EMAIL=benedikt.gross@generative-gestaltung.de
+          GG_EXAMPLES_PASS=generativedesign
+          ML5_LIBRARY_USERNAME=ml5
+          ML5_LIBRARY_EMAIL=examples@ml5js.org
+          ML5_LIBRARY_PASS=helloml5
+
+          # Mailgun — non-empty string required to pass the startup check in
+          # server/utils/mail.ts. No emails are sent during E2E tests.
+          MAILGUN_KEY=dummy-mailgun-key
+          MAILGUN_DOMAIN=dummy.mailgun.org
+
+          # AWS S3 — non-empty strings required. No file uploads in E2E tests.
+          AWS_ACCESS_KEY=dummy-aws-access-key
+          AWS_SECRET_KEY=dummy-aws-secret-key
+          AWS_REGION=us-east-1
+          S3_BUCKET=dummy-bucket
+          S3_BUCKET_URL_BASE=https://dummy-bucket.s3.amazonaws.com
+
+          # GitHub OAuth — not tested in E2E suite, dummy values prevent crash.
+          GITHUB_ID=dummy-github-id
+          GITHUB_SECRET=dummy-github-secret
+
+          # Google OAuth — not tested in E2E suite, dummy values prevent crash.
+          GOOGLE_ID=dummy-google-id
+          GOOGLE_SECRET=dummy-google-secret
+          EOF
+
+      - name: Install dependencies
+        run: npm install
+
       - name: Install Playwright browsers
         run: npx playwright install --with-deps chromium
+
       - name: Start the app
         run: npm run start &
-        env:
-          NODE_ENV: test
+
       - name: Wait for app to be ready
-        run: npx wait-on http://localhost:8000 --timeout 60000
+        run: |
+          for i in $(seq 1 120); do
+            if curl -sf http://localhost:8000 | grep -q "p5"; then
+              echo "App fully ready"
+              exit 0
+            fi
+
+            echo "Still bundling... $i/120"
+            sleep 5
+          done
+
+          echo "App did not become ready"
+          exit 1
+
       - name: Run E2E tests
-        run: npx playwright test
+        run: npx playwright test --config=playwright.config.ts
+
+      - name: Upload Playwright report on failure
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: playwright-report
+          path: playwright-report/
+          retention-days: 7
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   e2e:
     name: End-to-end tests
+    environment: e2e-tests
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -78,7 +78,7 @@ jobs:
           EOF
 
       - name: Install dependencies
-        run: npm install
+        run: npm ci
 
       - name: Install Playwright browsers
         run: npx playwright install --with-deps chromium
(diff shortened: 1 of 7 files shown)


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,112 @@
...
+
+on:
+  pull_request:
+    branches:
+      - develop
+
+jobs:
+  e2e:
+    name: End-to-end tests
+    runs-on: ubuntu-latest
```
clairep94: @raclim @khanniie we want to gate this on maintainer-approval to run, the same way that deploy-staging does

Could we add a new environment to the repo settings called `e2e-tests` so we can do the below?
Proposing that we add a new environment instead of using `staging` in case there are settings inside staging that we want to differ from

```suggestion
  e2e:
    name: End-to-end tests
    environment: e2e-tests
    runs-on: ubuntu-latest
```

I also can't currently see environments on this repo, would it be possible to update my role to be able to view?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,112 @@
...
+          S3_BUCKET_URL_BASE=https://dummy-bucket.s3.amazonaws.com
+
+          # GitHub OAuth — not tested in E2E suite, dummy values prevent crash.
+          GITHUB_ID=dummy-github-id
+          GITHUB_SECRET=dummy-github-secret
+
+          # Google OAuth — not tested in E2E suite, dummy values prevent crash.
+          GOOGLE_ID=dummy-google-id
+          GOOGLE_SECRET=dummy-google-secret
+          EOF
```
clairep94: Please make this into a file next to `.env.local` called `.env.e2e` instead and then do the following:

```suggestion
      - name: Create .env file
        run: cp .env.e2e .env
```

This makes the e2e .env file way easier to maintain

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,112 @@
...
+
+      - name: Install dependencies
+        run: npm install
```
clairep94:
```suggestion
        run: npm ci
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,112 @@
...
+      - name: Run E2E tests
+        run: npx playwright test --config=playwright.config.ts
+
+      - name: Upload Playwright report on failure
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: playwright-report
+          path: playwright-report/
+          retention-days: 7
```
clairep94: need to add a new-line at the end of this file

playwright.config.ts
```diff
@@ -0,0 +1,18 @@
+import { defineConfig, devices } from '@playwright/test';
+
+export default defineConfig({
+  testDir: './tests/playwright',
+  testMatch: '**/*.spec.ts',
+ timeout: 120_000,
```
clairep94:
```suggestion
  timeout: 120_000,
```
missing a white-space

playwright.config.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  use: {
+    baseURL: 'http://localhost:8000',
+    headless: true,
```
clairep94: remove this and see update for package.json commands

package.json
clairep94, line 20: Add the below new scripts:

```
  "e2e": "playwright test --headed", 
  "e2e:ci": "playwright test", 
```

Then update how the workflow calls the e2e test with the `e2e:ci` command instead
Then add docs for devs on how to run these tests

clairep94: Additionally can you update `.gitignore` to ignore the artifacts playwright generates?

Add the below:
```
# Playwright
test-results/
playwright-report/
.playwright/
```

tests/playwright/editor.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    if (await openConsoleButton.isVisible()) {
+      await openConsoleButton.click();
+    }
+
+    await expect
+      .poll(() => page.locator('.preview-console__messages').textContent(), {
+        timeout: 15000
+      })
+      .toContain('hi from sketch');
+  });
```
clairep94: Hi Geeth I think you have too many test blocks here.

This first initial test should just be something like "Can run sketch code from the editor"
I haven't run the below locally, so you may need to edit this, but it would also be helpful to add some comments on what the bot is doing

Remember that this is an e2e test, so you want to describe things that a human user would do, rather than describe the technical internals of the test code

```suggestion
test.beforeEach(async ({ page }) => {
  // go to homepage
  await page.goto('/');
  
  // close cookie dialog if it appears
  const cookieDialog = page.getByRole('dialog');
  if (await cookieDialog.isVisible({ timeout: 3_000 }).catch(() => false)) {
    await cookieDialog.getByText(/allow essential/i).click();
    await cookieDialog.waitFor({ state: 'hidden' });
  }
});

test.describe('p5.js Editor – Playwright E2E', () => {
  test('can execute code from the editor by clicking the Play button', async ({ page }) => {
    const editor = page.locator('.CodeMirror');
    await editor.click();

    // type in editor
    // add a console log that we should see in the terminal
    await page.keyboard.press('Control+A');
    await page.keyboard.type(
      [
        'function setup() {',
        '  createCanvas(400, 400);',
        '}',
        '',
        'function draw() {',
        '  background(220);',
        "  console.log('hi from sketch');",
        '  noLoop();',
        '}'
      ].join('\n'),
      { delay: 5 }
    );

    // click Play button
    await page.getByLabel('Play sketch').click();

    // check that console log from editor code is shown in editor console
    const openConsole = page.getByLabel('Open console');
    if (await openConsole.isVisible().catch(() => false)) {
      await openConsole.click();
    }

    await expect(page.locator('.preview-console__messages')).toContainText(
      'hi from sketch',
      { timeout: 15_000 }
    );
    
    // not sure if there's a way to find the iframe, check the src and verify that its' sourcecode comes from the editor? In order to verify that the iframe also displays the code from the editor.
  });
```

clairep94 (changes requested):
Hi Geeth great start!
Some small bits of clean up for your setup code + some suggestions for the first E2e test

clairep94:
Could we also pre-fix all the GSOC E2e PRs with "GSOC-E2E-<number>"?

tests/playwright/editor.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    await page.waitForSelector('.CodeMirror', { timeout: 30_000 });
+
+    // Dismiss cookie banner via JS — handles the case where the button
+    // is outside the viewport due to the Redux DevTools sidebar
+    await page.evaluate(() => {
+      const btn = Array.from(document.querySelectorAll('button')).find((b) =>
+        /allow essential|allow all/i.test(b.textContent ?? '')
+      ) as HTMLElement | undefined;
+      btn?.click();
+    });
```
clairep94, line 19: is page.evaluate necessary?

does the below not work?


```suggestion
  const cookieDialog = page.getByRole('dialog');
  if (await cookieDialog.isVisible({ timeout: 3_000 }).catch(() => false)) {
    await cookieDialog.getByText(/allow essential/i).click();
    await cookieDialog.waitFor({ state: 'hidden' });
  }
```

tests/playwright/editor.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    ].join('\n');
+
+    // Wait for CodeMirror to be ready
+    await page.waitForFunction(
+      () => {
+        const wrapper = document.querySelector('.CodeMirror') as any;
+        return (wrapper?.CodeMirror?.getValue?.() ?? '').length > 0;
+      },
+      { timeout: 30_000 }
+    );
```
clairep94: does the below not work?


```suggestion
    const editor = page.locator('.CodeMirror');
    await expect(editor).toBeVisible({ timeout: 30_000 });
```

tests/playwright/editor.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+        .getState()
+        .files.find((f: any) => f.isSelectedFile);
+      if (!selectedFile) throw new Error('No selected file');
+
+      store.dispatch({
+        type: 'UPDATE_FILE_CONTENT',
+        id: selectedFile.id,
+        content: code
+      });
+    }, newCode);
```
clairep94: ideally we would instead have the bot type into the code editor, then we click play. (vs. injecting into the redux store state)

does that not work?


```suggestion
    // simulate typing code into editor, then clicking play
    const editor = page.locator('.CodeMirror');
    await editor.click();
    
    await page.keyboard.press('Control+A');
    await page.keyboard.type(
      newCode,
      { delay: 5 }
    );
```

tests/playwright/editor.spec.ts
Geethegreat: Thanks for the suggestion! I tried this approach initially but the "Allow Essential" button is pushed outside the viewport by the Redux DevTools sidebar that's present in the dev build, it exists in the DOM but locator.click() throws "element is outside of the viewport". The JS evaluate approach bypasses the geometry check entirely. Happy to add a comment explaining this if that helps.

tests/playwright/editor.spec.ts
Geethegreat: Good point, toBeVisible would work to confirm the element is in the DOM, but the reason I'm waiting for getValue().length > 0 is to confirm CodeMirror has actually loaded the default sketch content into its buffer. The element can be visible before the CM instance is fully initialized with content, and the Redux dispatch in the next step depends on files.find(f => f.isSelectedFile) being populated, but the suggested approach works just fine too!

tests/playwright/editor.spec.ts
Geethegreat: I tried keyboard.type() with { delay: 5 } but CodeMirror's bracket auto-completion adds closing brackets/braces as you type, so the resulting code becomes malformed and the sketch throws a runtime error instead of running. For example typing function setup() { causes CM to auto-insert a }, resulting in function setup() {} with the rest of the body outside the braces.
I also tried keyboard.insertText() which pastes the whole string at once, but that triggered a NotFoundError in CodeMirror's internal DOM (removeChild on a non-child node) it seems CM's virtual rendering doesn't handle that input event correctly.
The Redux dispatch approach was the only reliable method I found. That said, if there's a way to disable CodeMirror's auto-close-brackets feature before typing in the test, that would solve it cleanly — something like cm.setOption('autoCloseBrackets', false) before calling keyboard.type(). Would that be an acceptable approach?

https://github.com/user-attachments/assets/ce65f30e-dcc5-49c1-ac21-8c636d7bb39e

Geethegreat:
Closing as the revised version(#4199 ) is merged

PR #4179, opened 2026-06-15
Enable Autocomplete Hinter by Default

### LINKED ISSUE #4099: Enable Autocomplete Hinter by default in preferences
### Increasing Access

The p5.js Web Editor is built for everyone — but autocomplete being off by default quietly works against that mission.
Beginners misspell function names like createCanvas or mousePressed, the sketch silently breaks, and they have no idea why. That moment of confusion is often where they give up. Non-English speakers carry an extra burden typing long camel-cased English names from memory. Educators lose classroom time correcting avoidable spelling errors. Developers arriving from VS Code expect autocomplete to be on and assume it simply doesn't exist.
The feature already works. The only barrier is a default value users never knew they could change.

Image

The p5.js Web Editor manages editor preferences through Redux. The initial state for all preferences is defined in:

```
client/modules/IDE/reducers/preferences.js
```

Currently, the autocomplete preference is set to `false` in that initial state, meaning every user starts with autocomplete disabled unless they manually navigate to Preferences and turn it on. Most users — especially beginners — never do this.

The fix is a single-line change:

```js
// current
autocomplete: false,

// proposed
autocomplete: true,
```

Because preferences are persisted via Redux dispatch, any user who has previously and explicitly turned autocomplete off will retain that setting — their stored preference takes precedence over the initial state. This change only affects users who have never modified the preference, which is the majority of users.

This is the same pattern used to enable autosave by default in a prior enhancement. No new feature is being built — only the default value of an existing, stable, already-shipped feature is being changed. It is a one-line diff with no risk to existing functionality and no effect on users who have already set their preference deliberately.

skyyash:
### Issue:
Fixes #4099 
Changes the default autocompleteHinter preference from false to true.
This only affects users who have never modified the setting. Existing users who have explicitly enabled or disabled autocomplete will retain their stored preference.

### Changes:
- Set autocompleteHinter default value to true in the preferences.
- Updated preference tests to reflect the new default state.

### DIFF
--- a/client/modules/IDE/components/Editor/utils/fileState.test.js
+++ b/client/modules/IDE/components/Editor/utils/fileState.test.js
@@ -7,7 +7,7 @@ import { createNewFileState, getFileMode } from './fileState';
 const defaultSettings = {
   linewrap: false,
   lineNumbers: false,
-  autocomplete: false,
+  autocomplete: true,
   autocloseBracketsQuotes: false,
   onUpdateLinting: () => {},
   onViewUpdate: () => {},
@@ -69,7 +69,7 @@ describe('createNewFileState - Settings', () => {
     const settings = {
       linewrap: true,
       lineNumbers: true,
-      autocomplete: false,
+      autocomplete: true,
       autocloseBracketsQuotes: false,
       onUpdateLinting: jest.fn(),
       onViewUpdate: jest.fn()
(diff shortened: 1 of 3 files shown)


ksen0 (approved):
Hi @raclim @doradocodes I think this could be added in an upcoming release?

PR #4193, opened 2026-06-25
fix switching p5.js version deactivates syntax highlighting

### LINKED ISSUE #4191: Switching p5.js version deactivates syntax highlighting
Reproduced in Chrome, Firefor, and Safari, on macOS 26.5.1 (Tahoe)

## Before (fresh editor tab)
Image

## After switching version
Image

skyash-dev:
### Issue:
Fixes #4191

my bad. i was checking for autocompleteHinter instead of p5Version earlier to configure the language, which also includes highlighting.

as a result, the highlight didn't work when switching versions if autocomplete was off.

### DIFF
--- a/client/modules/IDE/components/Editor/codemirror.js
+++ b/client/modules/IDE/components/Editor/codemirror.js
@@ -168,7 +168,7 @@ export default function useCodeMirror({
   }, [autocompleteHinter, referenceBaseUrl]);
   useEffect(() => {
     const reconfigureEffect = (fileState) => {
-      const fileMode = getFileMode(file.name);
+      const fileMode = getFileMode(fileState.filename);
 
       if (fileMode !== 'javascript') {
         return [];
--- a/client/modules/IDE/components/Editor/codemirror.js
+++ b/client/modules/IDE/components/Editor/codemirror.js
@@ -170,7 +170,7 @@ export default function useCodeMirror({
     const reconfigureEffect = (fileState) => {
       const fileMode = getFileMode(file.name);
 
-      if (fileMode !== 'javascript' && !p5Version) {
+      if (fileMode !== 'javascript') {
         return [];
       }
 
--- a/client/modules/IDE/components/Editor/codemirror.js
+++ b/client/modules/IDE/components/Editor/codemirror.js
@@ -167,10 +167,15 @@ export default function useCodeMirror({
     });
   }, [autocompleteHinter, referenceBaseUrl]);
   useEffect(() => {
-    const reconfigureEffect = (fileState) =>
-      fileState.languageCpt.reconfigure(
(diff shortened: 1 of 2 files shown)


client/modules/IDE/components/Editor/codemirror.js
```diff
@@ -169,7 +169,7 @@ export default function useCodeMirror({
   useEffect(() => {
     const reconfigureEffect = (fileState) =>
       fileState.languageCpt.reconfigure(
-        autocompleteHinter ? p5JavaScript(p5Version) : []
+        p5Version ? p5JavaScript(p5Version) : []
```
khanniie: I'm comparing this to what we have in `createNewFileState` in utils/fileStates and I'm a bit worried about what happens if the language was not javascript. we may be applying the wrong language configurations for html, css, etc. can you use the logic in createNewFile state, which checks for the language mode first?

skyyash: good catch, thanks!

khanniie (approved):
lgtm! thank you for the fix!

client/modules/IDE/components/Editor/codemirror.js
```diff
@@ -167,10 +167,15 @@ export default function useCodeMirror({
     });
   }, [autocompleteHinter, referenceBaseUrl]);
   useEffect(() => {
-    const reconfigureEffect = (fileState) =>
-      fileState.languageCpt.reconfigure(
-        autocompleteHinter ? p5JavaScript(p5Version) : []
-      );
+    const reconfigureEffect = (fileState) => {
```
khanniie, line 170: i tested this branch locally and i think it's still broken for non-javascript files. should this be if (fileMode!== 'javascript' || ...) instead of an &&?

khanniie (changes requested):

client/modules/IDE/components/Editor/codemirror.js
skyyash: i goofed. we dont need to check for p5Version at all, right?

khanniie: yeah I don't think so! but I think even so, I still see an error while testing if you change the mode it will apply the same language to all files. 

https://github.com/user-attachments/assets/9f01e615-5d66-4579-b8b8-5423f866f6b8

I did some investigation and I think it's because you're checking file.name, which is the current filename, not the filename of the actual filestate. You'll need to change createNewFileState to actually add filename to the data structure that gets returned, but then you can check it properly. I've added some screenshots of the changes below!

Image

Image

skyyash: thanks for digging into this and pointing out the fileState context

khanniie (approved):
thanks for the changes! just tested and looks fixed to me!

PR #4211, opened 2026-07-18
Improve speed of CI tests with cacheing

clairep94:
### Issue:
Fixes # 
- e2e.yml has some cacheing but test.yml has none
- These can both be improved in speed by cacheing node_modules
  - If anything in package-lock.json is updated, the cache is invalid & we do a full install instead

### Demo:

**test.yml** before vs after hitting the node_modules cache
Screenshot 2026-07-19 at 01 01 58

**e2e.yml** before and after hitting the node_modules cache
Screenshot 2026-07-19 at 01 01 29

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -5,6 +5,8 @@ on:
     branches: [develop]
   issue_comment:
     types: [created]
+env:
+  WAIT_ON_VERSION: '7'
 jobs:
   test-e2e:
     # temporary while new e2e environment is not yet created:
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e.yml
```diff
@@ -74,12 +84,23 @@ jobs:
...
+      uses: actions/cache@v4
+      id: wait-on-cache
+      with:
+        path: ~/.npm/_npx
+        key: npx-wait-on-${{ env.WAIT_ON_VERSION }}-${{ runner.os }}
+
+    - name: Install wait-on
+      if: steps.wait-on-cache.outputs.cache-hit != 'true'
+      run: npx -y wait-on@$WAIT_ON_VERSION --version
+
```
clairep94, line 68: previously we were doing the wait-on install in ln 103

I separated it in case cacheing it would help

I don't think it makes a huge difference (3s) but also doesn't hurt

.github/workflows/test.yml
```diff
@@ -7,12 +7,29 @@ jobs:
...
+
+      - name: Create .env file
+        run: cp .env.example .env
+      
+      - name: Cache node_modules
+        uses: actions/cache@v4
+        id: node-modules-cache
         with:
-          node-version: '18.20.x'
-      - run: npm install
```
clairep94, line 15: checkout v2 and setup-node v1 are really old so I bumped them both up to 4 to match e2e.yml. Also wasn't sure if cache@v4 required these to be v4

.github/workflows/test.yml
```diff
@@ -7,12 +7,29 @@ jobs:
...
+          path: node_modules
+          key: node-modules-${{ runner.os }}-node-${{ hashFiles('.nvmrc') }}-${{ hashFiles('package-lock.json') }}
+
+      - name: Install dependencies
```
raclim, line 29: Sorry if this is is something I could've noted earlier! I think I feel a little cautious about `caching `node_modules` because of the potential for consistency issues with cached installs. However, I don't think this is a major concern, especially since we want this to align with what we have in `e2e.yml`, and the runtime improvements are nice!

raclim (dismissed):

raclim:
Sorry for the belated review, I think this overall looks good! I think once the merge conflicts are resolved (which I'm assuming might also happen in 4210), we can get this in!

PR #4247, opened 2026-08-10
Fix line numbers getting clipped past 9999 lines

### LINKED ISSUE #4237: Editor does not display line numbers > 9999 properly (limited to 4 digits)
### p5.js version

2.3.2

Mac OS

### Web browser and version

Chrome 149.0.7827.201

### Actual Behavior

OK hear me out, I'm fully aware that having more than 10,000 lines of code in a single file is _ridiculous_ and probably something we don't want to encourage. Nevertheless, it can happen. And currently, when it does happen, the editor only displays the lesser 4 digits, as you can see in this screenshot.  

Image

The problem is that this gets confusing once I'm looking at line 11000, when it suddenly appears as if I'm reading line 1000. And I'm not able to expand the panel manually with any UI, as far as I'm aware.

### Expected Behavior

I was expecting the editor to be able to display 5-digit line numbers. Or, to expand to accommodate 5-digit line numbers if necessary.

Thanks for your consideration!

### Steps to reproduce

```js

// It says to paste my code here but I'm not going to paste 10K lines :)

```

NisargJasani0602:
Fixes #4237

### Issue:
Line numbers beyond 4 digits (e.g. 10000+) were clipped in the gutter — `11000` rendered as `1000` — because `.cm-gutters` had a fixed width sized for exactly 4 digits, even though CodeMirror is configured with `gutters({ fixed: false })` to let the gutter grow.

### Demo:
image

### Changes:
- Added `gutterWidth.js`: a small CodeMirror `ViewPlugin` that tracks the digit count of the current line count and writes it to a  `--cm-line-number-digits` CSS variable on the editor's DOM node.
- Wired it into `fileState.js`'s extensions list.
- Updated `.cm-gutters` in `_editor.scss` to size itself off that variable via `calc()`, so it grows past 4 digits instead of clipping. Also switched the numbers' `right` offset from a percentage to a fixed value, so the gap to the code stays constant as the gutter grows.


khanniie:
Hi @NisargJasani0602 I was also looking into the bug (my apologies, i should've assigned myself earlier), but since you've put together this PR you're definitely welcome to take the bug instead! Thank you for your contribution! I think there's a way we can still do a CSS-only fix. I'd prefer not to unnecessarily add javascript and more complexity to the editor if it could be avoided. 

I was thinking we could make the overall gutter container a display flex with a minimum width, and allow it to grow if the line number column gets too long? and then the three gutter columns (line numbers, folding, linting) can all have flex-shrink set to 0 (unshrinkable) with the line number column growing to meet the minimum width as-needed. let me know if that makes sense! 

if you give it a try and it seems unreasonable or doesn't work, we can def keep your existing approach!

NisargJasani0602:
Hi @khanniie, thanks for the suggestion! I've updated the PR to use your approach — turns out `.cm-gutters` was already `display: flex` with `flex-shrink: 0` gutters by default from CodeMirror's own base theme, so the fix ended up being mostly about removing the `position: absolute` overlay hack on the three gutter columns (line numbers, fold markers, lint markers) rather than adding anything. That alone lets CodeMirror's native flex sizing handle the growth correctly.

Changes:

- Removed the JS ViewPlugin and CSS custom property from my original approach entirely.

- `.cm-gutters` now uses `min-width` instead of a fixed width.

- Each gutter column no longer uses `position: absolute`, so they lay out normally in the flex row with` flex-shrink: 0,` and the line-number column has `flex-grow: 1 `to fill any leftover space up to the minimum width.

Tested locally with a 10,000+ line file — line numbers display in full without clipping, and normal-size files render unchanged. No JS added. Let me know if you'd like anything adjusted!

khanniie:
Hi @NisargJasani0602 Thank you for the change! It definitely looks much cleaner now : D appreciate it very much

I think we're almost there but I tested the branch locally and noticed a bug. I think the linter column needs to stay absolutely positioned and 100% of the width, it's meant to show behind the row gutter. You can test by triggering the linter by typing some incorrect javascript

Currently (correct behavior)
Screenshot 2026-08-13 at 10 39 30 AM


With your change
Screenshot 2026-08-13 at 10 39 13 AM

NisargJasani0602:
Thanks for catching that, and for the detailed screenshots — that made it easy to pin down!

You're right — `.cm-gutter-lint `needs to stay `position: absolute; width: 100%` since it's a full-width background layer meant to render behind the other columns, not its own side-by-side column. I've restored that.

While digging into it, I also found a second, related bug that wasn't visible in the screenshots: restoring the lint gutter's absolute positioning reintroduced a stacking-order issue where it started rendering on top of the line numbers and fold arrows instead of behind them (since those became position: static flex items in the rewrite, z-index was silently no longer doing anything for them). That was making the fold arrows unclickable, and causing multiple simultaneous lint errors on different lines to render incorrectly. Fixed by giving the shared `.cm-gutter class position: relative; z-index: 1;` so it's positioned enough for the stacking order to apply again, without pulling it out of the flex layout (which would've brought back the original digit-clipping bug).

image

khanniie (approved):
Thank you very much! that looks much better : )

PR #4252, opened 2026-08-12
Create `release` workflow

clairep94:
### Issue:
- [x] Stops deploy-staging from triggering on unsuccessful unit test runs & only occur on successful ones
- [x] Update `deploy` to run after a Github release occurs (currently happens on merges to the `release` branch)
- [x] Adds `release-reminder.yml` which runs Tuesday 15:30 UTC -- generates Issue with the `create-release` workflow link
- [x] Adds `release.yml` which does two jobs: `verify` (CI testing) & `release` (creates the release draft, version, tags, etc). `release` waits on maintainer approval. If `verify` fails, the workflow run fails.  


clairep94:
cc: @doradocodes @raclim

.github/workflows/release.yml
```diff
@@ -0,0 +1,95 @@
...
+
+      - name: Merge back into develop
+        run: |
+          git checkout develop
+          git pull origin develop
+          git merge --no-ff "release-${{ inputs.newversion }}" -m "Merge release-${{ inputs.newversion }} into develop"
+          git push origin develop
+
+      - name: Delete temporary release branch
+        run: git push origin --delete "release-${{ inputs.newversion }}"
```
doradocodes: Wondering if we should really delete after merging, in case we need the branch again later? @raclim what do you think?

contributor_docs/release.md
```diff
@@ -3,13 +3,39 @@
...
+
+### via GHA:
+
+1. An issue titled "Weekly release - DATE" will be created every Tuesday at 15:30 UTC. Go to the [Issues tab](https://github.com/processing/p5.js-web-editor/issues) and click the link in the description to the [Create a release](../.github/workflows/release.yml) GitHub Actions workflow.
+2. Run the **Create a release** workflow via "Run workflow".
+3. Enter the `version` to release
+  1. `version` can be `major`, `minor`, `patch`, etc. (see [npm-version](https://docs.npmjs.com/cli/version) for valid values.
+4. Wait for the workflow to run the `verify` job. 
+  1. This tests the checkout code from `develop`. 
+  2. If tests fail, the workflow will stop & record its result as a failure. Devs should fix any test failures, then attempt the [Create a release](../.github/workflows/release.yml) workflow again.
```
doradocodes, line 24: So if the tests fail, will the version get bumped twice?

doradocodes:
Thanks for working on this so quickly @clairep94 !

contributor_docs/release.md
clairep94: No the `release.yaml > verify` job doesn't include steps for bumping, those are in the `release.yaml > release` job
The input for `version` is just captured and used later on the `release.yaml > release` job (if `release.yaml > verify` passes)

.github/workflows/release.yml
clairep94: Do you think we should push the branch to retain a history of these?

If so, would it be useful to update this to "release-${{ inputs.newversion }}-auto" (or similar suffix) to track when we're merging releases from the workflow vs. manual?

.github/workflows/release.yml
```diff
@@ -0,0 +1,95 @@
...
+        run: npm run test:ci
+
+      # future: deploy to staging (+stop deploying to staging upon merges to develop) + e2e test pointed at staging
+
+  release:
+    needs: verify
+    runs-on: ubuntu-latest
+    # Requires a maintainer to approve in the Actions UI before this job runs.
+    # Configure required reviewers under Settings > Environments > release-approval.
+    environment: release-approval
```
clairep94, line 45: @doradocodes @raclim could we add this new github actions environment, with a list of designated approvers

Issue #4253, opened 2026-08-13
Keyboard shortcuts modal is out of sync with actual keybindings


NisargJasani0602:
Follow-up to #4222 (see [this comment](https://github.com/processing/p5.js-web-editor/issues/4222#issuecomment-5284411810) for context) — `KeyboardShortcutModal.jsx` has drifted from what's actually wired up in `keymaps.js`. A few known discrepancies:

**Advertised but not implemented:**
- The modal lists `{Ctrl/Cmd}+H` for "Replace," but there's no `replace` command wired up anywhere in `keymaps.js`. Worth deciding: implement it (CodeMirror 6's search panel already supports replace, just needs to be exposed on this shortcut), or remove the claim from the modal if it's not planned.

**Implemented but not advertised:**
- `Ctrl-Shift-Up/Down` (`Cmd-Ctrl-Up/Down` on Mac) — move line up/down, added in #4250.
- `Shift-Ctrl-D` (`Shift-Cmd-D` on Mac) — duplicate line, and `Ctrl-Up/Down` (`Ctrl-Alt-Up/Down` on Mac) — scroll without moving cursor, both pending in a follow-up PR.

## Proposal
1. Decide the fate of the Replace shortcut (implement vs. remove from modal).
2. Add entries to `KeyboardShortcutModal.jsx` for the shortcuts above.

Happy to take this on.

NisargJasani0602:
Found a few more discrepancies while continuing the audit:

**Phantom, needs a decision:**
- **`{Ctrl/Cmd}+.` "Comment Line"** — nothing implements this. Only `Mod-/` actually toggles comments. Worth restoring though: I found the old pre-CM6 config explicitly bound this as a fallback (`// Note: most adblockers use the shortcut ctrl+.`), so it looks like a deliberate accessibility/compatibility choice that got lost in the migration, not just a stale doc entry.
- **`F2`/`Ctrl+F2` "Rename Variable"** — also nothing implements this. There's dead code for it (`rename-variable.js`, `showRenameDialog.jsx`), but nothing in the entire client codebase imports or uses them — no keydown listener exists anywhere. Unlike `Mod-.`, I haven't verified this dead code even works, so before deciding to wire it up I'd want to check that first.

For both of these: implement (restore/wire up) or drop the claim from the modal? Same open question as the Replace shortcut on #4255.

[...]

Issue #4222, opened 2026-07-23
CTRL+SHIFT+UP no longer moves lines


not-without-text:
### What is your operating system?

None

### Actual Behavior

Pressing CTRL+SHIFT+UP and CTRL+SHIFT+DOWN act the same way as SHIFT+UP and SHIFT+DOWN.

https://github.com/user-attachments/assets/d4391562-596d-4b49-a765-466bbc481b5b

### Expected Behavior

Pressing CTRL+SHIFT+UP and CTRL+SHIFT+DOWN should move the lines up and down. Here I use the older p5.js editor as archived [here](https://web.archive.org/web/20260612190248/https://editor.p5js.org/).

https://github.com/user-attachments/assets/ae0b45f0-7828-4bbe-b85d-214071cf418c

### Steps:
1. Open https://editor.p5js.org.
2. Go to any line, or select a group of lines.
3. Press SHIFT+UP and SHIFT+DOWN to expand or contract the selection.
4. Press CTRL+SHIFT+UP and CTRL+SHIFT+DOWN, and observe that nothing different happens.

khanniie:
@raclim I think this was one of those sublime text specific keyboard shortcuts that didn't get moved over!

NisargJasani0602:
Hi! I dug into this and found the root cause, plus some related context that might be useful for future work.

Root cause: CodeMirror 6's `defaultKeymap` only binds line-moving to `Alt-ArrowUp/Alt-ArrowDown`, not `Ctrl-Shift-ArrowUp/Ctrl-Shift-ArrowDown`. With no binding for that combo, the keypress goes unhandled by CodeMirror and the browser's native "extend selection" behavior for `Shift-ArrowUp/Down` fires instead — matching exactly what's described here.

This is a casualty of the CM5 → CM6 migration: the old editor used CodeMirror 5's sublime keymap addon (`keyMap: 'sublime'`), which bound `Ctrl-Shift-Up/Down` (and a lot more) out of the box. That addon doesn't exist in CM6, and it looks like this specific mapping was never carried over.

Fix: I've opened a PR adding the missing binding via CM6's own `moveLineUp/moveLineDown` commands (same commands already powering `Alt-Up/Down`), for both `Ctrl-Shift-ArrowUp/Down` (Win/Linux) and `Cmd-Ctrl-ArrowUp/Down` (Mac, matching Sublime's actual Mac shortcut).

I'd like to work on this one — could it be assigned to me?

[...]

PR #4255, opened 2026-08-14
Restore Replace shortcut and document Move Line in shortcuts modal

NisargJasani0602:
## Summary
Partial fix for #4253 (keyboard shortcuts modal out of sync with actual
keybindings). Covers two of the items identified there:

- **Restored the Replace shortcut** (`Ctrl+H` on Windows/Linux, `Cmd+Alt+F` on Mac — matching Sublime and the pre-CM6 config). This was a genuine regression from the CM5 -> CM6 migration, not just a documentation gap: the old editor explicitly bound this via CM5's `replace` command, and the modal already advertised the correct platform-specific combo, but nothing implemented it. Fixed by wiring it to CodeMirror 6's `openSearchPanel`, whose panel already includes Replace/Replace All fields built in.
- **Added Move Line Up/Down to the shortcuts modal** (from #4222, already merged, just never documented).

## Scope note
While auditing the modal against the actual keymap, I found several more real gaps (a phantom "Comment Line" shortcut that isn't implemented, and several working-but-undocumented shortcuts like select-next-occurrence, multi-cursor, delete-line, go-to-line, etc).
Keeping those out of this PR to stay reviewable — tracking as a follow-up.

## Test plan
- [x] `npm run lint` passes
- [x] `npm test -- client/modules/IDE/components/Editor` passes
- [x] Manually verified in the running dev server: `Cmd+Alt+F` opens
      the search panel with Replace fields; the modal shows the new
      Move Line Up/Down entries


NisargJasani0602:
Quick note before this gets reviewed — I want to flag something I noticed after implementing this.

Ctrl+F (already bound to openSearchPanel) opens the same search panel that already includes the Replace field and buttons built in — there's no separate "replace mode" in CodeMirror 6's search UI. So the Ctrl+H/Cmd+Alt+F binding added here doesn't unlock any new functionality; it's just an additional key combo that calls the exact same openSearchPanel function Ctrl+F already calls.

I still think there's a reasonable case for keeping it (it restores what the pre-CM6 editor explicitly had, and the shortcuts modal has been advertising a dedicated Replace shortcut since before I touched anything, so this closes a documented promise), but it's also fair to call it redundant — one keystroke away either way, and the Mac binding (Cmd+Alt+F) is a little awkward specifically because Cmd+H is OS-reserved.

Two ways to go from here:

Keep this PR as-is (dedicated Replace shortcut).
Simplify instead — drop the separate shortcut, and just document Ctrl+F as "Find / Replace" in the modal, since it already covers both.
Happy to go either way — let me know your preference and I'll adjust.

khanniie:
Hi @NisargJasani0602 thank you for the PR! Apologies on the delay. I think we recently updated the find/replace modal a bit so you may need to either update the logic for pulling up the replace modal (which is both find/replace but now the replace is hidden by default so you could add logic to pull it up with replace opened) or just changing the keyboard shortcut to document them as one command.

NisargJasani0602:
Thanks for the pointer! I've merged in the recent develop changes and updated this PR:

- Removed the dedicated Replace keybinding (Ctrl-h/Cmd-Alt-f) since it relied on openSearchPanel, which no longer exists after the SearchPanel rewrite.

- Updated the shortcuts modal to document Find and Replace as a single shortcut (Find / Replace Text under ⌘/Ctrl+F), since Replace is now a toggle inside that panel rather than a separate command.

Also kept the Move Line Up/Down documentation from the original PR.

PR #4259, opened 2026-08-17
GSOC-E2E: Add unsaved changes warning test

Geethegreat:
### Issue:
Fixes # 

Adds a test that verifies the editor warns users before navigating away with unsaved changes.

### Demo:

https://github.com/user-attachments/assets/7133de99-c33f-4f3f-9cd8-ccdc9152935e

### Changes:

 - Log in as a test user
 - Type code into the editor without saving
 - Wait for the unsaved changes indicator to appear (the star icon that renders off state.ide.unsavedChanges), more reliable
 than an arbitrary timeout
 - Attempt to navigate to the sketches page via the nav dropdown
 - Verify the unsaved changes dialog appears with the correct message
 - Dismiss the dialog, confirm the editor is still visible (navigation was blocked)
 - Accept the dialog on the second attempt, confirm navigation goes through to the sketches page

### DIFF
--- a/e2e/specs/unsaved_changes.spec.ts
+++ b/e2e/specs/unsaved_changes.spec.ts
@@ -1,15 +1,13 @@
-import type { Dialog, Page } from '@playwright/test';
+import type { Page } from '@playwright/test';
 import { test, expect } from '../fixtures';
 import { dismissCookieBanner } from '../helpers/cookie-banner';
 import { createTestUser, loginAs, TestUser } from '../helpers/auth';
 
-/**
- * Renders off state.ide.unsavedChanges (see UnsavedChangesIndicator.jsx) —
- * the same state that gates both the <Prompt> and the beforeunload listener.
- */
+// Renders off state.ide.unsavedChanges (see UnsavedChangesIndicator.jsx).
 const unsavedIndicator = (page: Page) =>
   page.getByRole('img', { name: 'Sketch has unsaved changes' });
 
+// Covers the unsaved changes warning that appears when navigating away from the editor with unsaved changes.
 test.describe('unsaved changes warning', () => {
   let testUser: TestUser;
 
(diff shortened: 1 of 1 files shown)


Geethegreat:
The alert is not visible in the screen recording

e2e/specs/unsaved_changes.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+      await dialog.dismiss();
+    });
+
+    await page.getByRole('menuitem', { name: testUser.username }).click();
+    await page.locator('#account-sketches').click();
+
+    await expect
+      .poll(() => dialogMessage)
+      .toContain('You have unsaved changes');
+    await expect(editor).toBeVisible();
```
clairep94: Do we need to poll?

I think it would be easier to understand if we do the below (not auto-close the dialog)

```suggestion
    let dialogMessage = '';
    page.once('dialog', async (dialog) => {
      dialogMessage = dialog.message();
    });

    await page.getByRole('menuitem', { name: testUser.username }).click();
    await page.locator('#account-sketches').click();

    await expect(dialogMessage).toContain('You have unsaved changes');
    
    page.once('dialog', async (dialog) => {
      await dialog.dismiss()
    });
    
    await expect(editor).toBeVisible();
    // maybe add that the URL didn't change -- the attempt to nav to a different page was blocked
```

e2e/specs/unsaved_changes.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    await page.waitForTimeout(1000);
```
clairep94: what's this timeout for?

e2e/specs/unsaved_changes.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    await dismissCookieBanner(page);
+    await loginAs(page, testUser);
+  });
+
+  // Covers the <Prompt> in IDEView.jsx, which guards in-app navigation and
+  // shows the app's own message via window.confirm(). The sibling
+  // beforeunload handler (full page unload, e.g. reload/close) is a separate
+  // guard on the same state, not covered here.
+  // Note the Prompt deliberately allows /login, /signup and /feedback through
+  // without warning, so this navigates to the user's sketches instead.
```
clairep94: Not sure what this comment is referring to, was this generated?
Can you summarise? Not sure if this level of detail is necessary

e2e/specs/unsaved_changes.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    await editor.click();
+    await page.keyboard.press('ControlOrMeta+A');
+    await page.keyboard.type('function setup() {createCanvas(400, 400);}', {
+      delay: 5
+    });
+
+    // Wait for the app to actually register the unsaved change rather than
+    // sleeping past CodeMirror's 1s debounce. This indicator renders off
+    // state.ide.unsavedChanges — the same state that gates both guards.
+    await expect(unsavedIndicator(page)).toBeVisible({ timeout: 10_000 });
```
clairep94: If you want to do the current structure, where you have the root `test.describe('unsaved changes warning'`, we can move these to the beforeEach block

Then do more than one nested test:
`test('warns before navigating away from the editor with unsaved changes')`
- only check that a warning message pops up

`test('blocks navigation if the warning is dismissed')`
- only check that the url does not change after dismissing the popup

`test('allows navigation if the warning is accepted')`
- check that the url changes after accepting the popup

`test('does not appear when a sketch is saved before navigating')`

clairep94: OR do it as one long test and remove the beforeEach. add comments for each 'step', as you have done

clairep94 (approved):

PR #4282, opened 2026-09-01
GSOC-E2E: Add confirmation of sketch loading before playing it

Geethegreat:
### Issue:
Fixes # 

### DIFF
--- a/e2e/specs/sketches_management.spec.ts
+++ b/e2e/specs/sketches_management.spec.ts
@@ -126,6 +126,11 @@ test.describe.serial('sketch lifecycle', () => {
       .getByText('renamed-sketch')
       .click();
 
+    // Confirm the reopened sketch has actually loaded before playing it
+    await expect(
+      page.locator('button.editable-input__label')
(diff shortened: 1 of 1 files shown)


e2e/specs/sketches_management.spec.ts
```diff
@@ -126,6 +126,11 @@ test.describe.serial('sketch lifecycle', () => {
       .getByText('renamed-sketch')
       .click();
 
+    // Confirm the reopened sketch has actually loaded before playing it
+    await expect(
+      page.locator('button.editable-input__label')
+    ).toContainText('renamed-sketch', { timeout: 10_000 });
+
```
clairep94, line 133: @Geethegreat could you explain this a bit further?

How do we know that this is a good indicator of the sketch being loaded?

Geethegreat: The name and the files both come from the same API response and the same Redux dispatch (SET_PROJECT). As both reducers (project.js and files.js) react to that one action, state.project.name and state.files update together. So if we see the file name it implies that the file came with it.

clairep94 (approved):

